PR #1, opened 2020-04-22
migrate dht test plan from testground/testground.

raulk:
Unfortunately the composition files in this test plan use a few elements that make it non-portable:

- pre-built artifacts: this means that this test plan can only run from the machines/environments where those artifacts are stored.
- specific upstream commit hashes: those commits may no longer exist.

---

It has to be understood that this is the test plan that the IPFS Content Routing WG used to iterate and experiment with DHT changes across branches. It's part of the work product, rather than a result. Hence its state.

If in the future we intend to run this test plan on a continuous basis on CI, we will need to revisit some things. For now, I'm just copying it over as-is.

---

This is the inbound PR.

Outbound PR: https://github.com/testground/testground/pull/892.


aschmahmann:
@raulk I can't see anything obviously wrong with this PR (although it's a huge copy-paste so a bit difficult to look through), aside from the powershell scripts being copied twice. Once into `/dht` and another time into `/dht/scripts`. `/dht/scripts` is the correct location.

raulk:
@aschmahmann thanks for the prompt review! The scripts appear to be duplicated at origin, hence why they're duplicated here too. I'll fix it before merging.

aschmahmann:
Oh, thanks. Guess something strange happened there.

PR #9, opened 2020-07-22
add an instructional libp2p ping test plan.

raulk:
cc @nonsense 
cc @hacdias


ping/README.md
```diff
@@ -0,0 +1,161 @@
...
+Jul 22 12:31:01.688163	INFO	created file: 852eede73af1.tgz
+```
+
+You can now inspect the collected results by opening the file.
+
+If you want to see the metrics we recorded in InfluxDB visually, you can fire
+up a Chronograf console:
+
+```shell script
+$ docker run --network testground-control -p 8888:8888 chronograf
```
nonsense, line 112: You can use something like:

```
$ docker run -p 8888:8888 --net=testground-control chronograf --influxdb-url=http://testground-influxdb:8086
```

And avoid the extra manual steps.

P.S. I haven't tried that, just read the docs for Chronograf.

ping/main.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+// runPing is the test case logic.
+//
+// Its signature conforms to the SDK's run.InitializedTestCaseFn type. As a
+// result, the Testground SDK will perform a few useful preparation steps
+// for us:
+//
+//  1. Initializing a sync client, bound to the runenv.
+//  2. Initializing a net client.
```
schomatis, line 56: Can we clarify a bit more on the difference between a sync and a net client? (They kind of sound very similar to discern in the reader's mental model, I'm probably missing some background here though.)

ping/main.go
```diff
@@ -0,0 +1,284 @@
...
+		case ai := <-peersCh:
+			if ai.ID == id {
+				continue // skip over ourselves.
+			}
+			peers = append(peers, ai)
+		case err := <-sub.Done():
+			return err
+		}
+	}
+	cancel()
```
schomatis, line 171: nit: why do we cancel at the end here? if this is just about releasing resources we might avoid it as (or comment it explicitly as such), otherwise it might confuse the reader thinking this is needed for something more in the correct operation of the synchronization test.

ping/main.go
```diff
@@ -0,0 +1,284 @@
...
+	// ✨
+	// ✨  Now we know about all other libp2p hosts in this test.
+	// ✨
+
+	// This is a closure that pings all peers in the test in parallel, and
+	// records the latency value as a message and as a result datapoint.
+	pingPeers := func(tag string) error {
+		g, gctx := errgroup.WithContext(ctx)
+		for _, ai := range peers {
+			ai := ai
```
schomatis, line 182: Could we document this rebinding please? It might not be obvious to the readers what's happening with the reference here.

ping/main.go
```diff
@@ -0,0 +1,284 @@
...
+	rand.Seed(time.Now().UnixNano() + initCtx.GlobalSeq)
+
+	for i := 1; i <= iterations; i++ {
+		runenv.RecordMessage("⚡️  ITERATION ROUND %d", i)
+
+		// 🤹  Let's calculate our new latency.
+		latency := time.Duration(rand.Int31n(int32(maxLatencyMs))) * time.Millisecond
+		runenv.RecordMessage("(round %d) my latency: %s", i, latency)
+
+		// 🐌  Let's ask the NetClient to reconfigure our network.
```
schomatis, line 255: Do the UTF emojis help? No. Do they make the code more clear? No. Are they so god damn lovely and show the amount of dedication put into this document? Hell yes.

ping/main.go
```diff
@@ -0,0 +1,284 @@
...
+		//
+		// The sidecar will apply the network latency from the outside, and will
```
schomatis, line 257: Can we link to the sidecar doc here? It's a bit odd mentioning it only once without background otherwise.

ping/README.md
```diff
@@ -0,0 +1,161 @@
...
+$ make install       # builds testground and the Docker image, used by the local:docker runner.
+
+$ testground daemon  # will start the daemon listening on localhost:8042 by default.
+# => this will run indefinitely; open a different console to continue.
+```
+
+Now, import this test plan into your `TESTGROUND_HOME` dir (by default `$HOME/testground`):
+
+```shell script
+# clone this repo in your directory of choice; let's call it `$REPO_DIR`
```
schomatis, line 63: Maybe just use `go get` and assume this will be in `GOPATH` to simplify and give a ready-to-use command to the user?

ping/README.md
raulk: That worked! Thanks for the hint.

ping/README.md
```diff
@@ -0,0 +1,161 @@
...
+   counted against the `iterations` param.
+4. It then requests a network configuration change to set a random latency
+   (up to `max_latency_ms`) on our link.
+5. Once all instances have had their latencies applied, we perform a round of 
+   PINGs and record the observed latencies, just like in point 3.
+6. We do 4-6 as many times as `iterations`.
+
+## Installing the test plan
+
+First, install Testground:
```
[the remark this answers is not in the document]
raulk: Added a link to it, but kept it here too for convenience and for avoiding indirection.

ping/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+$ testground daemon  # will start the daemon listening on localhost:8042 by default.
+# => this will run indefinitely; open a different console to continue.
+```
+
+Now, import this test plan into your `TESTGROUND_HOME` dir (by default `$HOME/testground`):
+
+```shell script
+# clone this repo in your directory of choice; let's call it `$REPO_DIR`
+$ git clone https://github.com/libp2p/test-plans $REPO_DIR
```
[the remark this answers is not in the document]
raulk: In Testground we settled for this format, but we can definitely reconsider if it becomes a frustration for others.

ping/main.go
raulk: Was explained in the `main()` function godocs, but it doesn't hurt to expand here. Will do.

image

ping/main.go
raulk: This cancels the sync service subscription, as we're done consuming everything we need. Will comment and change the var name so it's not rebinding (less confusing).

ping/main.go
raulk: Absolutely.

raulk: Simplified this and made it more obvious by only capturing the ID from the AddrInfo.

ping/main.go
raulk: Hahaha, my intention was for them to serve as markers of the large "phases" of the test.

raulk: Glad they lightened your day.

ping/main.go
raulk: Yup, totally.

ping/README.md
raulk: Hum, the go get command has way too many side effects depending on what your `PWD` is at that moment. I'd refrain from doing that.

Issue #10, opened 2020-07-31
Problem compiling testcase dht


lucacampanile3:
Hi guys, i'm using libp2p testcase for test the performance evaluation of the DHT, i already installed succesfully testground and libp2p testing but when i run the command `testground run single --plan libp2p/dht --testcase find-peers  --builder docker:go --runner local:docker --instances 10` the shell print an unusual error

-- Running in ee5474c9ab62
 github.com/libp2p/test-plans/dht/test
test/common.go:214:34: undefined: getTaggedLibp2pOpts
test/common.go:236:14: undefined: createDHT
test/common.go:759:2: undefined: TableHealth
test/common.go:821:10: undefined: specializedTraceQuery
test/find_providers.go:130:54: undefined: getAllProvRecordsNum
test/tracers.go:77:12: undefined: specializedTraceQuery
Jul 31 17:18:57.547364	INFO	build failed	{"req_id": "e2b68ead", "plan": "dht", "groups": ["single"], "builder": "docker:go", "error": "docker build failed: The command '/bin/sh -c cd ${PLAN_DIR}     && go env -w GOPROXY=\"${GO_PROXY}\"     && GOOS=linux GOARCH=amd64 go build -o ${PLAN_DIR}/testplan.bin ${BUILD_TAGS} ${TESTPLAN_EXEC_PKG}' returned a non-zero code: 2"}
Jul 31 17:18:57.547464	WARN	engine build error: docker build failed: The command '/bin/sh -c cd ${PLAN_DIR}     && go env -w GOPROXY="${GO_PROXY}"     && GOOS=linux GOARCH=amd64 go build -o ${PLAN_DIR}/testplan.bin ${BUILD_TAGS} ${TESTPLAN_EXEC_PKG}' returned a non-zero code: 2	{"req_id": "e2b68ead"}
Jul 31 17:18:57.547502	INFO	request handled	{"req_id": "e2b68ead", "command": "build"}

Maybe there is an error with the dependencies in the file common.go am i right?
But i do not touch the code why i have this problem?

Thanks a lot for helping me 

Regards
Luca

aschmahmann:
The DHT plans must be run as compositions in order to have the correct dependencies and build tags. Try using the all-cypress.toml file for example.

LucaCampanile:
[...]

PR #11, opened 2020-08-01
remove private artifacts

nonsense:
This PR is addressing an issue described at https://github.com/libp2p/test-plans/issues/10 - seems like we are keeping private artifacts as part of our public composition files.

Generally we shouldn't need artifacts any more, since we have improved a lot the building and pushing of images in the last few releases of Testground. I believe artifacts were used heavily by @aschmahmann while Testground didn't have good support for caching and was making redundant pushes to remote registries.

cc @Drengot-p2p

### DIFF
--- a/dht/compositions/all-balsam-k8s.toml
+++ b/dht/compositions/all-balsam-k8s.toml
@@ -20,7 +20,6 @@
   [groups.build]
     selectors = ["balsam"]
   [groups.run]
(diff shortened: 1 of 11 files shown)


nonsense:
cc @raulk seems like I don't have access to this repo and also can't mark you as a reviewer, so just tagging you here ;)

PR #14, opened 2020-11-13
fix ping result metrics

nonsense:
This PR is:
1. Adding a composition for Kubernetes for the libp2p/ping testplan
2. Fixing the emitting of metrics from it (removing random value from `tag` so that cardinality doesn't explode.
3. Upgrading `sdk-go` so that we detect the metrics
4. Fixing the name of the testplan in the manifest.


ping/main.go
```diff
@@ -206,7 +206,7 @@ func runPing(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
 				// into InfluxDB when the test concludes.
 				//
 				// ping-result is the metric name, and round and peer are tags.
-				point := fmt.Sprintf("ping-result,round=%s,peer=%s", tag, id)
+				point := fmt.Sprintf("ping-result.ms,round=%s", tag)
```
nonsense, line 209: We should not have any random fields added as tags, as this will explode the cardinality in InfluxDB.

PR #19, opened 2022-02-21
update go-libp2p to v0.17.0 in ping test, remove secio secure channel

marten-seemann:


mxinden (approved):
Thanks for the quick follow-up @marten-seemann!

PR #20, opened 2022-03-03
compatibility/: Add basic test with Rust and Go establishing TCP conn

mxinden:
This pull request adds a basic composition Testground test, starting two
instances, one Golang and one Rust binary, establishing a TCP
connection to each other.

---

Status:

- I am opening this up for discussion purposes only. This is not meant to be merged.
- Only uses plain TCP, no libp2p involved.
- Plenty of hacks.


laurentsenta:
@mxinden that work was really helpful and a foundation for the interop test as it is now, thanks! I'm closing it now that we merged https://github.com/libp2p/test-plans/pull/26

PR #23, opened 2022-04-25
testground: cross-version interoperability

laurentsenta:
Add a `ping` test, and the composition file to build and run it with different versions of go-libp2p.

**General idea:**

- We agree on a ping test, and implement it using different libp2p implementations,
- Testground runs multiple instances of the same test, with different versions of the library.

### go test:

We add a single `ping-interop/go` test. This test can use multiple versions of libp2p with a single codebase:

- It defines a `go.vxxx.mod` file (and it's sum) for each version,
- It defines a `compat` module that provides a compatibility layer when there are API changes (shims).

### Testground composition

We add a testground composition file `go-cross-versions.toml`
that runs v0.11, v.0.17, v0.19, v0.20, the current `master`, and an optional branch (if `GitReference` is in the environment) together.

Locally you can run this with:

```shell
GitReference=optionnal-custom-libp2p-branch testground run composition \
    -f ./_compositions/go-cross-versions.toml \
    --collect --wait
```

Then the test output is shown on the testground dashboard (localhost:8042 for a default daemon that runs locally).
Beware of https://github.com/testground/testground/issues/1329 when you check results.

Note it'll use a go mod `replace` to test the latest master and latest version of an optional branch.

### CI

We define a reusable workflow we can from another repository, it's called with:

```yaml
jobs:
  run-libp2p-test-plans:
    uses: "singulargarden/test-plans/.github/workflows/test-compatibility.yml@master"
    with:
      composition_file: "ping-interop/_compositions/go-cross-versions.toml"
      test_case: "ping"
      custom_git_reference: "current-pr" # optional
      custom_git_target: "my-fork/go-libp2p" # optional
      testground_endpoint: "https://ci.testground.ipfs.team/" # optional
```

This action will build & run the test.

## Todo:

- [x] plan/ping: fix test & use go 1.17
- [x] plan/ping: fix test to make it (more) deterministic on local env
  - ⚠️ It is stable on an Ubuntu machine, but I couldn't get the macos testground to give use reliable result for now
- [x] plan/ping-interop: implement cross-version testing
  - [x] version 0.19
  - [x] version 0.11
- [x] ci: provide reusable workflo
- [x] implement interop feature support in testground
- [x] remove the shell scripts when https://github.com/testground/testground/issues/1329 is fixed and merged
- [x] confirm this is passing on github actions with our self-hosted runners
- [ ] Remove the `DEBUG` commits used for testing


ping/main.go
```diff
@@ -214,14 +216,15 @@ func runPing(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
 	// ☎️  Connect to all other peers.
 	//
 	// Note: we sidestep simultaneous connect issues by ONLY connecting to peers
-	// who published their addresses before us (this is enough to dedup and avoid
+	// that have a peerID smaller to our (this is enough to dedup and avoid
```
laurentsenta, line 219: As far as I understand this case can not happen because we never append ourselves (see line ~173)

mxinden: We never connect to our self. In addition, by `break`ing early we also prevent two nodes to both connect to each other, thus resulting in two connections between each pair of peers. Does that make sense @laurentsenta?

laurentsenta: That's the intent of the code,
but as far as I understand we never add ourselves to the `peers` array:

Note the `continue` when we find ourselves, at construction time:
https://github.com/libp2p/test-plans/blob/8497c05980adb9651aad78b15d515bf97f947434/ping/main.go#L167-L178

So later on, there is no item that verifies `ai.ID == id` => We don't dedup the connections.

galargh:
Unfortunately, I won't be able to meaningfully review this before my short break (until May 4). I'm OK with you moving forward with this though and I can review it post-merge.

marten-seemann (commented):
Not an in-depth review yet, but a few high-level thoughts:
1. iiuc, the current setup is limited to testing interop between exactly 2 version. It would be nice to have a setup where we could spin up nodes using multiple different versions, especially as we're going to cut more (and smaller) releases in the future. Is that possible?
2. code reuse: it would be nice if we could write the code once, and then run it using multiple version. In my mind, this raises two questions:
    1. What if a breaking API change happened between two versions?
    2. How to deal with go.mod
    
I don't have a good answer for 2., but for 1., it might be nice to use a build flag. The implementation could then work around the breaking interface change.

BigLep:
2022-05-27 conversation: there was a verbal call on how to move forward.  I'm going to move this to draft.  Please publish @laurentsenta again when it's ready for review.

laurentsenta:
Updated the content and description now that we have testground support with: https://github.com/testground/testground/issues/1337

Still a work in progress until the few items in the todo (test with self-hosted runners and https://github.com/testground/testground/issues/1329) are complete.

ping/main.go
```diff
@@ -13,7 +13,7 @@ import (
```
mxinden, line 13: I wonder whether we still need the normal ping example now that we have the ping-interop example. No strong opinion, though might be worth deduplicating.

.github/actions/setup-testground/action.yml
```diff
@@ -0,0 +1,35 @@
...
+        if [[ ! -z "${TESTGROUND_ENDPOINT}" ]]; then
+          mkdir -p ~/testground/;
+          cat <<EOF >> ~/testground/.env.toml
+
+          [client]
+          endpoint = "${TESTGROUND_ENDPOINT}"
+        EOF
+        else
+          testground daemon > testground.out 2> testground.err &
+        fi;
```
mxinden, line 35: Would it make sense to wait for the daemon to be ready here as well?

https://github.com/testground/sdk-rust/blob/f19ef9621e319b06f8def74a6f28b715feca7010/.github/workflows/ci.yml#L107-L115

galargh (commented):
Looks great :) Just one thought - maybe a little README on how to add a new version to ping-interop test would be useful. I know it's fairly simple but I think someone, in a couple of months, might appreciate it being written out anyway.

.github/actions/setup-testground/action.yml
laurentsenta: That make sense, thanks for the catch, fixed!

ping/main.go
laurentsenta: Agreed, fixed!

laurentsenta:
2 updates:

- I moved the proxy to proxy.golang.org, and increased every timeout, and now the test is passing consistently on regular GitHub runners! I still can't believe that was that easy, I've seen a lot of random freezes and fails when developing this on local, but it looks like I was wrong @mxinden, regular runners are enough for this test! 🙇
- I updated the composition to test forks too, so [we are able to test pull request](https://github.com/libp2p/go-libp2p/compare/master...laurentsenta:ci/cross-version-interop-testing?expand=1) from forks if we want too,
  - there is a [bug with branch names](https://github.com/testground/testground/issues/1366), but if we use commits `sha` we'll be fine

A few runs as example:
(check the "Run composition file" section for parameters & logs)

- test without any parameter https://github.com/singulargarden/pl-github/runs/7043100620?check_suite_focus=true (test master
- test with github.com/fulviodenza/go-libp2p:master https://github.com/singulargarden/pl-github/runs/7043021338?check_suite_focus=true
- test with marco/canonical-logging branch https://github.com/singulargarden/pl-github/runs/7043448820?check_suite_focus=true

mxinden (approved):

mxinden:
@laurentsenta once this is merged, I can provide the ping Rust implementation based on https://github.com/mxinden/test-plans/tree/compatibility-libp2p. Would appreciate some help hooking it up to the composition test though.

ping/go/main.go
```diff
@@ -13,13 +13,15 @@ import (
 
 	"github.com/libp2p/go-libp2p-core/peer"
 
-	"github.com/libp2p/go-libp2p-noise"
+	noise "github.com/libp2p/go-libp2p-noise"
```
marten-seemann, line 16: This repo is deprecated as of v0.20.0. Same for -tls.

To do things the correct way, we should define a `NoiseProtocolID` in `compat`.

ping/go/main.go
```diff
@@ -120,17 +122,20 @@ func runPing(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
 		security = libp2p.Security(noise.ID, noise.New)
 	case "tls":
 		security = libp2p.Security(tls.ID, tls.New)
+		// TODO: check w/Marten this is fine: We fall into the lowest common denominator for parameters (here, no secio anymore even for legacy versions).
```
marten-seemann, line 125: Yes, no need to test secio any more.

ping/go/manifest.toml
```diff
@@ -4,17 +4,17 @@
 name = "ping"
 
 [builders]
-"docker:go" = { enabled = true, enable_go_build_cache = true }
 "exec:go"   = { enabled = true }
 
+[builders."docker:go"]
+enabled                = true
+build_base_image       = "golang:1.17-buster"
```
marten-seemann, line 11: This might be a dumb question, but what does it mean to specify golang:1.17 here?

laurentsenta: Not dumb at all, it becomes the new default for this test plan, (testground uses 1.16 by default).
We overwrite it in the composition files when we run "real" tests.

I assume if you're trying to run a quick test, you're probably working on the latest version,
so you have one less testground parameter to remember:

```
testground run single --plan=libp2p/ping/go --testcase=ping --builder=docker:go --runner=local:docker --instances=2 --build-cfg modfile=my.new.go.mod
```

ping/go/main.go
laurentsenta: 👍

It looks like we also need to pass the Protocol constructor (`security(noise.ID, noise.New)`), I've implemented a getByName, then moved it into the compat package (less confident about this one). 

- https://github.com/libp2p/test-plans/pull/23/commits/907546613f25a989a35a3f5ef1e382157c487d87
- https://github.com/libp2p/test-plans/pull/23/commits/0f506ecda1042b76b2c722f7fa73d29ee6db7ec6

ping/go/manifest.toml
marten-seemann: That's why I was wondering, the current version is Go 1.18. Not a blocker for this PR though.

marten-seemann (approved):

Issue #29, opened 2022-08-14
interop tests are very slow


marten-seemann:
They're by far the slowest workflow in our CI configuration (even though we haven't even started addressing https://github.com/libp2p/go-libp2p/issues/1632, which would further speed up the other workflows).

Just building testground takes more than 5 minutes, see https://github.com/libp2p/go-libp2p/runs/7825338253?check_suite_focus=true. This could easily be cached: either testground itself could provide pre-built binaries / Docker images, or the workflow could cash its results (under the hash of the most recent master commit, so that updates would automatically trigger a rebuild).

Running the composition takes another 12.5 minutes. I'm sure there are also caching opportunities here, but that might be more challenging.

laurentsenta:
@marten-seemann what would be a "good enough" time for the interop tests,
- to test go versions together?
- to test go + rust versions together?

One thing to note, I'm tackling caching the build artifacts at the moment because these might also fix stability issues:
The issues we saw on go-libp2p happened during `go mod download` and `docker pull`.

marten-seemann:
Do I understand correctly that this is the step that failed? https://github.com/libp2p/test-plans/blob/072d351a5fe0f558d3c5c50eedb952e8159d41a2/.github/workflows/run-composition.yml#L57-L62

Would using one of the GitHub Actions that does automatic retries be an option? For example this one: https://github.com/marketplace/actions/retry-step. (There might be better ones, we'd have to research a bit, this one just showed up as the first search result).

laurentsenta:
Yes, the only failure we saw happened during the build, that's a discussion related to https://github.com/libp2p/go-libp2p/issues/1670 and https://github.com/testground/testground/issues/1422.

We can add retries and different timeouts now that we can split the build & run steps in Testground, but the thing is:

I need to know the "good enough" time, because adding retries will decrease the number of false-negative detected by the CI, but it will also increase the average time it takes to run the interop tests.
Which is between 15 minutes (go-interop) and 30 min (go + rust interop) atm.

marten-seemann:
> Adding retries will decrease the number of false-negative detected by the CI, but it will also increase the average time it takes to run the interop tests. Which is between 15 minutes (go-interop) and 30 min (go + rust interop) atm.

Is it possible to build incrementally? If I understand correctly, `testground build composition` builds _all_ runners (?) required in that test case. If there was a way to build them one by one, the cost for a retyr when one of them fails would be much lower.

Alternatively, `testground build composition` could contain the retry logic.

PR #30, opened 2022-08-22
ping/_composition/go.toml: disable go custom & master branches

laurentsenta:
We disabled interop tests in go-libp2p which means that the go-libp2p team might break the interoperability testing at any time.

That's a problem when it comes to interoperability testing: if we enable interop testing in rust, maintainers will experience random breakage in the CI because of the other repo. We can only play catch-up with the go-side until we re-enable the interop tests there.

This PR disables go's `master` and `custom`  tests until we fix the stability / caching issues and reenable the tests in go-libp2p.

## Related issues & PR

- https://github.com/libp2p/go-libp2p/issues/1670
- https://github.com/libp2p/rust-libp2p/pull/2835

## current go status

The tests started breaking after we moved go-libp2p-core into go-libp2p/core. It breaks every versions at the moment ([work in progress branch](https://github.com/laurentsenta/test-plans/commits/fix/go-libp2p-ping)).

### DIFF
--- a/ping/_compositions/go.toml
+++ b/ping/_compositions/go.toml
@@ -1,12 +1,12 @@
-[master]
-GoVersion = '1.18'
-Modfile = "go.v0.21.mod"
-Selector = 'v0.21'
+# [master]
+# GoVersion = '1.18'
+# Modfile = "go.v0.21.mod"
+# Selector = 'v0.21'
 
-[custom]
-GoVersion = '1.18'
-Modfile = "go.v0.21.mod"
-Selector = 'v0.21'
+# [custom]
+# GoVersion = '1.18'
+# Modfile = "go.v0.21.mod"
+# Selector = 'v0.21'
 
 [[groups]]
 Id = "v0.11"


marten-seemann:
This would be a sad outcome. I didn't mean to break our interop testing story when I disabled the tests, but the flakiness was holding us back from making progress. Do we have any timeline for a more stable version of the test setup?

laurentsenta:
@marten-seemann It really depends on  your definition of "good enough" (https://github.com/libp2p/test-plans/issues/29#issuecomment-1222380579)

If you're fine with ~25 min of CI time (https://github.com/libp2p/test-plans/issues/29), we add retries and re-enable the go cross versions tests by tomorrow.

If you need ~20 min of CI time, we add retries, and improve testground setup (should take less than a week, https://github.com/testground/testground/pull/1371), then re-enable the go cross versions tests.

If you need less, we implement caching with https://github.com/libp2p/test-plans/issues/31 (should take a week or two) then improve testground setup (a week). We might get the cross-version AND cross-implementation testing below the 10 minutes bar.

laurentsenta:
By the way, it's not ideal, but I think it's fine to disable the go master test if it means we can get some workload on the rust side. It will help with stability work more than if we're not running rust at all.

laurentsenta:
Replaced by https://github.com/libp2p/test-plans/pull/32

PR #33, opened 2022-08-30
ping/_compositions/*: fix rust version rewrite

laurentsenta:
Recent builds break because there is a version `0.47.0` in one of our rewrite strings,
 
- Demo run with a custom branch: https://github.com/libp2p/test-plans/runs/8087223516?check_suite_focus=true


mxinden (commented):
:+1:

PR #40, opened 2022-09-07
.github/workflows: use write-artifacts and prevent rebuild

laurentsenta:
fixes #39 

Use the `write-artifacts` option during the build.
With this option, Testground replaces the composition file with the one produced by the build step (template generation applied, build artifact set).

With this we:
- Never rebuild the containers, even on Docker cache miss,
- Can save the generated composition (templating applied) as a useful debug artifact

## Tasks

- [x] merge https://github.com/testground/testground/pull/1444
- [x] drop the debug commit

PR #37, opened 2022-09-05
ping/_composition: use latest release in interop-latest

laurentsenta:
Update the interop (latest) tests so that we build & test:

- latest released go version
- latest released rust version
- current branch (rust or go)

(before we would test against `master` branch for each of these projects).

Issue #31, opened 2022-08-22
Use CI Caching during Interop Tests


laurentsenta:
## Description

Running the ping interop test takes ~1min30.
Building Testground takes ~5min.
Building the Testground instances takes between 15 and 25 min.

This means 2 things:
* Maintainers don't want to wait for the test results
* For every run, we download every docker image, go packages, rust packages, etc. The test is likely to fail because of a network issue. (the total size of build + run images in the full interop test is ~14GB).

### Cache build steps

I am investigating this at the moment because I believe this is where we get the most improvement:
* We don't re-download & rebuild every image and package (lower failure rate due to downloads),
* We speed up build time
  * maybe 15min -> 3 min, it might be better with Rust that relies on Docker layers to cache the base packages.

It's not trivial:
* We need to pick precisely which artifacts need to be cached,
* We have to work around multi-staged builds and 1.10 changes in layers
  * Related: https://stackoverflow.com/questions/49965396/can-i-obtain-the-docker-layer-history-on-non-final-stage-docker-builds and https://stackoverflow.com/questions/35310212/docker-missing-layer-ids-in-output/35312577#35312577
  * I think the solution is to inject a few `LABELS` in our Dockerfile, I'm investigating this at the moment.

### Cache build artifacts

At the end of the build step, it is possible to retrieve the build artifacts (docker images) produced for each group in the composition file. We could send a "pre-filled" composition file to Testground and save the part that rebuild artifacts.

The issue here is: how do you know whether an artifact needs to be rebuilt?
Even for "old" releases (like go-libp2p v0.11), we might change a test in ways that require to rebuild the test.

* benefits: we cache a 50Mb image and never rebuild it,
* issues: we need a manual intervention to tell CI when a test needs to be rebuilt from scratch.

### Do not rebuild Testground on every test

This would save around 4 min of building go & docker images. But we wouldn't get the stability benefits of NOT downloading the go & rust images + packages.

- [ ] use the go cache
- [ ] use Testground binaries
- [ ] Use pre-built docker containers

### Current times
(measured with a single run, but these Github CI timings are pretty consistent)

* rust
  * without custom branch + success: 17 min  
  * with custom branch + success: 24 min 
  * on fail: ~28 min (timeout)
* go
  * with or without a custom branch, success or fail: ~15 min
* rust + go
    * case = latest (go master + rust master + custom branch):
       * on success: ~17min
       * on failure: ~28min
    * case = all (~10 versions + custom branch):
[...]

PR #41, opened 2022-09-10
ping/: Update to rust-libp2p release v0.48.0 and master v0.49.0

mxinden:
See release of v0.48.0 in https://github.com/libp2p/rust-libp2p/pull/2869.

Master bump to v0.49.0 in https://github.com/libp2p/rust-libp2p/pull/2813

Followed the instructions on https://github.com/libp2p/test-plans#how-to-add-a-new-version-to-pingrust.

I skipped step (3), namely to run the new test on my local machine. As far as I can tell this is done on CI here already. Please correct me if I am wrong.

### DIFF
--- a/ping/_compositions/go-rust-interop-latest.toml
+++ b/ping/_compositions/go-rust-interop-latest.toml
@@ -93,7 +93,7 @@
         CARGO_FEATURES = '{{ .CargoFeatures }}'
         CARGO_REMOVE = '{{ .CargoFeatures }}'
         CARGO_PATCH = """
-          {{ .CargoFeatures }} = {package = "libp2p", git = "https://{{ or $.Env.GitTarget "github.com/libp2p/rust-libp2p" }}", rev = "{{ $.Env.GitReference }}", default_features = false, features = [ "websocket", "mplex", "yamux", "tcp-async-io", "ping", "noise", "dns-async-std" ], optional = true}
+          {{ .CargoFeatures }} = {package = "libp2p", git = "https://{{ or $.Env.GitTarget "github.com/libp2p/rust-libp2p" }}", rev = "{{ $.Env.GitReference }}", default_features = false, features = [ "websocket", "mplex", "yamux", "tcp-async-io", "ping", "noise", "dns-async-std", "rsa" ], optional = true}
           """
   {{ end }}
   {{ end }}
--- a/ping/_compositions/go-rust-interop.toml
+++ b/ping/_compositions/go-rust-interop.toml
@@ -133,7 +133,7 @@
         CARGO_FEATURES = '{{ .CargoFeatures }}'
         CARGO_REMOVE = '{{ .CargoFeatures }}'
         CARGO_PATCH = """
-          {{ .CargoFeatures }} = {package = "libp2p", git = "https://{{ or $.Env.GitTarget "github.com/libp2p/rust-libp2p" }}", rev = "{{ $.Env.GitReference }}", default_features = false, features = [ "websocket", "mplex", "yamux", "tcp-async-io", "ping", "noise", "dns-async-std" ], optional = true}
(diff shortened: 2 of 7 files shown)


BigLep:
@mxinden : do we have a corresponding step in the rust-libp2p release process to remember to do this or to verify this was done?

mxinden:
@BigLep at this point we don't and I would argue that in itself is not a big issue. The only thing that is happening after a release is that the rust-libp2p CI fails. It does not affect the go-libp2p CI.

I have to give this more thought. Ideally this step would not be required in the first place. See also https://github.com/libp2p/rust-libp2p/pull/2813#issuecomment-1242700115

mxinden:
@laurentsenta can you take a look at the test failure? I am not sure what is the reason for it. In case you know it already, great. If not, I can take a more in-depth look.

BigLep:
@mxinden : I don't know all that involved here but the reasons I chimed in are:
1. I was under the impression that we aren't running the fully compatibility test-plan as part of each PR but rather a lighter version.  As a result, I assume before we do a release we ensure the full compatibility test-plan passes.
2. I think we want it so that any maintainer could be shuffled in an know how to execute a release.  I assume there are are a set of manual steps including updating version numbers some places (great if that's automated though), making public announcements, etc.  That should be standardized so anyone can know how to do it and not miss anything.

mxinden:
> I was under the impression that we aren't running the fully compatibility test-plan as part of each PR but rather a lighter version. As a result, I assume before we do a release we ensure the full compatibility test-plan passes.

If I am not mistaken, we are running the full test suit on each rust-libp2p pull request. (Which I am in favor of.)

> I think we want it so that any maintainer could be shuffled in an know how to execute a release. I assume there are are a set of manual steps including updating version numbers some places (great if that's automated though), making public announcements, etc. That should be standardized so anyone can know how to do it and not miss anything.

:+1: agree with this long-term. On documenting this specific release step, i.e. bumping the versions in `libp2p/test-plans`, I suggest we think about whether we can get rid of it first. All of this is relatively new, thus there might be room for improvement.

On the general matter, i.e. enabling folks other than Max to make rust-libp2p releases, this is related to https://github.com/libp2p/rust-libp2p/issues/2902 and needs some focus. Though I am not sure how to prioritize it with libp2p Day and the WebRTC efforts. Thoughts?

laurentsenta:
Thanks for the bump @mxinden on it,

The plan:

1. Fix the master build error to get rid of the current CI-wide error (https://github.com/libp2p/test-plans/issues/43)
	- Cause: we disabled testing `master` in https://github.com/libp2p/test-plans/pull/37, but we still let it break the build in the `Cargo.toml`
2. Then merge the write-artifact patch
	- https://github.com/libp2p/test-plans/pull/40
	- That should remove the error you got here (the run step timeouts because we rebuild too much code during the run step, which timeouts).
3. Which mean we can merge this PR (https://github.com/libp2p/test-plans/pull/41)
4. Which mean we can merge https://github.com/libp2p/test-plans/pull/42

laurentsenta:
On the larger discussion regarding what we tests, releases process, etc:

@BigLep is right with:

> I was under the impression that we aren't running the fully compatibility test-plan as part of each PR but rather a lighter version. As a result, I assume before we do a release we ensure the full compatibility test-plan passes.

With https://github.com/libp2p/rust-libp2p/pull/2835, we enabled the cross version test and the _**"small"**_ cross-implementation test. The "small" tests uses latest versions only (and takes ~10 minutes).

The big test takes >30 minutes, my plan is to propose this as a step in (go|rust)-libp2p release processes.


You two have different info, sorry about that: we talked about it during a 1:1 with Steve, then I went on break. I shared a few notes below, but I haven't had time to properly implement and communicate these with you @mxinden.

- https://github.com/libp2p/test-plans/issues/31#issuecomment-1238296965 (last, "Follow-Ups" part)
- https://pl-strflt.notion.site/libp2p-testing-story-5c49c8f79137416abdd61b091bc87a9f (steps 3 and 5)

laurentsenta:
@mxinden, I took the liberty to rebase over the latest master, please feel free to review and merge.

mxinden:
> please feel free to review and merge

I will go ahead here. Thanks @laurentsenta for the help!

PR #42, opened 2022-09-17
ping/rust: adapt to rust-sdk patch, remove unused dependencies

elenaf9:
- 01d0b18 adapts to an breaking api change in the rust-sdk (https://github.com/testground/sdk-rust/pull/36) which causes the build to fail after running `cargo update`.
- e9d601b removes unused dependencies. Was there a reason why they were added in the first place?
- 76a9ee2: fix instructions for importing the testplans in testground. Without the `--from` the testground CLI did not accept the current instruction described in the README. Does it work for other folks / did I do something wrong? 

Builds on #41.

Issue #35, opened 2022-08-30
EPIC: interop testing for go and rust libp2p - Ping Test


laurentsenta:
We want to run a "simple" test (ping) that verifies interoperability between versions and implementations.
This test should run at least before every (go|rust) libp2p release.

Use case:

- Test cross-version compatibility for Go
- Test cross-version compatibility for Rust
- Test cross-implementation compatibility for Rust and Go

We're aiming for:

- a single ping test implemented in different languages (go + rust)
- running interop + cross version for every PR on the go-libp2p repo
- running interop + cross version for every PR on the rust-libp2p repo

- @mxinden and @marten-seemann agreed to run this test if
  - it takes < 30 minutes
  - it's stable (aiming for < 1% failure rate)
- they want "reasonable" run time in the future
  - my goal right now is < 15 minutes which is the go test runtime on ubuntu,
  - I'm trying to push it down to ~ 5 minutes with caching. There is a O(n) relationship between build time & the number of tested versions and caching would make it O(1).
  - we're going to timebox this effort for a week, then discuss how far we want to go.

## Todo

- [x] implement cross-version interop testing in go
  - https://github.com/testground/testground/issues/1340
  - https://github.com/libp2p/test-plans/pull/23
- [x] implement cross-version interop testing in rust
  - https://github.com/libp2p/test-plans/pull/26
- [x] implement cross-implementation interop testing
  - https://github.com/libp2p/test-plans/pull/26
- [x] maintain & enables tests in go and rust repos
  - https://github.com/libp2p/rust-libp2p/pull/2835
  - https://github.com/libp2p/go-libp2p/pull/1725
- [ ] stability & perf improvements
  - https://github.com/libp2p/test-plans/pull/34
  - https://github.com/testground/testground/pull/1407
  - https://github.com/testground/testground/issues/1401
  - https://github.com/libp2p/test-plans/pull/32
  - https://github.com/libp2p/test-plans/issues/31
- [ ] Sync with the libp2p/interop repo https://github.com/libp2p/interop/issues

## References:

- https://www.notion.so/pl-strflt/Roadmap-Draft-c6b7a9e67b18468bac16cccb9a8bf8fe
- https://www.notion.so/pl-strflt/Testground-Interop-Features-Plan-with-Stability-in-Mind-dad6b6d25b79451ba01e0a28da1888b3

laurentsenta:
Gathering content in a single issue to share w/Bloxico.

laurentsenta:
[...]

PR #45, opened 2022-09-21
ping/rust: remove the master dependency

### LINKED ISSUE #43: rust-libp2p build breaks on master's update
the worfklow started to fail in go-libp2p, here is an example of a run:
https://github.com/libp2p/go-libp2p/actions/runs/3047048099/jobs/4910594908

Raised by @marten-seemann 

We need to investigate, my first intuition:

Now that we merged: https://github.com/libp2p/test-plans/pull/37
We are using "latest" release and not master.

We expect that when rust release a new version and updates `master`, that won't break other tests (because we only test releases known by `libp2p/test-plans`).

BUT we do use a `master` definition in the Cargo file, this is how we deal with multiple versions:
https://github.com/libp2p/test-plans/blob/520162a1881e7a8196728c3f979debd330cec328/ping/rust/Cargo.toml#L27
That line will break if the package is bumped to version 0.49 because it will try to load the repo and won't find the 0.48 version.

We might need to drop this line and use a rewrite in the `master` group.

laurentsenta:
Fixes #43 

Problem definition:
- We want to give the `rust-libp2p` repository the ability to build and test their `master` branch (which is also used when testing fork and PRs).
- We want to protect other client-repositories (such as `go-libp2p`) from failure due to unchecked changes in the `rust-libp2p` repository.

The case we've seen with #43:

- A release happens in `rust-libp2p`,
- The version is bumped in the `master` branch
- Now the line `libp2pv0480 = { git= ... branch = master, version = 0.48.0}` breaks because the version cannot be found.

This is a valid change (the `rust-libp2p` didn't break the interop contract, they "just" updated versions) which broke the `go-libp2p` repository.

## Root cause

The way we implement compatibility/features flags with rust has two constraints:

- we can't use "official" dependencies rewrites because rust libp2p uses feature flags. combining the two is not supported in cargo.
- even though a composition might build and use only versions `rust-1` and `rust-2`, the `Cargo.toml`, which contains every version, including `master`, has to be valid.

### DIFF
--- a/ping/rust/Cargo.toml
+++ b/ping/rust/Cargo.toml
@@ -24,4 +24,5 @@ libp2pv0440 = {package = "libp2p", default_features = false, features = [ "webso
 libp2pv0450 = {package = "libp2p", default_features = false, features = [ "websocket", "mplex", "yamux", "tcp-async-io", "ping", "noise", "dns-async-std" ], version = "0.45.0", optional = true}
 libp2pv0460 = {package = "libp2p", default_features = false, features = [ "websocket", "mplex", "yamux", "tcp-async-io", "ping", "noise", "dns-async-std" ], version = "0.46.0", optional = true}
 libp2pv0470 = {package = "libp2p", default_features = false, features = [ "websocket", "mplex", "yamux", "tcp-async-io", "ping", "noise", "dns-async-std" ], version = "0.47.0", optional = true}
-libp2pv0480 = {package = "libp2p", git = "https://github.com/libp2p/rust-libp2p", branch = "master", default_features = false, features = [ "websocket", "mplex", "yamux", "tcp-async-io", "ping", "noise", "dns-async-std" ], version = "0.48.0", optional = true}
+
+libp2pmaster = {package = "libp2p", git = "https://github.com/libp2p/rust-libp2p", branch = "master", default_features = false, features = [ "websocket", "mplex", "yamux", "tcp-async-io", "ping", "noise", "dns-async-std" ], optional = true}
--- a/ping/rust/Cargo.toml
+++ b/ping/rust/Cargo.toml
@@ -13,7 +13,7 @@ rand = "0.8"
 serde = {version = "1", features = ["derive"]}
 serde_json = "1"
 soketto = "0.7.1"
-testground = {git = "https://github.com/testground/sdk-rust", branch = "master", version = "0.4.0"}
+testground = {git = "https://github.com/testground/sdk-rust", rev = "94a9a72796f94cc7ca786a5f019d07f328c76d4b", version = "0.4.0"}
 thiserror = "1"
 tokio = { version = "1", default-features = false, features = ["sync", "rt-multi-thread", "macros", "net"] }
 tokio-stream = { version = "0.1", default-features = false, features = [] }
--- a/ping/rust/src/main.rs
+++ b/ping/rust/src/main.rs
@@ -11,8 +11,8 @@ use testground::network_conf::{
 };
 
 pub mod libp2p {
-    #[cfg(all(feature = "libp2pv0480",))]
-    pub use libp2pv0480::*;
+    #[cfg(all(feature = "libp2pmaster",))]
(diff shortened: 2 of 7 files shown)


laurentsenta:
This issue got me thinking about the need for a more permanent fix, notes in https://github.com/libp2p/test-plans/issues/35#issuecomment-1254991985

ping/rust/src/main.rs
```diff
@@ -11,8 +11,8 @@ use testground::network_conf::{
 };
 
 pub mod libp2p {
-    #[cfg(all(feature = "libp2pv0480",))]
-    pub use libp2pv0480::*;
+    #[cfg(all(feature = "libp2pmaster",))]
+    pub use libp2pmaster::*;
 
     #[cfg(all(feature = "libp2pv0470",))]
     pub use libp2pv0470::*;
```
laurentsenta, line 18: (creating here to get a thread)

@mxinden, I'm trying to catch other cases where the build might break after a push to `master`.

1. If `rust-libp2p` remove a feature flag in master, I guess the `libmaster` line in `Cargo.toml`will become invalid, and it will break the build. Correct?
2. `rust-libp2p` publishes an API breaking change in `master`, could it break the build of previous versions? I think the decoration `#[cfg(all(feature = "libp2pmaster",))]` completely removes the code.

mxinden: > If `rust-libp2p` remove a feature flag in master, I guess the `libmaster` line in `Cargo.toml`will become invalid, and it will break the build. Even for old versions, Correct?

Correct. Though this will change soon where we can replace all the features with a single `full`. See https://github.com/libp2p/rust-libp2p/pull/2918.

mxinden: > if `rust-libp2p` publishes an API breaking change in `master`, could it break the build of previous versions? I assume the decoration `#[cfg(all(feature = "libp2pmaster",))]` completely removes the code and we're safe here.

I am not sure I follow. Once a crate is published to crates.io it can **never** be changed again.

mxinden (approved):
From what I can tell this looks good to me. I hope we can find a more elegant fix in the future, though I think going with this is worth it given that it unblocks the status quo.

ping/rust/src/main.rs
laurentsenta: Apologies, I mixed different concepts of versions,

I'm wondering if there are other cases where a push to rust-libp2p master branch can break the build for previous versions of the ping test (`cargo build --feature libp2pv0470`, for example).

- We fixed one case with this PR (when we push a version bump),
- We know another case (when we change feature flags),
- Do you see other cases?

Trying to play around with macro it looks like it's be enough, but I might miss some rust's edge cases.

```rust
// cargo build --features libp2pmaster (FAIL)
// cargo build --features libp2pv2 (OK)
fn main() {
    println!("got: {}", s() + 1)
}

#[cfg(all(feature = "libp2pmaster",))]
fn s() -> String {
    return String::from("new master API that breaks the current helloworld test");
}

#[cfg(all(feature = "libp2pv2",))]
fn s() -> i32 {
    return 2;
}
```

ping/rust/Cargo.toml
```diff
@@ -13,7 +13,7 @@ rand = "0.8"
 serde = {version = "1", features = ["derive"]}
 serde_json = "1"
 soketto = "0.7.1"
-testground = {git = "https://github.com/testground/sdk-rust", branch = "master", version = "0.4.0"}
+testground = {git = "https://github.com/testground/sdk-rust", rev = "94a9a72796f94cc7ca786a5f019d07f328c76d4b", version = "0.4.0"}
```
elenaf9, line 16: Just saw this change while working on #42. Was this a temporary change because until #42 is merged we are not compatible with the master branch, or should I also use a concrete revision there?

mxinden: I just released `testground` `v0.4.0`. How about using that instead?

mxinden: https://github.com/testground/sdk-rust/pull/38

PR #46, opened 2022-09-25
ping/rust/src/main: Decrease ping interval

mxinden:
This will speed up test runs.

What do folks think?

### DIFF
--- a/ping/rust/src/main.rs
+++ b/ping/rust/src/main.rs
@@ -52,7 +52,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             development_transport(local_key).await?,
             ping::Behaviour::new(
                 ping::Config::new()
-                    .with_interval(Duration::from_secs(10))
+                    .with_interval(Duration::from_secs(1))
                     .with_keep_alive(true),
             ),
             local_peer_id,


ping/rust/src/main.rs
```diff
@@ -52,7 +52,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             development_transport(local_key).await?,
             ping::Behaviour::new(
                 ping::Config::new()
-                    .with_interval(Duration::from_secs(10))
+                    .with_interval(Duration::from_secs(1))
```
laurentsenta, line 55: Got it, it looks like this timeout was quite large to prevent ping signals from getting mixed up between iterations,

If we lower this threshold, we increase the probability that `node-rust-A` sends 2 pings during iteration one, which `go-rust-B` might interpret as one ping from iteration 1 and one ping during iteration 2,

Correct?

elenaf9: Would it be a problem if that happens?

mxinden: Nodes will only reconfigure their network once they received a ping from everyone *and* everyone received a ping from everyone else. In other words, they move in synchrony via the sync service. Thus a node should never receive a ping from another node that is still in a different round.

Does the above make sense? Am I missing something?

https://github.com/libp2p/test-plans/blob/a9292a67d280ecc48c668c1f7ac95867b603523e/ping/rust/src/main.rs#L216-L235

laurentsenta: It makes sense, thanks for the clarification!
The patch looks good to me, please feel free to merge.

> Would it be a problem if that happens?

Not a big one, I'd recommend dropping network configuration between iterations if we're not using them consistently.

laurentsenta: Feel free to ignore, I might be in deep dunning-krugger-syndrome,
this is the scenario that itches my mind: 

- We're in iteration 1,
- For some reason, `go` is slow (big test running on light machine, or just configured a slow data network)
- `rust` sends ping(1) to  `go` ([code](https://github.com/libp2p/test-plans/blob/96c70ff6dfb5eee2b5a4366e6fcd9eeeb0643e19/ping/rust/src/main.rs#L54)),
- `rust` waits for the response ([code](https://github.com/libp2p/test-plans/blob/96c70ff6dfb5eee2b5a4366e6fcd9eeeb0643e19/ping/rust/src/main.rs#L201))
- after 1 second, `rust` sends ping(2) to `go`,
- `go` replies to ping(1),
- `go` signals the end of iteration 1 (control network),
- `rust` registers the ping response ([code](https://github.com/libp2p/test-plans/blob/96c70ff6dfb5eee2b5a4366e6fcd9eeeb0643e19/ping/rust/src/main.rs#L209)),
- `rust` signals the end of iteration 1 ([code](https://github.com/libp2p/test-plans/blob/96c70ff6dfb5eee2b5a4366e6fcd9eeeb0643e19/ping/rust/src/main.rs#L228)) & exits the `signal_and_wait`,
- `go` replies to ping(2),
- We're in iteration 2,
- `rust` registers the ping response (sent from iteration 1) ([code](https://github.com/libp2p/test-plans/blob/96c70ff6dfb5eee2b5a4366e6fcd9eeeb0643e19/ping/rust/src/main.rs#L209)),

laurentsenta (approved):

ping/rust/src/main.rs
elenaf9: At least in rust this can not happen. The timer to trigger a ping is only reset once the previous ping has resolved: https://github.com/libp2p/rust-libp2p/blob/04b07c4d14a0a6ec350fc21261a33d859fc71289/protocols/ping/src/handler.rs#L333-L364
But I don't know how it is implemented in go.

laurentsenta: I was wrong, thanks for sharing the explanation!

PR #47, opened 2022-09-25
ping/rust/src/main: Reduce noise on logs

mxinden:
- Don't log every `SwarmEvent`.
- Log connected and ping only once per peer.
- Log format somewhat consistent with Golang implementation.


laurentsenta (approved):

elenaf9 (approved):
Why are all logs from the rust client shown as `ERROR` in the CI log?

mxinden:
> Why are all logs from the rust client shown as `ERROR` in the CI log?

Oh, sorry, just saw this comment @elenaf9.

As far as I know they are shown as `ERROR` given that rust-libp2p logs to stderr and Testground thus interprets that as an error. That said, I don't know what the go-sdk does.

PR #48, opened 2022-09-28
ping/rust/Dockerfile: Use correct feature when building deps

mxinden:


ping/rust/Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN cd ./plan/ && \
         cargo update --dry-run && \
         cargo fetch && \
         # use a default feature to rely on docker caching.
-        cargo build --release --features "libp2pv0450"
+        cargo build --release --features="${CARGO_FEATURES}"
```
laurentsenta, line 17: Some context,

We set a default feature on purpose, this line is used to build and cache dependencies before doing the real build below.

If you use a different `CARGO_FEATURES`, the layer won't be cached between versions, build time will increase a lot. It's really similar to the other [Dockerfiles we used](https://github.com/testground/sdk-rust/blob/fa00058a8f2c7850b712c0143b4e723bbf189535/Dockerfile#L4-L11) but this one takes into account feature flags.

mxinden: Understood. Thanks for clarifying @laurentsenta.

mxinden, line 17:
```suggestion
        cargo build --release --features="libp2pv0480"
```

How about we bump the version to a more recent one instead of taking the second to oldest one? My hope would be an increase in overlap. Just an idea. Feel free to close @laurentsenta.

mxinden:
Will close here. Needs more investigation on whether it is actually worth it.

PR #49, opened 2022-09-29
Adding a webrtc test

John-LittleBearLabs:


John-LittleBearLabs:
Doesn't look like I can add reviewers, likely related to the target repo's rules.
@GlenDC @ckousik @ddimaria 

I don't think I should switch to "ready" until I can stop pointing at Chinmay's repository (i.e. after webrtc lands in go-libp2p). Does that sound right to you folks?

webrtc/_compositions/go2go.toml
```diff
@@ -0,0 +1,64 @@
...
+    build_args = { image = "golang:1.19-bullseye" }
+    modfile = "go.orig.mod"
+
+  [groups.run.test_params]
+    expected_implementation = "go"
+
+  [groups.build_config.dockerfile_extensions]
+    pre_build = '''
+RUN cd ${PLAN_DIR} && go get github.com/minio/sha256-simd@v1.0.0
+RUN cd ${PLAN_DIR} && go get github.com/prometheus/client_golang/prometheus@v1.13.0
```
John-LittleBearLabs, line 33: These go gets are one of the weirdnesses I was discussing in standup.

webrtc/_compositions/go2go.toml
```diff
@@ -0,0 +1,64 @@
...
+RUN cd ${PLAN_DIR} && go mod tidy
+'''
+#TODO change url & branch here and below
+    pre_mod_download = '''
+RUN git clone --single-branch --branch ckousik/webrtc https://github.com/ckousik/go-libp2p.git /go-libp2p/
```
John-LittleBearLabs, line 38: These clones are the other.

webrtc/_compositions/go2go.toml
```diff
@@ -0,0 +1,64 @@
+[metadata]
+  name = "go2go {{ .Env.GitReference }}"
```
GlenDC, line 2: Is that something they do? I was for now just giving it a static name

webrtc/_compositions/go2go.toml
```diff
@@ -0,0 +1,64 @@
...
+  builder = "docker:go" # override builder here
+
+  [groups.build_config]
+    path = "./go"
+    build_base_image = 'golang:1.19-buster'
+    build_args = { image = "golang:1.19-bullseye" }
+    modfile = "go.orig.mod"
+
+  [groups.run.test_params]
+    expected_implementation = "go"
```
GlenDC, line 28: this parameter was copied over, no? But not really that relevant for your test I think?

webrtc/_compositions/go2go.toml
GlenDC: I think it's fine to keep this TODO for no

webrtc/go/.gitignore
```diff
@@ -0,0 +1,5 @@
+# we might ln -s these files to help during development,
+# but we don't want to commit these or the testground setup
+# will break (the mod file will overwrite the custom versions).
+/go.sum
```
GlenDC, line 4: Is this a known issue for them, or expected? Because not having Go mod files to dos make IDE's not very happy in regular setups.

webrtc/_compositions/go2go.toml
```diff
@@ -0,0 +1,64 @@
...
+
+  [global.build_config]
+    enable_go_build_cache  = false # see https://github.com/testground/testground/issues/1361
+    go_proxy_mode          = "remote"
+    go_proxy_url           = "https://proxy.golang.org"
+
+[[groups]]
+  id = "server"
+  instances = { count = 1 }
+  builder = "docker:go" # override builder here
```
GlenDC, line 19: no need to define, right? As it's already `docker:go` by default

webrtc/_compositions/go2go.toml
```diff
@@ -0,0 +1,64 @@
...
+'''
+#TODO change url & branch here and below
+    pre_mod_download = '''
+RUN git clone --single-branch --branch ckousik/webrtc https://github.com/ckousik/go-libp2p.git /go-libp2p/
+'''
+
+[[groups]]
+  id = "client"
+  instances = { count = 1 }
+  builder = "docker:go" # override builder here
```
GlenDC, line 44: no need to define, right? As it's already `docker:go` by default

webrtc/go/.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
GlenDC, line 1: You probably do not want to commit the `.idea/` dir and its files, no?

webrtc/go/go.orig.mod
```diff
@@ -0,0 +1,126 @@
...
+
+require (
+	github.com/libp2p/go-libp2p v1.0.5
+	github.com/libp2p/go-libp2p-core v0.20.1
+	github.com/testground/sdk-go v0.3.0
+	golang.org/x/sync v0.0.0-20220819030929-7fc1605a5dde
+)
+
+require (
+	github.com/avast/retry-go v2.6.0+incompatible // indirect
```
GlenDC, line 13: That are really a lot of direct requirements o.O Looks more like a `go.sum` file (length wise) than a go.mod file. You do not rely on these directly, right? So why mention them all here instead of only the ones you're importing?

webrtc/go/main.go
```diff
@@ -0,0 +1,312 @@
...
+				continue
+			}
+
+			id := ai.ID // capture the ID locally for safe use within the closure.
+
+			g.Go(func() error {
+				// a context for the continuous stream of webrtcs.
+				pctx, cancel := context.WithCancel(gctx)
+				defer cancel()
+				res := <- ping.Ping(pctx, id)
```
GlenDC, line 193: This is a naked channel block. You do not want to pair it with something that can time out or w/e?

webrtc/go/manifest.toml
```diff
@@ -0,0 +1,22 @@
...
+[builders]
+"exec:go"   = { enabled = true }
+
+[builders."docker:go"]
+enabled                = true
+build_base_image       = "golang:1.19-bullseye"
+
+[runners]
+"local:docker"  = { enabled = true }
+"local:exec"    = { enabled = true }
```
GlenDC, line 12: why is `local:exec` and `cluster:k8s` enabled?

webrtc/go/manifest.toml
```diff
@@ -0,0 +1,22 @@
...
+"cluster:k8s"   = { enabled = true }
+
+[[testcases]]
+name = "ping"
+instances = { min = 1, max = 10000, default = 1 }
```
GlenDC, line 17: why a max of `10000`? How do you ever expect such numbers to be reached?

GlenDC (changes requested):

webrtc/_compositions/go2go.toml
John-LittleBearLabs: Yes that came from the existing one. That said, the existing one is doing something I'm not - they're iterating over _different_ versions of each codebase and testing their interoperability against each other.

That's not a bad idea to do... but since I haven't done it yet that extra tagging is unnecessary complexity for now.

webrtc/_compositions/go2go.toml
John-LittleBearLabs: Probably not relevant, I'd guess. But I'm not really sure, to be honest.

webrtc/go/.gitignore
John-LittleBearLabs: Definitely a known issue - that text including the comment is ALL copied from an existing test (libp2p/test-plans/ping).

It was definitely a point of confusion for me.

webrtc/_compositions/go2go.toml
John-LittleBearLabs: Yes you're probably right about that. "by default" meaning it gets picked up from the plan, though, right?

webrtc/go/.idea/.gitignore
John-LittleBearLabs: No probably not. Removing.

webrtc/go/go.orig.mod
John-LittleBearLabs: There are 4 direct requirements.  I'm 100% sure 2 of them are necessary. I guess I could experiment with removing the other 2 if you'd like.

The `// indirect` stuff was generated by... I think `go mod tidy` maybe?

webrtc/go/main.go
John-LittleBearLabs: If it gets stuck here, the test itself will eventually time out (by order of testground daemon), so I'm not terribly worried about it. But if you think it's important I'm open to the idea.

webrtc/go/manifest.toml
John-LittleBearLabs: Didn't think about it too much, but is there any particular reason this test couldn't be run in those circumstances?

webrtc/go/manifest.toml
John-LittleBearLabs: The thing I changed here was to lower the default.

I would guess if someone asked for 10000 they're actually expecting the system to break and want to see what happens when it does. Though I don't really know.

GlenDC: Hmmm ok. Seems odd, but plausible.

webrtc/go/manifest.toml
GlenDC: I would think that local:exec needs a different builder, the k8s one I do not know

webrtc/go/main.go
GlenDC: Dunno, might be more useful to have a decent helpful error that you can explicitly throw, than some vague testdaemon timeout

webrtc/go/go.orig.mod
GlenDC: well the only ones required are the ones you import yourself. The indirect stuff is indeed possible, bit annoying though, but I guess it is what it is :P

webrtc/_compositions/go2go.toml
GlenDC: It's because you define it as the default plan in `[global]`

webrtc/_compositions/go2go.toml
GlenDC: These parameters are things you use to pass info at start-up time to your test, in order to perhaps configure something. So if you are not doing something with the `expected_implementation` param, you do not need it. These variables are in the go sdk accessed using the methods from https://pkg.go.dev/github.com/testground/sdk-go@v0.3.0/runtime#RunParams ( e.g. `StringParam`)

webrtc/_compositions/go2go.toml
John-LittleBearLabs: Yeah I think what happened here is the original mixed & matched different implementations. So it couldn't be just the global setting. 
But since at the moment I'm _only_ dealing with go, I could drop this.

mxinden:
Wonderful to see this happening. Thanks @John-LittleBearLabs for working your way through this stack. Thanks @GlenDC for the review.

Today we only test the TCP transport via the [ping](https://github.com/libp2p/test-plans/tree/master/ping) tests. Long term we will test:

- TCP
- QUIC
- WebRTC
- WebTranport
- ...

With the above in mind, I think it makes sense to think of this pull request not as a one-off, but as a pull request that introduces an abstraction to https://github.com/libp2p/test-plans that enables us to test a variety of transports.

I haven't put enough thoughts into the ideal abstraction, thus I am happy for input. What I have today:

- Extend the [ping test](https://github.com/libp2p/test-plans/tree/master/ping) with an additional `transport` parameter (we have `max_latency_ms` and `iterations` today). For now `transports` could be one of `tcp` and `webrtc`.
- Depending on the `transport` parameter value, listen on either `tcp` or `webrtc`.
- Extend the [composition files](https://github.com/libp2p/test-plans/tree/master/ping/_compositions) to run with both `tcp` and `webrtc`. I would appreciate input from @laurentsenta here on how to best do this. Templating? Separate composition files?

John-LittleBearLabs:
@mxinden per your comment, `ping/_compositions/go-webrtc.toml` sets the `transport` flag to "webrtc" (the default is "tcp").

A couple complications:
* go.toml - the versions of go that are compared under go-cross-versions.toml - none of these versions have the webrtc transport in them. So I added recent.go-toml with basically just the one version that does (and I think that file should probably be renamed).
* The "version" that _does_ have webrtc is in Chinmay's fork, which is why I consistently refer to that "version" as "ckousik". And the modfile & composition file still have some wonkiness to deal with that oddity.

mxinden:
Thanks @John-LittleBearLabs for looking into this!

> A couple complications:
> 
>     * go.toml - the versions of go that are compared under go-cross-versions.toml - none of these versions have the webrtc transport in them. So I added recent.go-toml with basically just the one version that does (and I think that file should probably be renamed).
> 
>     * The "version" that _does_ have webrtc is in Chinmay's fork, which is why I consistently refer to that "version" as "ckousik". And the modfile & composition file still have some wonkiness to deal with that oddity.

This sparked the following proposal https://github.com/libp2p/test-plans/issues/53. We would define the supported transports per version. If I am not mistaken, this should resolve the issues above. Does that sound reasonable? //CC @laurentsenta

John-LittleBearLabs:
@mxinden 
> Does that sound reasonable?

Yes I do see how that approach does solve this. I'll think about this for a bit, and then I think I'll have a question or two but should probably put it in that other PR.

p-shahi:
@John-LittleBearLabs Should this be closed in favor of #67 ?

John-LittleBearLabs:
@p-shahi 

> @John-LittleBearLabs Should this be closed in favor of #67 ?

This one has the advantage that the last time I tried to run it it worked.

That said, neither of these should get merged in while they still depend on Chinmay's private repo. 
Once https://github.com/libp2p/go-libp2p/pull/1655 is resolved & merged & released, I'd have to update one or both of these. Might as well only update the one that's more in line with the future vision.

So yeah, let's close this.

PR #50, opened 2022-09-29
ping/rust/Dockerfile: Build in debug mode instead of release mode

mxinden:
How about we build in debug mode **for now**.

```
+ I expect this to significantly reduce build times.
+ I doubt it slows down execution time. 
- In production, people run in release mode.
```

### DIFF
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -14,7 +14,7 @@ RUN cd ./plan/ && \
         cargo update --dry-run && \
         cargo fetch && \
         # use a default feature to rely on docker caching.
-        cargo build --release --features "libp2pv0450"
+        cargo build --features "libp2pv0450"
 
 ARG CARGO_PATCH=""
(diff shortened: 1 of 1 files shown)


laurentsenta (approved):
looks good to me, feel free to merge.

I don't think you need benchmarking here (it'll take some time to setup and run), but I created an EPIC to make sure we'll share these.

mxinden:
CI continuously runs into the below error.

> task of type build cancelled: docker build failed: ApplyLayer exit status 1 stdout:  stderr: write /go/pkg/mod/github.com/klauspost/compress@v1.15.9/zstd/testdata/decoder.zip: no space left on device

I am assuming debug builds are larger than release builds and thus decompression fails?@laurentsenta any ideas? Otherwise I will take another look.

mxinden:
@laurentsenta would you mind taking a look here?

mxinden:
Merging here with CI all green.

PR #60, opened 2022-10-20
Update for rust-libp2p 0.49 release

### LINKED ISSUE #59: ping/_compositions/rust: Add `v0.49.0` and update features of `master`
With https://github.com/libp2p/rust-libp2p/pull/3001 `master` is now `v0.50.0` AND removes the `dns-async-std` etc. features.

This results in test failures, see e.g. https://github.com/libp2p/rust-libp2p/pull/3047.

We need to do the upgrade of `libp2p/test-plans` similar to https://github.com/libp2p/test-plans/pull/41.

Most prominently, we need to update the versions below:

https://github.com/libp2p/test-plans/blob/c71602e88c8a20c6ecc7c18fe1f5b5cf570d3b11/ping/_compositions/rust.toml#L1-L9

and update the `Cargo.toml.

//CC the rust-libp2p team @thomaseizinger @jxs @elenaf9 for visibility and in case one of you wants to pick this up.

thomaseizinger:
I am not sure if I did all of this right. I was a bit confused as to why there is seemingly so much duplication between the various `toml` files and whether they all need updating or not.

Hoping that CI here will tell me whether this works :D

Resolves #59.

### DIFF
--- a/ping/_compositions/go-rust-interop-latest.toml
+++ b/ping/_compositions/go-rust-interop-latest.toml
@@ -93,7 +93,7 @@
         CARGO_FEATURES = '{{ .CargoFeatures }}'
         CARGO_REMOVE = '{{ .CargoFeatures }}'
         CARGO_PATCH = """
-          {{ .CargoFeatures }} = {package = "libp2p", git = "https://{{ or $.Env.GitTarget "github.com/libp2p/rust-libp2p" }}", rev = "{{ $.Env.GitReference }}", default_features = false, features = [ "websocket", "mplex", "yamux", "tcp", "ping", "noise", "dns", "async-std", "rsa" ], optional = true}
+          {{ .CargoFeatures }} = {package = "libp2p", git = "https://{{ or $.Env.GitTarget "github.com/libp2p/rust-libp2p" }}", rev = "{{ $.Env.GitReference }}", default_features = false, features = [ "websocket", "mplex", "yamux", "ping", "noise", "rsa" ], optional = true}
           """
   {{ end }}
   {{ end }}
--- a/ping/_compositions/go-rust-interop-latest.toml
+++ b/ping/_compositions/go-rust-interop-latest.toml
@@ -73,7 +73,7 @@
 
     [groups.build_config]
       path = "./rust/"
-      
+
     [groups.build_config.build_args]
       CARGO_FEATURES = '{{ .CargoFeatures }}'
   {{ end }}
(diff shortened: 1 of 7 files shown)


thomaseizinger:
I don't quite understand how this works. Are we properly isolating the builds for the various versions so the features are not conflicting with each other? I can't figure out how to tell testground that it should use one feature set for one version and another for another version.

Without tesground, the way I'd do this is by having a matrix in GitHub actions such that the builds are isolated. The approach of having a single manifest file with multiple dependencies seems odd to me. We might just need to accept the duplication between multiple instances of the `rust` ping composition where each crate has its own dependency and therefore its own buildgraph. Once it is built for one version of `rust-libp2p`, it shouldn't need touching again unless the testground Rust crate changes so the maintenance effort doesn't seem to be too bad?

Anyway, just my 2c on interacting with this for the first time :)

thomaseizinger:
Okay, I think I found the curlpit. The `Cargo.lock` wasn't updated :roll_eyes: 

I still think the way `ping/rust` is currently structured is a bit dodgy :sweat_smile: 

```rust
pub mod libp2p {
    #[cfg(all(feature = "libp2pmaster",))]
    pub use libp2pmaster::*;

    #[cfg(all(feature = "libp2pv0480",))]
    pub use libp2pv0480::*;

    #[cfg(all(feature = "libp2pv0470",))]
    pub use libp2pv0470::*;

    #[cfg(all(feature = "libp2pv0460",))]
    pub use libp2pv0460::*;

    #[cfg(all(feature = "libp2pv0450",))]
    pub use libp2pv0450::*;

    #[cfg(all(feature = "libp2pv0440",))]
    pub use libp2pv0440::*;
}
```

This is just asking to break with changes like https://github.com/libp2p/rust-libp2p/pull/2859. Here we deprecated the `with_keep_alive` builder on the ping protocol. The mitigation is to compose it with a `keep_alive::Behaviour` but that only exists from `0.49` onwards.

I think we should split `ping/rust` into one crate per libp2p version and try to cut down the duplication through other means. We can probably create a workspace local crate with some shared functionality. I reckon we are going to need that anyway once we test more things in testground.

thomaseizinger:
I think I actually got it now haha

Seems to be passing.

mxinden:
Thanks for the work here @thomaseizinger! Impressive how fast you got started on this.

I added 06b6527. If I am not mistaken, `tcp` and `async-std` and `dns` were missing. As far as I can tell, CI was still green as we don't test the git ref path here. As we are not triggering the workflows from rust-libp2p or go-libp2p, the workflow doesn't know which git reference to use.

mxinden (approved):
:pray:

thomaseizinger:
> I added 06b6527. If I am not mistaken, `tcp` and `async-std` and `dns` were missing.

I had moved them to the actual dependency that is declared in `Cargo.toml` but I guess that bit is actually `sed`-ed out?

mxinden:
> > I added [06b6527](https://github.com/libp2p/test-plans/commit/06b652727be19b9e805701e55cd6e26b1ef1113a). If I am not mistaken, `tcp` and `async-std` and `dns` were missing.
> 
> I had moved them to the actual dependency that is declared in `Cargo.toml` but I guess that bit is actually `sed`-ed out?

That is my understanding, yes.

PR #67, opened 2022-10-26
Add WebRTC transport interop ping test

John-LittleBearLabs:
Adding webrtc to the SupportedTransports in the composer version of the ping tests, which are in turn, I believe, also a work-in-progress.

Currently it depends heavily on https://github.com/ckousik/go-libp2p/tree/ckousik/webrtc since that has webrtc implemented for golang. Some goofy hacks are involved in getting that to work properly. We'll want to have that all sorted and webrtc merged in (and probably released) before this hits test-plans master.


mxinden:
@John-LittleBearLabs I am not quite sure what the goal is here, which is fine for now given that it is a draft. Let me know once this is ready for a review.

John-LittleBearLabs:
@mxinden 

> @John-LittleBearLabs I am not quite sure what the goal is here, which is fine for now given that it is a draft. Let me know once this is ready for a review.

Hopefully the new description helps to clear this up?

@GlenDC - I think you're pretty familiar with what I'm doing here. Does this all make sense to you? Assuming the answer is no, as this is rather opaque, perhaps the better question would be where should I elaborate more and/or things I should obviously fix right away.

mxinden (commented):
This pull request contains many changes that are unrelated, i.e. that are already in `master`. @john-littlebearlabs can you please clean up the commit history?

GlenDC:
> @mxinden
> 
> > @John-LittleBearLabs I am not quite sure what the goal is here, which is fine for now given that it is a draft. Let me know once this is ready for a review.
> 
> Hopefully the new description helps to clear this up?
> 
> @GlenDC - I think you're pretty familiar with what I'm doing here. Does this all make sense to you? Assuming the answer is no, as this is rather opaque, perhaps the better question would be where should I elaborate more and/or things I should obviously fix right away.

You’re doing more advanced testground work than me. I’m not yet at this level. There’s a lot i can learn from you. But from what i can see. This looks pretty good to me :) that being said, let’s perhaps meet tomorrow over video so I can pick some stuff up from you.

PR #68, opened 2022-10-27
Run `cargo update` before build

thomaseizinger:
I am hoping that this resolves the issue in https://github.com/libp2p/rust-libp2p/pull/2972.

What I think is happening here is that we can't resolve a dependency because some patch-versions of dependencies are outdated.

### DIFF
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -37,6 +37,7 @@ RUN mv /tmp/Cargo.toml /tmp/Cargo.lock ./plan
 ARG CARGO_FEATURES=""
 
 RUN cd ./plan/ && \
+        cargo update && \
         cargo build --release --features="${CARGO_FEATURES}"
 
 FROM debian:bullseye-slim


mxinden (commented):
Unfortunately this would make our CI less reproducable. Consecutive runs might have different dependencies.

I hope that https://github.com/libp2p/test-plans/pull/69 resolves the issue. If not, I am in favor of this change. At least for now.

laurentsenta:
I shared a few notes in: https://github.com/libp2p/test-plans/pull/69#issuecomment-1305758560

It would make sense to call `cargo update` but only when we build master and the current branch (via a docker arg).

We do something similar in go-libp2p:
https://github.com/libp2p/test-plans/blob/b97b36aa2754c77795635facbf96c8504bdd9135/ping/_compositions/go-cross-versions.toml#L49-L51

The only way to make this reproducible I think is to move the test to the `rust-libp2p` repo so that the repo (and PR author) is also responsible for bumping the Cargo files.

thomaseizinger:
Closing in favor of #72.

PR #69, opened 2022-11-02
ping/rust: Run cargo update

mxinden:
Updates dependency. In the hope to resolve conflict from https://github.com/libp2p/rust-libp2p/pull/2972/.

### DIFF
--- a/ping/rust/Cargo.lock
+++ b/ping/rust/Cargo.lock
@@ -107,9 +107,9 @@ dependencies = [
 
 [[package]]
 name = "async-global-executor"
-version = "2.3.0"
+version = "2.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


mxinden:
Proceeding here given that @laurentsenta is out. Happy to do any follow ups.

laurentsenta:
@mxinden thanks for taking care of this fix.
Would it make sense to call `cargo update` when we build master and custom branches? Since the dependencies might change there.

We do something similar in go-libp2p:
https://github.com/libp2p/test-plans/blob/b97b36aa2754c77795635facbf96c8504bdd9135/ping/_compositions/go-cross-versions.toml#L49-L51

laurentsenta:
We'd do this in https://github.com/libp2p/test-plans/blob/b97b36aa2754c77795635facbf96c8504bdd9135/ping/rust/Dockerfile#L39-L40

mxinden:
For the record, discussion moved here https://github.com/libp2p/test-plans/pull/68#pullrequestreview-1165154883

Issue #53, opened 2022-10-10
EPIC: support multi-dimensional testing


laurentsenta:
> As a libp2p maintainer, I want the ability to define test suites that combine many implementations, many muxers, many transports, etc. Defining and running these test suites should be simple, and the outcome should be clear. It should be easy to trigger these test suites before a release. It should be easy to display these results in a readable form on a website.

eta: 2022Q4

## Tasks

- [x] https://github.com/libp2p/test-plans/issues/55
  - this will add the tooling used to describe, build, and run multiple combinations of tests in Testground
- [ ] Implement new parameters (muxers, transports, etc)
  - we need to define what the matrix looks like
  - we need to update the ping test to support these parameters
     - :warning: we need to make sure we test the connection from A to B AND from B to A (we don't at the moment)
- [ ] add the "expected RTT" test
  - This requires the ability to "generate" the expected outcomes for every cell in this N-D matrix,
- [ ] https://github.com/libp2p/test-plans/issues/61
- [ ] Generate a "nice" dashboard https://github.com/libp2p/test-plans/issues/62
  - @BigLep shared some notes in https://github.com/libp2p/test-plans/issues/55#issuecomment-1279304776
  - [ ] use case 1: As a maintainer, I want to check the status of our interop testing before making a release.
  - [ ] use case 2: As a user, I want to know more about libp2p interoperability capabilities.

### Follow-up tasks

- [ ] "optimize" by skipping versions libraries that are not used anymore (see notes below),
- [ ] "optimize" using artifact caching,

### A high-level approach:

1. first, we use versions resources file to generate "complex" test matrixes,
2. then we use another resource (data or code) to produce the expected RTT matrix,
3. then we generate the relevant composition file (as shown below), with the expected RTT as a parameter,
4. then we iterate through these test cases and call `testground run`.

### Configurations

Ideally, the libp2p team provides a resource file that contains the versions and their features:

```toml
[[groups]]
# go v0.42
GoVersion = '1.18'
Modfile = "go.v0.22.mod"
Selector = 'v0.42'
Implementation = 'go'
SupportedTransports = ["tcp", "quic", "webrtc"]
Muxer = ["yamux"]

[[groups]]
# go v0.22
GoVersion = '1.18'
Modfile = "go.v0.22.mod"
Selector = 'v0.22'
Implementation = 'go'
SupportedTransports = ["tcp", "quic"]
Muxer = ["mplex"]


[[groups]]
# rust v0.51
[...]

PR #70, opened 2022-11-02
Nim ping test

Menduist:
Should work fine now! No metrics though

Old message:

---------
Works fine against itself, could use a hand to setup the interop tests themselves

I'm also not sure what format the pubsub payload should have, because it's being escaped
If I send {"publish": {"topic": "whatev", "payload": {"test": "whatev"}}
I receive something like "{\"test\": \"whatev\"}"

So I sent it as a string, which may fail with other implementations of the SDK

Also missing metrics since I don't support them in the SDK (yet)


laurentsenta:
Awesome, thanks for sharing,

If you want to give this a shot:
you could create a `all-interop-latest.toml`, copy the content of https://github.com/libp2p/test-plans/blob/master/ping/_compositions/go-rust-interop-latest.toml and add a `[groups]` section with the nim builder. The section should look something like: https://github.com/libp2p/test-plans/blob/b97b36aa2754c77795635facbf96c8504bdd9135/ping/_compositions/go-rust-interop-latest.toml#L18-L29 (ignore templating for now).

On the payload:
that's how the sync service works at the moment (on my todo is to describe the current, implicit, spec).
You get the JSON back, encoded as a string.

Example in rust:
https://github.com/testground/sdk-rust/blob/2b85e4c9578037bd8704de40b7f179bf7fcc8649/src/responses.rs#L110

Menduist:
It works!

Had to fix a few things (the string thing, and the nim SDK doesn't signal `initialized_global` which is apparently required for the other SDKs, so I did it manually here), but seems to work fine now :)

laurentsenta:
Nice!

If all you need is a git commit, recommendation:
create a `nim.toml` which specifies multiple groups with an `Id` and a `GitRef` value.

Check the `rust` part, I guess it would look something like:

```toml
{{ with (load_resource "./nim.toml") }}
  {{ with (index .groups 0) }} // get the latest
    [[groups]]
    id = "nim-latest-{{ .Id }}"
    instances = { count = 1 }
    builder = "docker:generic"

    [groups.build_config]
      path = "./rust/"

    [groups.build_config.build_args]
      GIT_REF = '{{ .GitRef }}'
  {{ end }}
{{ end }}
```

Menduist:
Done, + an attempt at adding to CI

Menduist:
Workflow ran successfully on the fork: https://github.com/status-im/libp2p-test-plans/actions/runs/3427378367
So I would say this is ready for review

.github/workflows/ping-interop-all.yml
```diff
@@ -0,0 +1,45 @@
+on:
```
mxinden, line 1: Instead of introducing a new workflow file, how about including nim in `ping-interop.yml`. I don't see a reason why we should treat Go or Rust special.

Maybe @MarcoPolo has opinions here.

mxinden (commented):
:rocket: happy to see this happening.

.github/workflows/ping-interop-all.yml
Menduist: `ping-interop.yml` will call the `go-rust-interop.toml` composition
So it could just call the `all-interop.toml` instead, but then the `go-rust-interop.toml` will never be called, and maybe it should be removed?
We would just have:
- `[language].toml`: list of versions of that implem
- `[language]-cross-versions.toml`: test every version of that implem
- `all-interop-latest.toml`: test last version of every implem
- `all-interop.toml`: test every version of every implem

Menduist: Did this in my latest commit, we can revert if not happy

mxinden: > We would just have:
> 
>     * `[language].toml`: list of versions of that implem
> 
>     * `[language]-cross-versions.toml`: test every version of that implem
> 
>     * `all-interop-latest.toml`: test last version of every implem
> 
>     * `all-interop.toml`: test every version of every implem

On a high level I am in favor of this. I will take a detailed look now.

.github/workflows/ping-interop-latest.yml
```diff
@@ -14,22 +14,22 @@ on:
...
       custom_interop_target:
-        description: in the case of cross-implementation testing, the implementation target (go | rust | ...)
+        description: in the case of cross-implementation testing, the implementation target (go | rust | nim)
         required: false
         type: string
   push:
   pull_request:
 
-name: libp2p ping - go + rust test (latest) with testground.
+name: libp2p ping - go + rust + nim test (latest) with testground.
```
mxinden, line 23: Nit pick, while we are at it, how about being consistent with the naming on `ping-interop.yml`?
```suggestion
name: libp2p ping - go, rust and nim test (latest) with testground.
```

mxinden (commented):
Very excited for us to have interoperability tests beyond Go, JS and Rust.

The proposed changes here look good to me.

I suggest waiting for another review from @MarcoPolo (as self-requested).

In addition, I wonder how we should coordinate this with https://github.com/libp2p/test-plans/issues/53, i.e. how we should prioritize this and the refactoring of our test matrix generation. As far as I can tell, it would be fine to proceed here before https://github.com/libp2p/test-plans/issues/53. What do you think @jxs and @MarcoPolo?

mxinden:
>  are the other implementation caching their dependencies properly to be reused by multiple images?

Would you mind merging `master` here @Menduist? https://github.com/libp2p/test-plans/pull/78 might resolve the issues you are seeing.

mxinden:
> I suggest waiting for another review from @MarcoPolo (as self-requested).

Friendly ping @MarcoPolo. Any objections moving forward here?

mxinden:
Triggered the tests. All green @Menduist :tada:

MarcoPolo (approved):

MarcoPolo:
Looks great! @Menduist I'll ping you for some help in getting the multidimensional testing in the near future. Maybe we can pair on it? Might be fun to pair on some Nim 👑 👑 👑

Menduist:
Sure! Thanks for the merge 🎉

PR #78, opened 2022-11-22
feat(ping/rust): Cache built dependencies in separate docker layer

thomaseizinger:
cc @mxinden @laurentsenta

### DIFF
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -4,7 +4,12 @@ WORKDIR /usr/src/testplan
 RUN apt-get update && apt-get install -y cmake protobuf-compiler
 
 ARG PLAN_PATH="./"
-COPY ./plan/${PLAN_PATH} ./plan
+
+RUN mkdir -p ./plan/src
+COPY ./plan/${PLAN_PATH}/src/main.rs ./plan/src/main.rs
+COPY ./plan/${PLAN_PATH}/Cargo.toml ./plan/Cargo.toml
+COPY ./plan/${PLAN_PATH}/Cargo.lock ./plan/Cargo.lock
+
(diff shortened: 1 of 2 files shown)


ping/rust/Dockerfile
```diff
@@ -13,6 +18,9 @@ RUN if [ ! -z "${GIT_TARGET}" ]; then sed -i "s,^git.*,git = \"https://${GIT_TAR
 ARG GIT_REF=""
 RUN if [ ! -z "${GIT_REF}" ]; then sed -i "s/^rev.*/rev = \"${GIT_REF}\"/" ./plan/Cargo.toml; fi
 
+COPY ./plan/${PLAN_PATH}/src/lib.rs ./plan/src/lib.rs
+COPY ./plan/${PLAN_PATH}/src/bin/ ./plan/src/bin/
```
laurentsenta, line 22: You'll need to `touch` the main file as well
https://github.com/ackintosh/discv5-testground/blob/9e72e49b333f44de6d51a8658ecf9108c64d8e19/Dockerfile#L15-L18

thomaseizinger: I don't think that is the case here because all the main's we are interested live under `src/bin`, i.e. we never replace the `src/main.rs` one. It is a permanent dummy.

thomaseizinger: I tried it locally and it successfully rebuilt things when I changed `lib.rs` or one of the main's in `src/bin`.

laurentsenta: Ah my bad, you're right, looks good to me then!

laurentsenta (approved):
lgtm 👍

mxinden (approved):
Thanks @thomaseizinger!

PR #79, opened 2022-11-23
ping/rust: Don't set PR ref for master build

mxinden:
When building the `master` image we would previously mistakingly set the ref to the PR commit hash.

This surfaced when testing a pull request from a fork. `master` would be pointed at the forks HEAD commit hash, but not the fork URL. Thus the build could not find the commit and would fail. See https://github.com/libp2p/rust-libp2p/pull/3154

I would consider this a hotfix. Long term it would be great to be able to support pull requests that target other branches than `master`.

What do folks think? Does the above make sense?

### DIFF
--- a/ping/_compositions/rust-cross-versions.toml
+++ b/ping/_compositions/rust-cross-versions.toml
@@ -25,7 +25,7 @@
 
     [groups.build_config.build_args]
       BINARY_NAME = '{{ .BinaryName }}'
-      GIT_REF = '{{ $.Env.GitReference }}'
+      GIT_REF = "master"
   {{ end }}
 
   {{ if $.Env.GitReference }}


mxinden:
This only affects the Rust CI, not the Golang CI. To unblock https://github.com/libp2p/rust-libp2p/pull/3154 and thus unblock the rust-libp2p v0.50.0 release, I will merge here. @jxs @thomaseizinger please still review the change.

thomaseizinger (commented):
I would have expected the GitHub workflow to pass the correct revision instead of us setting `master` here.

`pull_request` should be undefined [here](https://github.com/libp2p/rust-libp2p/blob/b99a2137ac12a295808c08931422dda86b52284a/.github/workflows/interop-test.yml#L25) when we push to master, thus I would have expected `github,sha` to be used which should be the new commit pushed to master.

ping/rust/Dockerfile
```diff
@@ -11,7 +11,7 @@ ARG GIT_TARGET=""
 RUN if [ ! -z "${GIT_TARGET}" ]; then sed -i "s,^git.*,git = \"https://${GIT_TARGET}\"," ./plan/Cargo.toml; fi
 
 ARG GIT_REF=""
-RUN if [ ! -z "${GIT_REF}" ]; then sed -i "s/^rev.*/rev = \"${GIT_REF}\"/" ./plan/Cargo.toml; fi
+RUN if [ ! -z "${GIT_REF}" ]; then sed -i "s/^rev.*/branch = \"${GIT_REF}\"/" ./plan/Cargo.toml; fi
```
mxinden, line 14: This change breaks the `custom` build, as that one uses a rev and not a branch :man_facepalming: silly me.

thomaseizinger: `rev` should accept a branch name actually: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

thomaseizinger: Have you tried `origin/master`?

thomaseizinger: You might need to prefix with `refs/`:

- `refs/remotes/origin/master`
- `refs/heads/master`

thomaseizinger: > `rev` should accept a branch name actually: [doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories)

This is talking about named references being supported so I think one of those two should work.

mxinden: Wonderful. Thanks @thomaseizinger. `refs/heads/master` resolves to the current master commit (https://github.com/libp2p/rust-libp2p/commit/babacc5555e8c6d2fdd25ad41a8654548870d2a9). For some reason `refs/removes/origin/master` points to https://github.com/libp2p/rust-libp2p/commit/b99a2137ac12a295808c08931422dda86b52284a, a previous commit on master.

I created https://github.com/libp2p/test-plans/pull/81/ adopting the former. Please take a look.

jxs (commented):
Thanks Max, yeah this makes sense as a hotfix, Agree with Max and Thomas that:

- We should allow passing other revisions than master to test against
- github should allow passing the base reference and target

PR #80, opened 2022-11-23
fix(ping/rust): Only specify branch when building master

mxinden:
Follow up to https://github.com/libp2p/test-plans/pull/79/. See discussion https://github.com/libp2p/test-plans/pull/79#discussion_r1030337736.


mxinden:
This should only affect the Rust CI. With that in mind and in order to unblock the v0.50.0 release, I am merging here.

PR #83, opened 2022-11-29
fix(ping/rust): Implemented 0.50 changes

umgefahren:
As requested by @thomaseizinger, this changes the test plans to take the removed APIs into account. See libp2p/rust-libp2p#3170 for details.

### DIFF
--- a/ping/rust/Cargo.toml
+++ b/ping/rust/Cargo.toml
@@ -31,5 +31,5 @@ libp2pv0490 = { package = "libp2p", version = "0.49.0", features = ["websocket",
 [dependencies."libp2pv0500"]
 package = "libp2p"
 git = "https://github.com/libp2p/rust-libp2p"
-rev = "368705a1465c4322948d14ee46d42475c472ca1e" # This should usually point to a commit on master.
+tag = "v0.50.0"
 features = ["websocket", "mplex", "yamux", "tcp", "ping", "noise", "dns", "async-std", "rsa", "macros"]
--- a/ping/rust/src/bin/testplan_0500.rs
+++ b/ping/rust/src/bin/testplan_0500.rs
@@ -12,7 +12,7 @@ async fn main() -> Result<()> {
     let local_key = identity::Keypair::generate_ed25519();
     let local_peer_id = PeerId::from(local_key.public());
 
-    let swarm = OrphanRuleWorkaround(Swarm::with_tokio_executor(
+    let swarm = OrphanRuleWorkaround(Swarm::with_async_std_executor(
         development_transport(local_key).await?,
         Behaviour {
             keep_alive: keep_alive::Behaviour,
--- a/ping/rust/src/bin/testplan_0500.rs
+++ b/ping/rust/src/bin/testplan_0500.rs
@@ -12,7 +12,7 @@ async fn main() -> Result<()> {
     let local_key = identity::Keypair::generate_ed25519();
     let local_peer_id = PeerId::from(local_key.public());
 
-    let swarm = OrphanRuleWorkaround(Swarm::new(
+    let swarm = OrphanRuleWorkaround(Swarm::with_async_std_executor(
         development_transport(local_key).await?,
         Behaviour {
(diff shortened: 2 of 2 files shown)


thomaseizinger (approved):
Thanks!

thomaseizinger:
@mxinden I can't merge this myself, please have a look!

ping/rust/src/bin/testplan_0500.rs
```diff
@@ -12,7 +12,7 @@ async fn main() -> Result<()> {
     let local_key = identity::Keypair::generate_ed25519();
     let local_peer_id = PeerId::from(local_key.public());
 
-    let swarm = OrphanRuleWorkaround(Swarm::new(
+    let swarm = OrphanRuleWorkaround(Swarm::with_async_std_executor(
```
mxinden, line 15:
```
   Compiling testplan v0.1.0 (/usr/src/testplan/plan)
error[E0599]: no function or associated item named `with_async_std_executor` found for struct `libp2p::Swarm` in the current scope
  --> src/bin/testplan_0500.rs:15:45
   |
15 |     let swarm = OrphanRuleWorkaround(Swarm::with_async_std_executor(
   |                                             ^^^^^^^^^^^^^^^^^^^^^^^ function or associated item not found in `libp2p::Swarm<_>`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `testplan` due to previous error
```
```suggestion
    let swarm = OrphanRuleWorkaround(Swarm::with_tokio_executor(
```

We are using tokio as an executor here. `libp2p` is configured with the `tokio` feature and not the `async-io` feature.

mxinden (commented):
Thanks for the patch @umgefahren!

mxinden:
> [19a8de2](https://github.com/libp2p/test-plans/commit/19a8de2567ff5f14dcbbeac1758a1ca808172be1) is not correct. Since it's called in an async_std block Tokio is the wrong executor and might lead to a runtime panic if I'm correct. In other words: If Tokio is correct, we should also use Tokio::main

You are absolutely right @umgefahren. My bad. Sorry for pushing a wrong commit to your branch.

Regarding the previous CI failure, my best next guess is that the commit hash referenced in the `Cargo.toml` is too old.

https://github.com/libp2p/test-plans/blob/31a128833da19f1ded62a73296c9e1764d97ac13/ping/rust/Cargo.toml#L34

Would you mind reverting my commit and updating the commit hash to a recent commit on `master`, ideally the one of `v0.50.0` tag?

ping/rust/Cargo.toml
```diff
@@ -31,5 +31,5 @@ libp2pv0490 = { package = "libp2p", version = "0.49.0", features = ["websocket",
 [dependencies."libp2pv0500"]
 package = "libp2p"
 git = "https://github.com/libp2p/rust-libp2p"
-rev = "368705a1465c4322948d14ee46d42475c472ca1e" # This should usually point to a commit on master.
+tag = "v0.50.0"
```
mxinden, line 34: I don't think moving to `tag` would work here, given that we use `sed` in our Dockerfile matching on the string `rev`.

I might have been not very clear. Sorry. Let's use the commit hash that `v0.50.0` points at.

umgefahren: Oh I see

thomaseizinger: Now that we have released `v0.50.0` we should actually add that as a release and have this point to the most recent hash on master here!

umgefahren: Should this be part of this PR?

thomaseizinger: It doesn't have to be. I think we can merge this separately ;)

thomaseizinger: I opened https://github.com/libp2p/test-plans/pull/87 and also fixed the deprecation warnings there.

mxinden:
Closing here in favor of https://github.com/libp2p/test-plans/pull/87. Thanks for the help!

Issue #61, opened 2022-10-20
[tracking issue] Interop test-plans for all existing/developing libp2p transports


p-shahi:
Part of the libp2p/test-plans roadmap: https://github.com/libp2p/test-plans/blob/master/ROADMAP.md#2-interop-test-plans-for-all-existingdeveloping-libp2p-transports

## Done Criteria
Using tooling from https://github.com/libp2p/test-plans/issues/53

- [x] All features of `go-libp2p`, `rust-libp2p`, and `js-libp2p` that should be interoperable have test-plans (i.e. transports (TCP, QUIC, WebRTC, WebTransport), multiplexers (mplex, yamux), secure channels (TLS, Noise), etc.) across versions.
- [x] Features currently in development across implementations (like WebRTC in go-libp2p and rust-libp2p, or QUIC & TLS in rust-libp2p) are not merged without at least manually running these test suites.
- [x] Test suites are run in `libp2p/test-plans` CI and before releasing a version of go-libp2p,  rust-libp2p, and js-libp2p (GitHub workflow added so that these suites run against the `master` branch on a nightly basis (updating the status check.))

## Tasks

- [x] #97 https://github.com/libp2p/test-plans/issues/101
    - [x] go-libp2p (DRI: @MarcoPolo)
      - [X] TCP, QUIC, WebSocket: #97 
      - [x] WebTransport: (link to PR)
      - [ ] WebRTC: blocked on https://github.com/libp2p/go-libp2p/pull/2337
    - [x] rust-libp2p (DRI: @MarcoPolo)
      - [X] TCP, QUIC, WebSocket: #97
      - [x] WebRTC: https://github.com/libp2p/test-plans/pull/100
      - [x] WebRTC browser: https://github.com/libp2p/test-plans/pull/130
    - [x] node.js-libp2p (DRI: @MarcoPolo )
      - [X] TCP, WebSocket: #97 
      - [x] ~WebTransport: (link to PR)~ (this is not supported in node right now)
      - [x] ~WebRTC: (link to PR)~ (not supported in node)
      - [x] In repo testing https://github.com/libp2p/js-libp2p/pull/1615
    - [x] browser js-libp2p https://github.com/libp2p/test-plans/pull/104
      - [x] WebSocket: https://github.com/libp2p/test-plans/pull/144
      - [x] WebTransport: (link to PR)
      - [x] WebRTC: https://github.com/libp2p/test-plans/pull/130
[...]

PR #84, opened 2022-11-29
Jt/82 webrtcinrustping

John-LittleBearLabs:
Finally appears to be working, with proviso below.

`Connection established! PeerId("12D3KooWNfGZwVxzv4Eu2RvmCrLKQibdp2U57E93WH36rwygYk8Y")=Dialer { address: "/ip4/16.13.0.2/udp/1234/webrtc/certhash/uEiBLxsx_9DM981WqZCG8H0LDEcHRzJxJM8zgkpqXhVH95Q", role_override: Dialer }`

`Outcome:	success`

Before this is merged in we'd want to undo a couple things that I have to short-circuit to quickly test this:
* The default value for the transport testparam must be tcp, not webrtc
* Need to re-enable the older versions in rust.toml

### DIFF
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -26,7 +26,7 @@ COPY ./plan/${PLAN_PATH}/src/bin/ ./plan/src/bin/
 ARG BINARY_NAME
 RUN echo BINARY_NAME=${BINARY_NAME}
 RUN cd ./plan/ && cargo build --bin=${BINARY_NAME}
-RUN cd ./plan/ && mv /usr/src/testplan/plan/target/debug/${BINARY_NAME} /usr/local/bin/testplan
+RUN cd ./plan/ && mv -v /usr/src/testplan/plan/target/debug/${BINARY_NAME} /usr/local/bin/testplan
 
 FROM debian:bullseye-slim
 COPY --from=builder /usr/local/bin/testplan /usr/local/bin/testplan
@@ -35,5 +35,5 @@ EXPOSE 6060
 #TODO revert
 ENV UNIQUELY_NAMED_ENV=webrtc
 
-ENV RUST_BACKTRACE=1
+ENV RUST_BACKTRACE=full
 ENTRYPOINT ["testplan"]
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -10,6 +10,7 @@ COPY ./plan/${PLAN_PATH}/src/main.rs ./plan/src/main.rs
 COPY ./plan/${PLAN_PATH}/Cargo.toml ./plan/Cargo.toml
 COPY ./plan/${PLAN_PATH}/Cargo.lock ./plan/Cargo.lock
 
+RUN echo initial build
 RUN cd ./plan/ && cargo build  # Initial build acts as a cache.
 
 ARG GIT_TARGET=""
@@ -23,12 +24,14 @@ COPY ./plan/${PLAN_PATH}/src/bin/ ./plan/src/bin/
 
 # Build the requested binary: Cargo will update lockfile on changed manifest (i.e. if one of the above `sed`s patched it).
 ARG BINARY_NAME
-RUN cd ./plan/ \
-    && cargo build --bin=${BINARY_NAME} \
-    && mv /usr/src/testplan/plan/target/debug/${BINARY_NAME} /usr/local/bin/testplan
+RUN echo BINARY_NAME=${BINARY_NAME}
+RUN cd ./plan/ && cargo build --bin=${BINARY_NAME}
+RUN cd ./plan/ && cargo build --bin=${BINARY_NAME}
+RUN cd ./plan/ && mv /usr/src/testplan/plan/target/debug/${BINARY_NAME} /usr/local/bin/testplan
 
 FROM debian:bullseye-slim
 COPY --from=builder /usr/local/bin/testplan /usr/local/bin/testplan
 EXPOSE 6060
+ENV TRANSPORT=webrtc # TODO revert
 ENV RUST_BACKTRACE=1
 ENTRYPOINT ["testplan"]
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -32,8 +32,5 @@ FROM debian:bullseye-slim
 COPY --from=builder /usr/local/bin/testplan /usr/local/bin/testplan
 EXPOSE 6060
 
-#TODO revert
-ENV UNIQUELY_NAMED_ENV=webrtc
-
 ENV RUST_BACKTRACE=full
 ENTRYPOINT ["testplan"]
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -26,12 +26,14 @@ COPY ./plan/${PLAN_PATH}/src/bin/ ./plan/src/bin/
 ARG BINARY_NAME
 RUN echo BINARY_NAME=${BINARY_NAME}
 RUN cd ./plan/ && cargo build --bin=${BINARY_NAME}
-RUN cd ./plan/ && cargo build --bin=${BINARY_NAME}
 RUN cd ./plan/ && mv /usr/src/testplan/plan/target/debug/${BINARY_NAME} /usr/local/bin/testplan
 
 FROM debian:bullseye-slim
 COPY --from=builder /usr/local/bin/testplan /usr/local/bin/testplan
 EXPOSE 6060
-ENV TRANSPORT=webrtc # TODO revert
+
+#TODO revert
+ENV UNIQUELY_NAMED_ENV=webrtc
+
 ENV RUST_BACKTRACE=1
 ENTRYPOINT ["testplan"]
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.62-bullseye as builder
+FROM rust:1.65-bullseye as builder
 WORKDIR /usr/src/testplan
 
 RUN apt-get update && apt-get install -y cmake protobuf-compiler
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -17,7 +17,7 @@ ARG GIT_TARGET=""
 RUN if [ ! -z "${GIT_TARGET}" ]; then sed -i "s,^git.*,git = \"https://${GIT_TARGET}\"," ./plan/Cargo.toml; fi
 
 ARG GIT_REF=""
-RUN if [ "master" = "${GIT_REF}" ]; then sed -i "s/^rev.*/branch= \"master\"/" ./plan/Cargo.toml; elif [ ! -z "${GIT_REF}" ]; then sed -i "s/^rev.*/rev = \"${GIT_REF}\"/" ./plan/Cargo.toml; fi
+RUN if [ ! -z "${GIT_REF}" ]; then sed -i "s,^rev.*,rev = \"${GIT_REF}\"," ./plan/Cargo.toml; fi
 
 COPY ./plan/${PLAN_PATH}/src/lib.rs ./plan/src/lib.rs
 COPY ./plan/${PLAN_PATH}/src/bin/ ./plan/src/bin/
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -10,7 +10,6 @@ COPY ./plan/${PLAN_PATH}/src/main.rs ./plan/src/main.rs
 COPY ./plan/${PLAN_PATH}/Cargo.toml ./plan/Cargo.toml
 COPY ./plan/${PLAN_PATH}/Cargo.lock ./plan/Cargo.lock
 
-RUN echo initial build
 RUN cd ./plan/ && cargo build  # Initial build acts as a cache.
 
 ARG GIT_TARGET=""
@@ -24,13 +23,12 @@ COPY ./plan/${PLAN_PATH}/src/bin/ ./plan/src/bin/
 
 # Build the requested binary: Cargo will update lockfile on changed manifest (i.e. if one of the above `sed`s patched it).
 ARG BINARY_NAME
-RUN echo BINARY_NAME=${BINARY_NAME}
-RUN cd ./plan/ && cargo build --bin=${BINARY_NAME}
-RUN cd ./plan/ && mv -v /usr/src/testplan/plan/target/debug/${BINARY_NAME} /usr/local/bin/testplan
+RUN cd ./plan/ \
+    && cargo build --bin=${BINARY_NAME} \
+    && mv /usr/src/testplan/plan/target/debug/${BINARY_NAME} /usr/local/bin/testplan
 
 FROM debian:bullseye-slim
 COPY --from=builder /usr/local/bin/testplan /usr/local/bin/testplan
 EXPOSE 6060
-EXPOSE 1234
-ENV RUST_BACKTRACE=full
+ENV RUST_BACKTRACE=1
 ENTRYPOINT ["testplan"]
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -31,6 +31,6 @@ RUN cd ./plan/ && mv -v /usr/src/testplan/plan/target/debug/${BINARY_NAME} /usr/
 FROM debian:bullseye-slim
 COPY --from=builder /usr/local/bin/testplan /usr/local/bin/testplan
 EXPOSE 6060
-
+EXPOSE 1234
 ENV RUST_BACKTRACE=full
 ENTRYPOINT ["testplan"]
--- a/ping/rust/Dockerfile
+++ b/ping/rust/Dockerfile
@@ -4,14 +4,22 @@ WORKDIR /usr/src/testplan
 RUN apt-get update && apt-get install -y cmake protobuf-compiler
 
 ARG PLAN_PATH="./"
-COPY ./plan/${PLAN_PATH} ./plan
+
+RUN mkdir -p ./plan/src
+COPY ./plan/${PLAN_PATH}/src/main.rs ./plan/src/main.rs
+COPY ./plan/${PLAN_PATH}/Cargo.toml ./plan/Cargo.toml
+COPY ./plan/${PLAN_PATH}/Cargo.lock ./plan/Cargo.lock
+
 RUN cd ./plan/ && cargo build  # Initial build acts as a cache.
 
 ARG GIT_TARGET=""
 RUN if [ ! -z "${GIT_TARGET}" ]; then sed -i "s,^git.*,git = \"https://${GIT_TARGET}\"," ./plan/Cargo.toml; fi
 
 ARG GIT_REF=""
-RUN if [ ! -z "${GIT_REF}" ]; then sed -i "s/^rev.*/rev = \"${GIT_REF}\"/" ./plan/Cargo.toml; fi
+RUN if [ "master" = "${GIT_REF}" ]; then sed -i "s/^rev.*/branch= \"master\"/" ./plan/Cargo.toml; elif [ ! -z "${GIT_REF}" ]; then sed -i "s/^rev.*/rev = \"${GIT_REF}\"/" ./plan/Cargo.toml; fi
+
+COPY ./plan/${PLAN_PATH}/src/lib.rs ./plan/src/lib.rs
+COPY ./plan/${PLAN_PATH}/src/bin/ ./plan/src/bin/
 
 # Build the requested binary: Cargo will update lockfile on changed manifest (i.e. if one of the above `sed`s patched it).
 ARG BINARY_NAME
--- a/ping/rust/src/bin/testplan_0440.rs
+++ b/ping/rust/src/bin/testplan_0440.rs
@@ -31,7 +31,7 @@ struct OrphanRuleWorkaround(Swarm<ping::Behaviour>);
 
 #[async_trait]
 impl PingSwarm for OrphanRuleWorkaround {
-    async fn listen_on(&mut self, address: &str) -> Result<()> {
+    async fn listen_on(&mut self, address: &str) -> Result<Option<String>> {
         let id = self.0.listen_on(address.parse()?)?;
 
         loop {
@@ -42,7 +42,7 @@ impl PingSwarm for OrphanRuleWorkaround {
             }
         }
 
-        Ok(())
+        Ok(None)
     }
 
     fn dial(&mut self, address: &str) -> Result<()> {
--- a/ping/rust/src/bin/testplan_0440.rs
+++ b/ping/rust/src/bin/testplan_0440.rs
@@ -21,8 +21,8 @@ async fn main() -> Result<()> {
         ),
         local_peer_id,
     ));
-    let client = testground::client::Client::new_and_init().await.unwrap();
-    run_ping(swarm,client).await?;
+    let client = testground::client::Client::new_and_init().await?;
+    run_ping(swarm, client).await?;
 
     Ok(())
 }
--- a/ping/rust/src/bin/testplan_0440.rs
+++ b/ping/rust/src/bin/testplan_0440.rs
@@ -21,8 +21,8 @@ async fn main() -> Result<()> {
         ),
         local_peer_id,
     ));
-
-    run_ping(swarm).await?;
+    let client = testground::client::Client::new_and_init().await.unwrap();
+    run_ping(swarm,client).await?;
 
     Ok(())
 }
--- a/ping/rust/src/bin/testplan_0440.rs
+++ b/ping/rust/src/bin/testplan_0440.rs
@@ -66,9 +66,9 @@ impl PingSwarm for OrphanRuleWorkaround {
 
         while received_pings.len() < number {
             if let Some(SwarmEvent::Behaviour(ping::Event {
-                                                  peer,
-                                                  result: Ok(ping::Success::Ping { .. }),
-                                              })) = self.0.next().await
+                peer,
+                result: Ok(ping::Success::Ping { .. }),
+            })) = self.0.next().await
             {
                 received_pings.insert(peer);
             }
--- a/ping/rust/src/bin/testplan_0440.rs
+++ b/ping/rust/src/bin/testplan_0440.rs
@@ -21,7 +21,9 @@ async fn main() -> Result<()> {
         ),
         local_peer_id,
     ));
-    let client = testground::client::Client::new_and_init().await.expect("Unable to init testground cient.");
+    let client = testground::client::Client::new_and_init()
+        .await
+        .expect("Unable to init testground cient.");
     run_ping(swarm, client).await?;
 
     Ok(())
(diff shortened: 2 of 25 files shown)


ping/rust/src/lib.rs
```diff
@@ -10,7 +10,7 @@ use testground::network_conf::{
     FilterAction, LinkShape, NetworkConfiguration, RoutingPolicyType, DEFAULT_DATA_NETWORK,
 };
 
-const LISTENING_PORT: u16 = 1234;
+const LISTENING_PORT: &str = "1234";
```
mxinden, line 13: Nit pick: Let's revert this change before merging.

mxinden (commented):
Direction here looks good to me.

@John-LittleBearLabs does a Testground run with these changes succeed for you locally?

Would you mind resolving the merge conflicts and marking it ready for review? That way we can test your changes on our CI.

p-shahi:
Approved and ran the CI workflow, I think it'll be good to add you guys as libp2p contributors in github-mgmt

jxs (commented):
Hi John, this already looks very good! Left some minor nits. Btw @mxinden should we add a composition (manually for now) testing webrtc to webrtc transport?

ping/rust/Dockerfile
```diff
@@ -23,12 +24,13 @@ COPY ./plan/${PLAN_PATH}/src/bin/ ./plan/src/bin/
 
 # Build the requested binary: Cargo will update lockfile on changed manifest (i.e. if one of the above `sed`s patched it).
 ARG BINARY_NAME
-RUN cd ./plan/ \
-    && cargo build --bin=${BINARY_NAME} \
-    && mv /usr/src/testplan/plan/target/debug/${BINARY_NAME} /usr/local/bin/testplan
+RUN echo BINARY_NAME=${BINARY_NAME}
```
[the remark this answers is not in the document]
John-LittleBearLabs: :+1: reverting

ping/rust/src/lib.rs
```diff
@@ -27,34 +27,40 @@ pub trait PingSwarm: Sized {
...
-
     env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
 
-    let client = testground::client::Client::new_and_init().await.unwrap();
+    info!("Running ping test: {}", swarm.local_peer_id());
 
-    let local_addr = match if_addrs::get_if_addrs()
+    let transport = transport_param(&client);
+    let local_ip_addr = match if_addrs::get_if_addrs()
         .unwrap()
```
jxs, line 40: sorry John, oversaw these two, can we also update these two `unwrap's`?

jxs (approved):
LGTM, thanks John! @thomaseizinger maybe you also wanna take a look.

thomaseizinger:
> LGTM, thanks John! @thomaseizinger maybe you also wanna take a look.

Thanks for the ping, I'll take a look tomorrow!

ping/rust/src/bin/testplan_0500.rs
```diff
@@ -2,30 +2,47 @@ use anyhow::Result;
...
+use testplan::*;
 
 #[async_std::main]
 async fn main() -> Result<()> {
     let local_key = identity::Keypair::generate_ed25519();
     let local_peer_id = PeerId::from(local_key.public());
-
-    let swarm = OrphanRuleWorkaround(Swarm::with_async_std_executor(
-        development_transport(local_key).await?,
+    let client = testground::client::Client::new_and_init().await.unwrap();
```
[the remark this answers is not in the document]
jxs: ah, I see. It's because of the return type of `new_and_init` no worries John.

>Is there a particular reason why unwrapping in main is problematic? The program is going to exit with a non-zero return code either way. If I did an expect it'll have a message and stack trace.

it's mostly consistency, as we have the `-> Result<()>` kinda implies we should be returning a Result

ping/rust/src/bin/testplan_0500.rs
```diff
@@ -37,18 +51,23 @@ struct OrphanRuleWorkaround(Swarm<Behaviour>);
 
 #[async_trait]
 impl PingSwarm for OrphanRuleWorkaround {
-    async fn listen_on(&mut self, address: &str) -> Result<()> {
+    async fn listen_on(&mut self, address: &str) -> Result<Option<String>> {
```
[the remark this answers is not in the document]
John-LittleBearLabs: The Option is because none of the existing tests return an address and I didn't want to update them all (and the signature must be the same across versions for the trait).

In older tests the address being listened on was also dialable. The return value is really only helpful for WebRTC with the `/certhash/` business.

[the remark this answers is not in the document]
thomaseizinger: In theory, the returned address is also useful if you listened on port 0 for other transports as this address will contain the actual port chosen by the kernel.

If you look at the other tests, you end up defaulting to the passed in address so it will be cleaner if we just always return the address here because there is one less conditional operation.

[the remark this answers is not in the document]
John-LittleBearLabs: It's not conditional in the old tests (everything prior to testplan_0500). Their versions of this function always returns None in the Ok case, and test always ignores it.

I could make the newer tests always rely on this function and bail if they got None. So, one less conditional there.

Or, I can change the trait to return `Result<String>` and convert all the old implementations of listen_on to return an address. And then, if you like, we could change the old tests to use this, or continue ignoring it.

Since they already work, and the goal was to test new functionality, I was hesitant to change the old tests more than necessary. But whatever works.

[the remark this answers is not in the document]
thomaseizinger: > It's not conditional in the old tests

Matching on an `Option` sounds pretty conditional to me :man_shrugging: 

> Or, I can change the trait to return `Result<String>` and convert all the old implementations of listen_on to return an address. And then, if you like, we could change the old tests to use this

That would be my preference, yes.

> Since they already work, and the goal was to test new functionality, I was hesitant to change the old tests more than necessary.

That is a good goal overall. I created the `PingSwarm` abstraction with the back-then usecases in mind. It is however not set in stone and I'd prefer to iterate on it to keep as much functionality as possible shared between the tests.

It is not ideal that we have to maintain old tests. My personal long-term goal is to do something like https://github.com/testground/testground/issues/1522 to fix this where "old tests" are somewhere in our Git history and it is only the back-then created docker containers that are still around and used for the tests.

ping/rust/src/bin/testplan_0500.rs
```diff
@@ -1,27 +1,41 @@
...
+    let transport = match transport_param(&client).as_str() {
+        "tcp" => tokio_development_transport(local_key)?,
+        "webrtc" => webrtc::tokio::Transport::new(
+            local_key,
+            webrtc::tokio::Certificate::generate(&mut thread_rng())?,
+        )
+        .map(|(peer_id, conn), _| (peer_id, StreamMuxerBox::new(conn)))
+        .boxed(),
+        unhandled => unimplemented!("Transport unhandled in test: '{}'", unhandled),
+    };
```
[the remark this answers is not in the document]
John-LittleBearLabs: So if I take your suggestion perhaps a little too literally, I copy-paste the guts of tokio_development_transport here, and add an or_transport(webrtc), I get an error. Here's a minimal example:

````
//This works - TCP can do this
tcp.upgrade(core::upgrade::Version::V1).authenticate(noise_auth);
//This has a compiler error...
webrtc.upgrade(core::upgrade::Version::V1).authenticate(noise_auth);
// the trait `futures::AsyncRead` is not implemented for `(libp2p::PeerId, libp2p::libp2p_webrtc::tokio::Connection)`
````

Authenticated expects the Transport's Output type to impl AsyncRead, etc.

````
C: AsyncRead + AsyncWrite + Unpin,
````

Which is true for tcp's `type Output = T::Stream;` , but not webrtc's `type Output = (PeerId, Connection);`

Maybe...
1) I'm doing this wrong and I misread your intent
2) This isn't a supported use case for WebRTC (I believe it's doing noise stuff anyhow)
3) This is something we need to fix about transports/webrtc
4) Some combination of the above?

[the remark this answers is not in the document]
Comment (thomaseizinger) [reply ping/rust/src/bin/testplan_0500.rs]: This patch should sort things:

```patch
Index: ping/rust/src/bin/testplan_0510.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/ping/rust/src/bin/testplan_0510.rs b/ping/rust/src/bin/testplan_0510.rs
--- a/ping/rust/src/bin/testplan_0510.rs	(revision 24c1f23b92672756c12810d3b31752b8c4387a9b)
+++ b/ping/rust/src/bin/testplan_0510.rs	(date 1670822469058)
@@ -2,16 +2,19 @@
 use async_trait::async_trait;
 use futures::StreamExt;
 use libp2pv0510::{
+    core::either::EitherOutput,
     core::muxing::StreamMuxerBox,
-    identity, ping,
+    core::upgrade::SelectUpgrade,
+    core::upgrade::Version,
+    identity, mplex, noise, ping,
     swarm::{keep_alive, NetworkBehaviour, SwarmEvent},
-    tokio_development_transport, webrtc, Multiaddr, PeerId, Swarm, Transport,
+    tcp, webrtc, yamux, Multiaddr, PeerId, Swarm, Transport,
 };
 use log::{debug, info};
 use rand::thread_rng;
 use std::collections::HashSet;
 use std::time::Duration;
-use testplan::{run_ping, transport_param, PingSwarm};
+use testplan::{run_ping, PingSwarm};
 
 #[tokio::main]
 async fn main() -> Result<()> {
@@ -20,16 +23,25 @@
     let client = testground::client::Client::new_and_init()
         .await
         .expect("Unable to init testground cient.");
-    let transport = match transport_param(&client).as_str() {
-        "tcp" => tokio_development_transport(local_key)?,
-        "webrtc" => webrtc::tokio::Transport::new(
+
+    let transport = tcp::tokio::Transport::default()
+        .upgrade(Version::V1)
+        .authenticate(noise::NoiseAuthenticated::xx(&local_key).unwrap())
+        .multiplex(SelectUpgrade::new(
+            yamux::YamuxConfig::default(),
+            mplex::MplexConfig::default(),
+        ))
+        .timeout(Duration::from_secs(20))
+        .or_transport(webrtc::tokio::Transport::new(
             local_key,
             webrtc::tokio::Certificate::generate(&mut thread_rng())?,
-        )
-            .map(|(peer_id, conn), _| (peer_id, StreamMuxerBox::new(conn)))
-            .boxed(),
-        unhandled => unimplemented!("Transport unhandled in test: '{}'", unhandled),
-    };
+        ))
+        .map(|either, _| match either {
+            EitherOutput::First((p, conn)) => (p, StreamMuxerBox::new(conn)),
+            EitherOutput::Second((p, conn)) => (p, StreamMuxerBox::new(conn)),
+        })
+        .boxed();
+
     let swarm = OrphanRuleWorkaround(Swarm::with_tokio_executor(
         transport,
         Behaviour {

```

Once we get multi-dimensional testing (and thus more parameters on what we need to enable (yamux/mplex, noise/tls, etc) we likely need the flexibility anyway of configuring exactly what to use so I just opted for inlining `tokio_development_transport` and stripping the stuff we don't need.

ping/rust/src/bin/testplan_0440.rs
```diff
@@ -66,9 +66,9 @@ impl PingSwarm for OrphanRuleWorkaround {
 
         while received_pings.len() < number {
             if let Some(SwarmEvent::Behaviour(ping::Event {
-                peer,
-                result: Ok(ping::Success::Ping { .. }),
-            })) = self.0.next().await
+                                                  peer,
+                                                  result: Ok(ping::Success::Ping { .. }),
+                                              })) = self.0.next().await
```
thomaseizinger, line 71: Can we undo these formatting changes please?

ping/rust/src/lib.rs
```diff
@@ -70,19 +73,20 @@ where
         // two peers dialling each other at the same time).
         //
         // We can do this because sync service pubsub is ordered.
-        .take_while(|a| ready(a != &local_addr));
+        .take_while(|a| ready(a != &local_addr && a != &dialable_multiaddr));
```
thomaseizinger, line 76: I think we can now simplify this.

```suggestion
        .take_while(|a| ready(a != &dialable_multiaddr));
```

thomaseizinger (commented):
Two more comments.

ping/rust/src/bin/testplan_0500.rs
```diff
@@ -1,27 +1,68 @@
...
+#[tokio::main]
 async fn main() -> Result<()> {
     let local_key = identity::Keypair::generate_ed25519();
     let local_peer_id = PeerId::from(local_key.public());
-
-    let swarm = OrphanRuleWorkaround(Swarm::with_async_std_executor(
-        development_transport(local_key).await?,
+    let client = testground::client::Client::new_and_init()
+        .await
+        .expect("Unable to init testground cient.");
```
thomaseizinger, line 38: We no longer need the client here now so I'd like to move its creation back into `PingSwarm`.

ping/rust/src/bin/testplan_0500.rs
```diff
@@ -37,18 +78,23 @@ struct OrphanRuleWorkaround(Swarm<Behaviour>);
...
+                            listener_id,
+                            address,
+                        }) = self.0.next().await
+            {
                 if listener_id == id {
-                    break;
+                    debug!(
+                        "NewListenAddr event: listener_id={:?}, address={:?}",
+                        &listener_id, &address
+                    );
```
thomaseizinger, line 93: Unless you deem the listener ID very important, we could move this log to the ping swarm so all tests benefit from it.

thomaseizinger (commented):
Thanks! The final step is to remove the `client` parameter again as we no longer need it now outside the `PingSwarm`.

ping/rust/src/lib.rs
```diff
@@ -28,33 +29,39 @@ pub trait PingSwarm: Sized {
...
-        .unwrap()
+    let client = testground::client::Client::new_and_init()
+        .await
+        .expect("Unable to init testground cient.");
+    let transport = transport_param(&client);
+    let local_ip_addr = match if_addrs::get_if_addrs()?
         .into_iter()
         .find(|iface| iface.name == "eth1")
-        .unwrap()
+        .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "Can't find iface eth1"))?
```
thomaseizinger, line 46:
```suggestion
        .context("Can't find iface eth1"))?
```

We are using anyhow so this would work too!

ping/rust/src/lib.rs
```diff
@@ -28,33 +29,39 @@ pub trait PingSwarm: Sized {
...
     env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
 
-    let client = testground::client::Client::new_and_init().await.unwrap();
+    info!("Running ping test: {}", swarm.local_peer_id());
 
-    let local_addr = match if_addrs::get_if_addrs()
-        .unwrap()
+    let client = testground::client::Client::new_and_init()
+        .await
+        .expect("Unable to init testground cient.");
```
thomaseizinger, line 41:
```suggestion
        .context("Unable to init testground client.")?;
```

thomaseizinger (approved):
Thanks! All looks good to me. We are not enforcing it in here but can you make sure that everything is formatted with `cargo fmt`?

ddimaria:
> Thanks! All looks good to me. We are not enforcing it in here but can you make sure that everything is formatted with `cargo fmt`?

This is done.  I'll run clippy as well shortly.  Did we want to add any fmt/lint to CI @thomaseizinger?

thomaseizinger:
> > Thanks! All looks good to me. We are not enforcing it in here but can you make sure that everything is formatted with `cargo fmt`?
> 
> This is done.  I'll run clippy as well shortly.  Did we want to add any fmt/lint to CI @thomaseizinger?

That would be great! If you are up for it, please send a separate PR! :)

ping/rust/src/lib.rs
```diff
@@ -193,3 +201,11 @@ where
 
     Ok(())
 }
+
+pub fn transport_param(client: &testground::client::Client) -> String {
+    client
+        .run_parameters()
+        .test_instance_params
+        .get("transport").cloned()
```
thomaseizinger, line 209: Pretty sure rustfmt would want that on a separate line now 🙃

ddimaria: Ha, you're too fast for me @thomaseizinger 😉

ddimaria: That, and rust analyzer needed a restart.

ping/rust/src/bin/testplan_0440.rs
ddimaria: This was resolved with a whole pass of `cargo fmt`

ping/rust/src/bin/testplan_0500.rs
ddimaria: Confirmed this has been removed.

ping/rust/src/lib.rs
ddimaria: Thanks for the suggestion!  This has been updated.

ddimaria:
Thanks for the comments and suggestions @thomaseizinger!  I believe all of them have been resolved.  Please let me know if there anything else you'd like done here.  I left the unwraps alone that were in existing code...I'm not familiar with this codebase but figured there was a reason they were in there 🤷.  Did you want those done in a separate PR?

thomaseizinger:
> Thanks for the comments and suggestions @thomaseizinger!  I believe all of them have been resolved.  Please let me know if there anything else you'd like done here.  I left the unwraps alone that were in existing code...I'm not familiar with this codebase but figured there was a reason they were in there 🤷.  Did you want those done in a separate PR?

It doesn't matter much because either will abort the test program and report a failure. Just a matter of consistency really but nothing too important :)

ping/rust/src/lib.rs
```diff
@@ -31,30 +31,37 @@ pub async fn run_ping<S>(mut swarm: S) -> Result<()>
...
 
-    let client = testground::client::Client::new_and_init().await.unwrap();
+    info!("Running ping test: {}", swarm.local_peer_id());
 
-    let local_addr = match if_addrs::get_if_addrs()
-        .unwrap()
+    let client = testground::client::Client::new_and_init()
+        .await
+        .map_err(|e| anyhow::anyhow!(e.to_string()))
+        .context("Unable to init testground cient.")?;
```
thomaseizinger, line 41:
```suggestion
        .context("Unable to init testground cient.")?;
```

Double mapping is a bit useless :)

thomaseizinger, line 41:
```suggestion
        .context("Unable to init testground client")?;
```

Typo.

ping/rust/src/lib.rs
```diff
@@ -193,3 +201,12 @@ where
...
     Ok(())
 }
+
+pub fn transport_param(client: &testground::client::Client) -> String {
+    client
+        .run_parameters()
+        .test_instance_params
+        .get("transport")
+        .cloned()
+        .unwrap_or_else(|| "tcp".to_owned())
```
thomaseizinger, line 211:
```suggestion
        .as_deref()
        .unwrap_or("tcp")
```

Just curious whether this would work? (Together with `&str` as a return type).

ping/rust/src/lib.rs
```diff
@@ -94,7 +102,7 @@ where
     swarm
         .await_connections(client.run_parameters().test_instance_count as usize - 1)
         .await;
-
+    info!("Connections awaited.");
```
thomaseizinger, line 105:
```suggestion
    info!("Connected to {} peers", client.run_parameters().test_instance_count - 1);
```

thomaseizinger (approved):
A few more suggestions, not pressing. Otherwise looks good to me! :)

ping/rust/src/lib.rs
ddimaria: I'll look at this in the morning (diff TZ than you), but the suggestion caused compilation errors:
```
error[E0599]: the method `context` exists for enum `Result<Client, Box<dyn std::error::Error>>`, but its trait bounds were not satisfied
   --> src/lib.rs:41:10
    |
41  |           .context("Unable to init testground cient.")?;
    |            ^^^^^^^ method cannot be called on `Result<Client, Box<dyn std::error::Error>>` due to unsatisfied trait bounds
    |
   ::: .rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:504:1
    |
504 |   pub enum Result<T, E> {
    |   --------------------- doesn't satisfy `_: anyhow::Context<Client, Box<dyn std::error::Error>>`
    |
   ::: .rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/alloc/src/boxed.rs:194:1
    |
194 | / pub struct Box<
195 | |     T: ?Sized,
196 | |     #[unstable(feature = "allocator_api", issue = "32838")] A: Allocator = Global,
197 | | >(Unique<T>, A);
    | | -
    | | |
    | | doesn't satisfy `Box<dyn std::error::Error>: Sync`
    | |_doesn't satisfy `Box<dyn std::error::Error>: std::marker::Send`
    |   doesn't satisfy `_: anyhow::context::ext::StdError`
    |
    = note: the following trait bounds were not satisfied:
            `Box<dyn std::error::Error>: anyhow::context::ext::StdError`
            which is required by `Result<Client, Box<dyn std::error::Error>>: anyhow::Context<Client, Box<dyn std::error::Error>>`
            `Box<dyn std::error::Error>: std::marker::Send`
            which is required by `Result<Client, Box<dyn std::error::Error>>: anyhow::Context<Client, Box<dyn std::error::Error>>`
            `Box<dyn std::error::Error>: Sync`
            which is required by `Result<Client, Box<dyn std::error::Error>>: anyhow::Context<Client, Box<dyn std::error::Error>>`
```

ping/rust/src/lib.rs
ddimaria: `.get("transport")` returns an `Option<&String>` and `.as_deref()` has no effect on the return type.  From the docs:
```
Converts from Option<T> (or &Option<T>) to Option<&T::Target>.
```

ping/rust/src/lib.rs
ddimaria: Good call, much more meaningful log output!

ping/rust/src/lib.rs
ddimaria: If we're looking to avoid the extra allocation, we could `.remove()` (instead of `.get()`) to pull the value out of the hashmap if the value is no longer needed?

thomaseizinger: > `.get("transport")` returns an `Option<&String>` and `.as_deref()` has no effect on the return type. From the docs:
> 
> ```
> Converts from Option<T> (or &Option<T>) to Option<&T::Target>.
> ```

Hmm, the intention was to to convert `&String` to `&str` which _should_ work because `String` implements `Deref<Target = str>`. Seems like we are missing something. It is not important though, I was just curious :)

ping/rust/src/lib.rs
thomaseizinger: Ah damn, that trait object is not `Send + 'static`! I think it would work otherwise. That is a shame. Let's leave it then!

ddimaria:
CI is failing on `run-ping-all`: https://github.com/John-LittleBearLabs/tg-libp2p/actions/runs/3706321649/jobs/6282598377

```
LLVM ERROR: IO failure on output stream: No space left on device
```

ddimaria:
@thomaseizinger do you have opinions about the failing CI?  More specifically, do you think tackling that is in the scope of this PR?  If so, how has PL handled this in the past (disk cleanup, hosted runners, ..etc.)?  There are other options to investigate, but I wanted to check in with you first.

galargh:
> @thomaseizinger do you have opinions about the failing CI? More specifically, do you think tackling that is in the scope of this PR? If so, how has PL handled this in the past (disk cleanup, hosted runners, ..etc.)? There are other options to investigate, but I wanted to check in with you first.

Thanks to https://github.com/pl-strflt/tf-aws-gh-runner we can set up self-hosted runners in no time. Once we clear this repo to use PL self-hosted runners and we prepare a runner that has all the deps testground needs, all it should require is changing labels here.

I know @laurentsenta also wanted to set up a long running testground daemon on EC2. For this, the only difference would be setting a testground daemon url that the CLI should communicate with.

The latter is a tad bit more work. I think I'd be able to get the former ready today. Worth noting, the two are not mutually exclusive. Also, since the changes required are so minimal, we're not locking ourselves in in any given solution.

I'll update once self-hosted runners are ready for use here. I'll propose the changes in a separate PR.

thomaseizinger:
> @thomaseizinger do you have opinions about the failing CI? More specifically, do you think tackling that is in the scope of this PR? If so, how has PL handled this in the past (disk cleanup, hosted runners, ..etc.)? There are other options to investigate, but I wanted to check in with you first.

Thanks for the ping and sorry for the late reply.

GH-hosted runners give us 14GB of disk space. The fact that we are exceeding this is in my opinion another pointer that the current architecture of these tests simply does not scale.

It is completely unnecessary that we rebuild all of these tests on every test run. I am of the strong opinion that these docker containers are an artifact that is produced once (1x), uploaded to a registry and then only pulled for each test.

It is also unnecessary that we maintain them in parallel. Old versions of the test should never need to change. The test should be co-located with the library code over at https://github.com/libp2p/rust-libp2p.

That is the proper solution in my opinion. I opened an issue about this in testground a while ago: https://github.com/testground/testground/issues/1522

I have no opinion on how we fix this in the short term. Happy to throw more disk-space at the problem.

ddimaria:
Thanks for the reply @thomaseizinger!  It looks like @galargh will take care of CI in a separate PR, so I'll close this one out, or can wait for that to land before merging?  I agree with the pre-built binaries/images approach.  I'll read through that Testground issue.  A possible challenge might be features in Rust, though pulling the [features](https://github.com/libp2p/test-plans/blob/753cd1bd241dbf5edeed340b833ad646c1e3b232/ping/rust/Cargo.toml#L23-L36) in this PR and moving to a rust-libp2p CI build-image job (specifically for Testground?) seems sensible.

galargh:
Quick update: I didn't manage to get to fully setting up/testing self-hosted runners for this repo today but it's the first item on my TODO list for next week.

galargh:
> Quick update: I didn't manage to get to fully setting up/testing self-hosted runners for this repo today but it's the first item on my TODO list for next week.

Ready 🥳 https://github.com/libp2p/test-plans/pull/92 🎉

ddimaria:
@p-shahi I wanted to make sure this PR close was intentional?  I was waiting for https://github.com/libp2p/test-plans/pull/92 to merge in order to validate this work before merging this PR.

jxs:
Hi @ddimaria, this PR targets the old dir which has been superseeded by the new one introduced with #97. I started working on introducing WebRTC support for the Rust "test-plans" there, see [here](https://github.com/libp2p/test-plans/issues/61#issuecomment-1379063266), is that ok with you folks?

ddimaria:
Thanks for the info @jxs! This sounds good to me, I just didn't want any work to go to waste unintentionally.

PR #85, opened 2022-11-30
feat: libp2p implementation interoperability dashboard on various dimensions. Uses testplans-dsl.

MarcoPolo:
This PR adds interop testing across various dimensions (transports, muxer, secure channel) against the Go and Rust libp2p implementations. This is the foundation work on this project, and enables adding more libp2p implementations to the `versions.ts` file. This also renders the result into a markdown file to give an overview of the test results. Example result here: https://gist.github.com/MarcoPolo/965e1f6fd16df004173131199444c62d

Recommendation for reviewer:
1. Look at `ping-with-dsl/testplans.ts` first to get an idea of what the DSL looks like.
2. Review the DSL implementation in `dsl/`.
3. Look at the multidimensional testing:
  a. The `version.ts` file defines the capabilities of each version.
  b. This file is consumed by the composer and loaded into an in-memory sqlite table.
  c. The table is queried to produce the various tests we run in the form of the test-plans dsl.
  d. The DSL is then transformed to the testground files and then run.
Feedback appreciated!


ping-with-dsl/go/v0.23.0/Makefile
```diff
@@ -0,0 +1,5 @@
+image.json: Dockerfile main.go go.mod
```
MarcoPolo, line 1: Long term, I see CI building these images, uploading them to some container registry, then committing this json file directly into the repo.

ping-with-dsl/testplans.ts
```diff
@@ -0,0 +1,66 @@
...
+import { run } from "dsl/src/testground-runner"
+import goV0230 from "./go/v0.23.0/image.json"
+import rustV0490 from "./rust/image.0.49.0.json"
+
+type PingParams = {
+    max_latency_ms: number,
+    iterations: number,
+}
+
+const testplans: TestPlans<PingParams> = {
```
MarcoPolo, line 11: This is the definition of the test we are running with testground.

mxinden: Do I understand correctly that this definition will be auto-generated?

mxinden (commented):
Thanks for creating this. I am in favor of the simpler API.

Do I understand correctly that the (automated) steps now are:

1. User builds images resulting in images and `image.json` files.

2. Composer generating test plans based on `versions.ts` which itself references above `image.json` files. It does so by loading the information into a sqlite database, querying it and returning the test matrix.

3. User runs `npm run tests` which triggers the `dsl` package.
4. The `dsl` package takes the generated test matrix, transforms it into a file format that Testground understands, sets up the symlinks, and then calls the Testground binary.
5. From here Testground takes over, itself walking through its various abstraction layers.

I might be missing something. From the above, I am not sure introducing yet another layer of abstraction (dsl) is the way forward.

I understand the strategy of creating a simple API (dsl) on top of the much more complex Testground API first, and only then iteratively simplifying the Testground API. I am afraid that we do the former but never get to the latter, resulting in yet another layer.

In my eyes this project is important but not urgent. At this point I think we know of multiple changes (no symlinking, prebuild images, ...) that we want inside of Testground for sure.

Instead of adding another layer (dsl) what do you think of pushing these changes upstream (Testground) right away?

I am sure I don't have the full picture at this point. I suggest we discuss this in our 1on1 on Monday @MarcoPolo and then follow up here.

ping-with-dsl/testplans.ts
MarcoPolo: Not this one. This is a minimal example test plan to demonstrate the DSL.

MarcoPolo:
> In my eyes this project is important but not urgent. At this point I think we know of multiple changes (no symlinking, prebuild images, ...) that we want inside of Testground for sure.

I think this project is urgent in that we've committed to doing this this year, and adding more time may lead to a rabbit hole rather than solving the actual problem of creating this multidimensional test cases.

> Instead of adding another layer (dsl) what do you think of pushing these changes upstream (Testground) right away?

I want to try this API first before committing to it. I also don't want to block on first changing testground before being able to advance with this project.

MarcoPolo:
This is ready for review. I'll keep iterating on making the CI a bit better, but this does run in CI.

Recommendation for reviewer:
1. Look at `ping-with-dsl/testplans.ts` first to get an idea of what the DSL looks like.
2. Review the DSL implementation in `dsl/`.
3. Look at the multidimensional testing:
  a. The `version.ts` file defines the capabilities of each version.
  b. This file is consumed by the composer and loaded into an in-memory sqlite table.
  c. The table is queried to produce the various tests we run in the form of the test-plans dsl.
  d. The DSL is then transformed to the testground files and then run.

MarcoPolo:
@mxinden / @laurentsenta Could I get a review here when you get a chance please :)

multidim-interop/go/v0.22/Makefile
```diff
@@ -0,0 +1,5 @@
+image.json: Dockerfile main.go go.mod go.sum
+	echo "{\"imageID\": \"$(shell docker build . -q)\"}" > $@
+
+clean:
```
mxinden, line 4:
```suggestion
.PHONY: clean
clean:
```

Nit pick, this is a `PHONY` target, right?

multidim-interop/go/v0.23/Makefile
```diff
@@ -0,0 +1,5 @@
+image.json: Dockerfile main.go go.mod go.sum
+	echo "{\"imageID\": \"$(shell docker build . -q)\"}" > $@
+
+clean:
```
mxinden, line 4:
```suggestion
.PHONY: clean
clean:
```

multidim-interop/rust/manifest.toml
```diff
@@ -0,0 +1,19 @@
+name = "compatibility-rust"
```
mxinden, line 1: I wonder whether Testground still needs these. I was operating under the assumption that these are only needed to test the Rust version individually. With the DSL abstraction, I don't think this is needed, right?

multidim-interop/versions.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export type Version = {
+    id: string,
+    containerImageID: string,
+    transports: string[],
+    secureChannels: string[],
+    muxers: string[],
+}
+
+export const versions: Array<Version> = [
```
mxinden, line 15: :rocket: 

This is very clean.

.github/workflows/run-testplans-dsl.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Build images
+        working-directory: ./test-plans/${{ env.TEST_PLAN_DIR }}/
+        run: make
+      - name: Run the test
+        timeout-minutes: 120
+        working-directory: ./test-plans/${{ env.TEST_PLAN_DIR }}/
+        run: PATH=~/testground:$PATH npm run test
+      - name: Print the results
+        working-directory: ./test-plans/${{ env.TEST_PLAN_DIR }}/
+        run: cat results.csv
```
mxinden, line 50: I like this. This is easy to understand. :pray:

dsl/src/testground-runner.ts
```diff
@@ -0,0 +1,83 @@
+import { tmpdir } from 'os'
+import { promises as fs } from 'fs';
+import * as dsl from './dsl'
+import { transform } from './testground-transform'
+import path from 'path';
+import { execSync, spawn } from 'child_process';
+import { env } from 'process';
+
+export async function run(testplans: dsl.TestPlans) {
```
mxinden, line 9: Took a more in depth look here. Keeping this sounds good to me.

mxinden (commented):
Thanks @MarcoPolo for pushing this forward.

Do I understand correctly that:

- This would entirely replace the `ping/` folder? I would be in favor of that.
- `ping-with-dsl/` is only here to showcase the dsl and would be removed before merging as well?

multidim-interop/go/v0.24/main.go
```diff
@@ -0,0 +1,284 @@
+package main
```
laurentsenta, line 1: We're replacing the shims with multiple  mains:

We initially implemented "one main, multiple shims" to address the explosions of versions (https://github.com/libp2p/test-plans/pull/23#pullrequestreview-960603753).

I'm not sure what's motivating this change; I think this is a regression:
Changing the test specs, like adding benchmarking or new parameters will become exponentially difficult (N implementations x M main files).

ping-with-dsl/testplans.ts
```diff
@@ -0,0 +1,66 @@
...
+    testPlans: [
+        {
+            name: "Go v0.23",
+            instances: [
+                {
+                    name: "Go v0.23 x Go v0.23",
```
laurentsenta, line 17: I was confused by the format here, I think the instance name and test plan name are meant to be swapped

ping-with-dsl/testplans.ts
```diff
@@ -0,0 +1,66 @@
...
+                    containerImageID: rustV0490.imageID,
+                    runtimeEnv: {
+                        "max_latency_ms": 500,
+                        "iterations": 5,
+                    }
+                }
+            ]
+        }
+    ]
+}
```
laurentsenta, line 64: We're creating a DSL / configuration abstraction above Compositions:

This PR does two things: move the build step out of Testground AND introduce a new configuration format.

The new configuration looks MUCH simpler because it IS: we removed the essential complexity of the building configuration.

With a regular composition (no change to Testground required), this is what the same workflow would look like:
https://gist.github.com/laurentsenta/bde72a8ad2e8cfdcecfb3add75338a44

Turned into yaml which is more familiar than toml:
https://gist.github.com/laurentsenta/f85b2249b1598586a01071b61862e68a

I would recommend not going for a complete "DSL" abstraction yet.

laurentsenta:
Thanks for sharing and pinging for review @MarcoPolo,
I left a few comments on dropping the builds and the configuration part,

The rest looks great, especially that move from templates to ts generation!

ping-with-dsl/testplans.ts
MarcoPolo: Yeah that's right. Sorry for the confusion. Fixed

ping-with-dsl/testplans.ts
MarcoPolo: > With a regular composition (no change to Testground required), this is what the same workflow would look like:

I know this is an example, but this doesn't work today. You still need a plan and a test case. These are vestiges that I think unnecessarily complicated.

If we want to iterate to something closer to this DSL (minimal config) and change this DSL later, that's fine by me. But I don't want to block this work on first changing testground.

MarcoPolo:
> This would entirely replace the ping/ folder? I would be in favor of that.

Not yet. I think that would be a great next step, but that's not a goal right now.

> ping-with-dsl/ is only here to showcase the dsl and would be removed before merging as well?

I wasn't planning on removing it since it's useful as a stepping stone to the DSL, but I can remove it if requested.

multidim-interop/go/v0.24/main.go
MarcoPolo: I should talk with Marten. But I don't really think this is a problem in this specific test case. We want to have implementation A ping implementation B and nothing else. I bootstrapped off the existing ping cases because it was convenient.

In fact, one version of this branch had me define patches that can be applied to other versions to do something similar to the "one main, multiple shims" idea, but I realized it was needlessly complex. Once a version is out and published there shouldn't be much changes to the implementation. Users using v0.22, for example, aren't going to update their implementation to interact with newer versions. Why are we assuming we are going to go back and change the implementation of this test case in v0.22?

If we want to add benchmarking or new parameters that's a new test. Any shim we design before this new test isn't guaranteed to avoid the N x M problem.

MarcoPolo:
I think some blockers right now for this are:
- https://github.com/testground/testground/issues/1545
- https://github.com/testground/testground/issues/1547

There's a question of whether we should merge the DSL changes into testground first. I think our goal here is to have a straightforward interop testing system. To accomplish this we are only blocked on the things above. There's no reason to block on making the upstream changes. This doesn't mean that we will never upstream these changes. Likely other users of testground want a simpler configuration. But we should prioritize the things we all agreed are important in the roadmap first.

multidim-interop/go/v0.22/Makefile
MarcoPolo: Correct, thanks

multidim-interop/rust/manifest.toml
MarcoPolo: Yup, vestige of copying this code.

ping-with-dsl/testplans.ts
MarcoPolo: A couple other things when comparing this to a regular composition file:
- The DSL leverages typescript to typecheck the runtime params
- The composition file has the idea of "groups" and "runs" as separate things. This is vestige of having testground build things. But really we only need "runs" here. 

Here's both configs in JSON: https://gist.github.com/MarcoPolo/e4488ce57c58469298918af896ded192. This is subjective, but I find the DSL version simpler.

multidim-interop/go/v0.24/main.go
thomaseizinger: Once a version is out and published there shouldn't be much changes to the implementation. Users using v0.22, for example, aren't going to update their implementation to interact with newer versions. Why are we assuming we are going to go back and change the implementation of this test case in v0.22?
> 
> If we want to add benchmarking or new parameters that's a new test. Any shim we design before this new test isn't guaranteed to avoid the N x M problem.

I did the exact same thing for Rust with the same reasoning :)

Eventually I hope we can track these tests in Git by having them next to our source code which means there is only just one to maintain and it moves along with the APIs and features as we go, just the published binary remains as a test artifact and gets run by testground when requested.

p-shahi:
@Jorropo caught a bug in go-libp2p v0.24.0 where a node configured to use Noise could dial a node configured to use TLS (https://github.com/libp2p/go-libp2p/pull/1946)

In a subsequent pull request, we should extend the test cases in here to test the non happy-path (either as a part of the same github action as these or as a separate one that runs as a regression test in a nightly job.)

multidim-interop/rust/src/bin/testplan_0490.rs
```diff
@@ -0,0 +1,175 @@
...
+                    .boxed(),
+                "mplex" => authenticated
+                    .multiplex(mplex::MplexConfig::default())
+                    .timeout(Duration::from_secs(5))
+                    .boxed(),
+                _ => panic!("Unsupported muxer"),
+            }
+        }
+        _ => panic!("Unsupported"),
+    };
```
thomaseizinger, line 92: The question is, do we want to have a node that _only_ speaks the requested configuration or should it just be compatible with it.

Even in the first case, the easier option here is probably to use `OptionalTransport` and `OptionalUpgrade`. That allows you move the conditional code into a value that you can compose as if it were unconditional!

multidim-interop/composer/index.ts
```diff
@@ -0,0 +1,127 @@
+import sqlite3 from "sqlite3";
+import { open } from "sqlite";
+import { promisify } from "util";
+import { TestPlan, TestPlans } from "dsl/src/dsl";
+import { createHmac } from "crypto";
+import { Version } from "../versions";
+
+
+export async function buildTestplans(versions: Array<Version>): Promise<TestPlans> {
```
thomaseizinger, line 9: It would be nice to have a test for this.

ping-with-dsl/go/v0.23.0/Makefile
thomaseizinger: I share the same vision apart from the committing to the repo part. Can you elaborate why do you think that is necessary?

dsl/src/testground-runner.ts
```diff
@@ -0,0 +1,83 @@
...
+    // Check if directory exists
+    try {
+        await fs.access(dir)
+        await fs.rm(dir, { recursive: true, force: true })
+    } catch (e) {
+    }
+    await fs.mkdir(dir)
+    await fs.mkdir(path.join(dir, "testplans"))
+
+    const filesToCreate = transform(testplans)
```
thomaseizinger, line 21: Within `transform` we assume that all containers have been built already (which they are currently via the Makefile). What do you think of moving the building of the images also into here?

I am asking because a next step for this could be to pull images instead of building them but that would probably be a bit clunky to express in the Makefile so it might be better done in here.

thomaseizinger: https://github.com/testground/testground/issues/1522 is what I am aiming at!

multidim-interop/versions.ts
```diff
@@ -0,0 +1,51 @@
...
+        containerImageID: rustv049.imageID,
+        transports: ["tcp", "ws"],
+        secureChannels: ["noise"],
+        muxers: ["mplex", "yamux"],
+    },
+    {
+        id: "rust-v0.50.0",
+        containerImageID: rustv050.imageID,
+        transports: ["ws", "tcp", "quic-v1"],
+        secureChannels: ["noise"],
```
thomaseizinger, line 27:
```suggestion
        secureChannels: ["tls", "noise"],
```

:innocent:

ping-with-dsl/go/v0.23.0/Makefile
thomaseizinger: In general, I don't quite understand the purpose of the JSON files.

ping-with-dsl/testplans.ts
thomaseizinger: When would we run the tests defined here vs the multi-dimensional matrix?

thomaseizinger: It seems to be fairly excessive for an example, in terms of LoC. Did you plan on merging this?

thomaseizinger (commented):
I like the direction this is going, left a couple of comments :)

multidim-interop/go/v0.24/main.go
laurentsenta: Thanks for sharing detailed answers!

> Users using v0.22, for example, aren't going to update their implementation to interact with newer versions. Why are we assuming we are going to go back and change the implementation of this test case in v0.22?

> there is only just one to maintain and it moves along with the APIs and features as we go,

> If we want to add benchmarking or new parameters that's a new test. Any shim we design before this new test isn't guaranteed to avoid the N x M problem.

Got it, I see the nuance.

I was starting from the constraints: we want to add and update tests often. New tests should cover _most_ known versions, even old ones, as easily as possible. The team wants to write the minimum amount of test code.

The new approach was itching because it seems we end up with a lot of duplicated code and we'll have to revisit all these mains (and later release branches) whenever the spec changes. Whereas with shims, we can update one `main.go` (the spec) and it applies to every known version.

That makes sense now, we're less worried about updating test specs and reusing code because we start from the assumptions: test specs won't change often, and most of these specs will be written for the current and future versions.

Thanks for the clarification.

ping-with-dsl/testplans.ts
laurentsenta: > I know this is an example, but this doesn't work today. You still need a plan and a test case. These are vestiges that I think unnecessarily complicated.

The example would work, except for a typo with artifacts maybe.
I'm confused because we could do:

```
params -- (generator) --> composition
```

instead of:

```
params -- (generator) --> "dsl" -- (transformer) --> composition
```

- typing is supported just as well, type the boundaries of your generator, you could even have fun with generics there,
- the main change in the DSL is that we're going to duplicate every group, so code generation will be the default, no one is going to manually edit this file.
- there's nothing to block for in Testground.

It seems more efficient. Let's have a chat async on why we can't do this and why we call this a DSL instead of a new configuration format, Feel free to resolve this whenever you want.

ping-with-dsl/testplans.ts
MarcoPolo: Happy to remove this if/when we decide we want to use this DSL for ping tests.

ping-with-dsl/testplans.ts
MarcoPolo: > The example would work, except for a typo with artifacts maybe.

At the very least you currently need a `manifest.toml` that defines a the plan name, defaults, builders, runners, and testcases.

> the main change in the DSL is that we're going to duplicate every group, so code generation will be the default, no one is going to manually edit this file.

I don't follow, sorry.

> there's nothing to block for in Testground.

What's blocking now?

>  why we call this a DSL instead of a new configuration format,

A rose by any other name. Happy to call this a new configuration format.

multidim-interop/rust/src/bin/testplan_0490.rs
MarcoPolo: We want something that _only_ speaks the requested configuration. That way we are sure we are testing the requested configuration and not accidentally making something else work.

This code is definitely hacked together. I'll take a look at the OptionalTransport, thanks!

MarcoPolo:
@thomaseizinger I haven’t started investigating this issue yet, but you might have a lot more context here and can probably help me out a ton. I’m seeing flakiness in the Rust tests. This shows up mostly in the rust x go, but also a rust x rust test (https://github.com/libp2p/test-plans/actions/runs/3798668092 scroll to summary below). And has shown up before this test [here](https://github.com/testground/testground/issues/1542#issuecomment-1352886909). 

At first I suspect it’s a testground networking issue, but it’s curious that it doesn’t affect the Go side. One total guess is that the rust test isn’t properly waiting for the network to be ready.

dsl/src/testground-runner.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    await Promise.all(
+        filesToCreate.files.map(({ filename, filecontents }) =>
+            fs.writeFile(path.join(dir, "testplans", filename), filecontents)
+        )
+
+    )
+    await fs.writeFile(path.join(dir, ".env.toml"), `
+[daemon.scheduler]
+task_timeout_min = 2
```
MarcoPolo, line 56: Would be better for this to be configured in the DSL rather than hardcoded here.

dsl/src/testground-runner.ts
MarcoPolo: I think it's best if we don't try to build here. I think that's a sizable bit of complexity in testground. The Makefile is just one strategy to build the container image. Using Nix can be another. Maybe we want to use [warpforge](https://github.com/warpfork/warpforge) at some point.

It's nice that this just has to reference some existing image id, and then it'll work.

multidim-interop/go/v0.24/main.go
marten-seemann: I have mixed feelings on this one. I see the argument that existing test plans will not be updated frequently, although I do expect us to make changes to the ping test plan, 300 LOC are just waaaay too complex for a simple ping test. For example, I'd like to see the network configuration being moved outside of the test plan - this doesn't need to be a runtime configuration. We can also reduce complexity by making this a test that only 2 hosts are participating in explicitly.

I imagine that future tests will share a fair bit of code with the ping test, e.g. constructing a host with a certain transport configuration. I can see us introducing helper functions _per version_ (and not _cross_ version, as a shim would do) to avoid code duplication.

Just my 2 cents, definitely not a blocker for this PR.

multidim-interop/composer/index.ts
```diff
@@ -0,0 +1,127 @@
...
+                     AND b.id == sb.id
+                     AND a.id == ha.id
+                     AND b.id == hb.id
+                     AND a.transport == b.transport
+                     AND sa.sec == sb.sec
+                     AND ma.muxer == mb.muxer
+                     AND ha.hash < hb.hash
+                     -- quic only uses its own muxer/securechannel
+                     AND a.transport != "quic"
+                     AND a.transport != "quic-v1";`);
```
marten-seemann, line 62: This also applies to webtransport, doesn't it?

multidim-interop/go/v0.24/main.go
MarcoPolo: >  I can see us introducing helper functions per version (and not cross version, as a shim would do) to avoid code duplication.

:+1:

As a data point, I'm trying to fix a [race condition](https://github.com/libp2p/test-plans/pull/85/commits/102f01fcec39ce6c634bbf961f9b8a102e74a4dd#diff-324e6c18346c026bb1db7c265dfd87cdcca67e99f52a317966321453187e1505R223) in our existing Go ping test-plan and using the compat shim is way harder than doing it once per version. In this branch I did the once per version and was done in <5 minutes. With the compat shim, I can't even run `go build` to see if I did the right thing.

multidim-interop/versions.ts
marten-seemann: I really like this as well, this is great!

marten-seemann:
> > ping-with-dsl/ is only here to showcase the dsl and would be removed before merging as well?
> 
> I wasn't planning on removing it since it's useful as a stepping stone to the DSL, but I can remove it if requested.

I think I share @mxinden's question here. The multidimensional configuration looks, whereas ping-with-dsl contains the `ping-with-dsl/testplans.ts` file, which in my understanding should be an intermediate step and not committed to VCS (I might be misunderstanding things here).

@MarcoPolo Can you clarify the relationship between ping-with-dsl and multidim?

ping-with-dsl/go/v0.23.0/Makefile
MarcoPolo: > Can you elaborate why do you think that is necessary?

Depends on our use case, but maybe for this it's not worth it. It would be nice if you could pull the images built in CI if your code doesn't affect these things. It's like using it as a cache. 

> In general, I don't quite understand the purpose of the JSON files.

It's the simplest way to record the container image id. Since it's json we can import it in our typescript code directly as opposed to reading and parsing a file manually.

multidim-interop/composer/index.ts
MarcoPolo: I don't understand. This whole thing is the test, isn't it?

multidim-interop/composer/index.ts
MarcoPolo: yes. Although the quic-v1/webtransport vs quic/webtransport change is why I haven't added webtransport yet here.

MarcoPolo:
Lets get rid of ping-with-dsl. It was mostly used as an example for now and I think it complicates things.

multidim-interop/rust/src/bin/testplan_0490.rs
MarcoPolo: @jxs Do you think you can help me clean this up? You'll certainly do this faster than me.

MarcoPolo:
@laurentsenta / @mxinden CI is passing and I've done a pass at some cleanup and addressed comments. Could I get another review here please?

In a follow up PR I'm going to:
- Add the distinction of A dials B and B dials A.
- Add a test where A dials A.

multidim-interop/rust/src/bin/testplan_0490.rs
MarcoPolo: I switched to using an EitherUpgrade

mxinden (commented):
Other than the missing go-libp2p and rust-libp2p versions, this looks good to me.

@jxs in case you have some capacity, can you help with the outstanding refactorings in the Rust transport construction? If not, I am happy to jump in.

.github/actions/setup-testground/action.yml
```diff
@@ -50,6 +53,7 @@ runs:
       shell: bash
 
     - name: Run the daemon or configure the client
+      if: ${{ inputs.start_testground == 'true' }}
```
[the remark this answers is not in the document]
mxinden: > The existing ping tests use the `run-composition` GH action which depends on this being true.

Do I understand correctly that long term that GH action would go away? Do I also understand correctly that we can not yet remove it here given that that would break CIs in rust-libp2p and go-libp2p?

For the sake of cleaning it up right away, would that breakage of our CI be that bad?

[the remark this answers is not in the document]
MarcoPolo: > Do I understand correctly that long term that GH action would go away?

I wonder if there's still value in these existing ping tests. We get most of the value from them in the new multidim interop tests. The other value is having a fast test that checks if current master works against the other implementations.

>  Do I also understand correctly that we can not yet remove it here given that that would break CIs in rust-libp2p and go-libp2p?

Yeah, and we wouldn't have anything to replace them.

> For the sake of cleaning it up right away, would that breakage of our CI be that bad?

We don't have a plan right now to replace the existing ping tests. I'd rather focus on replacing it and then update the other repo's CI rather than breaking CI and then forcing us to fix it in a reactive way.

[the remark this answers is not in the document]
mxinden: > The other value is having a fast test that checks if current master works against the other implementations. 

That would as well happen via the multidimensional tests, no?

> Right now there's still some value in the ping tests that isn't captured by this PR.

What value would they provide long term?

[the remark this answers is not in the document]
MarcoPolo: > That would as well happen via the multidimensional tests, no?

Long term yes. Right now, no.

> What value would they provide long term?

Long term, we should get rid of them. Right now though, I think the priority is to have the multidim interop tests. Replacing the ping tests imo is lower priority.

What the ping test gives us that multidim doesn't right now:
* CI testing in the impl repo.
* Interop testing with Nim (I haven't spent any time getting Nim into the multidim)

Long term, multidim will do this and we can get rid of the ping tests. I would save that for a future PR rather than add to this one. And I would avoid removing the existing ping tests until we're ready to replace them.

MarcoPolo:
I'm a bit worried on the scalability of testground for this. Looking at the last run: https://github.com/libp2p/test-plans/actions/runs/3833933397/jobs/6525831766 each test took 5.5 seconds.

I can reduce the time a bit here by only doing a single ping instead of the two currently. and there are probably some other optimizations I can make. What would be really nice is to keep the simplicity of two instances where A pings B (not a test where we have many instances all pinging each other), have these run faster, and maybe have these run in parallel. What can we do to reduce the 5.5 seconds? I'm not sure why this ping test would take more than a second to complete. So my guess here is that there is some overhead with testground for each test.

Jorropo:
> caught a bug in go-libp2p v0.24.0 where a node configured to use Noise could dial a node configured to use TLS
@p-shahi 

No the dial didn't worked, the bug I caught was the error was shadowed (so it was returning a `nil` error even tho it failed.)

.github/workflows/run-testplans-dsl.yml
```diff
@@ -0,0 +1,79 @@
...
+        working-directory: ./test-plans/${{ env.TEST_PLAN_DIR }}/
+        run: cat ./dashboard.md >> $GITHUB_STEP_SUMMARY
+      - name: Exit with Error
+        working-directory: ./test-plans/${{ env.TEST_PLAN_DIR }}/
+        run: |
+          if grep -q ":red_circle:" ./dashboard.md; then
+            exit 1
+          else
+            exit 0
+          fi
```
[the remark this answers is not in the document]
MarcoPolo: I'll defer this to a future change.

p-shahi:
closing in favor of #97

PR #87, opened 2022-12-01
feat(ping/rust): Add 0.50.0 as a release and fix deprecation warnings

thomaseizinger:


ping/_compositions/rust.toml
```diff
@@ -1,7 +1,11 @@
 [master]
-BinaryName = 'testplan_0500'
+BinaryName = 'testplan_0510'
```
mxinden, line 2: Neat trick. Just need to remember that this is not actually `v0.51.0` (yet).

mxinden (approved):
Thanks @umgefahren for starting this and thanks @thomaseizinger for finishing it.

ping/_compositions/rust.toml
thomaseizinger: We should probably rename it to `testplan_master`. Really the name doesn't matter :)

PR #92, opened 2022-12-30
Run test plans on self-hosted GHA machines

galargh:
###### Description

This PR moves GHA workflows that are defined in this repository to self-hosted runners. The runners definition can be found here: https://github.com/pl-strflt/tf-aws-gh-runner/blob/main/runners.tf#L28

###### Why?

To address issues with:
- storage needs
- speed
- flakiness

###### How?

- more powerful [machine](https://github.com/pl-strflt/tf-aws-gh-runner/blob/main/runners.tf#L28)
- ephemeral runners
- dockerhub registry proxy backed by S3
- docker buildx cache backed by S3 (testground doesn't take advantage of this... yet; https://github.com/testground/testground/issues/1554)

###### Testing

- [x] CI (running on self-hosted runners)

### DIFF
--- a/.github/workflows/multidim-interop.yml
+++ b/.github/workflows/multidim-interop.yml
@@ -1,6 +1,9 @@
 on:
   workflow_dispatch:
   pull_request:
+  push:
+    branches:
+      - "master"
 
 name: libp2p multidimensional interop test
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Testground test plans for libp2p
 
+[![Interop Dashboard](https://github.com/libp2p/test-plans/actions/workflows/multidim-interop.yml/badge.svg?branch=master)](https://github.com/libp2p/test-plans/actions/workflows/multidim-interop.yml?query=branch%3Amaster)
+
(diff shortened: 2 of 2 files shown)


.github/workflows/run-composition.yml
```diff
@@ -28,10 +28,31 @@ on:
...
+  choose_runner:
+    runs-on: ubuntu-latest
+    outputs:
+      config: ${{ steps.config.outputs.result }}
+    steps:
+      - uses: actions/github-script@v6
+        id: config
+        with:
+          script: |
+            if (`${context.repo.owner}/${context.repo.repo}` === 'libp2p/test-plans') {
```
laurentsenta, line 42: That approach looks great. Do we need to take into account the fact that these jobs are used through reusable workflows?

https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-github-hosted-runners

galargh: If I understand the docs correctly, the context here will reference the `caller` repo which is exactly what we want. So if we call the reusable workflow from `libp2p/test-plans`, then we're going to use self-hosted runners. Otherwise, we'll fall back to the gh hosted ones.

PR #93, opened 2022-12-31
Fix race in ping test case for Go

MarcoPolo:
There's a race issue where the local node emits the connected event, even if it doesn't have connections to all peers. This shows up in rust x go tests quite often.

Dealing with the shim layer was quite tough and involved a very slow feedback cycle. A simple `go build` doesn't work so you have to try against testground, which takes a while.

### DIFF
--- a/ping/go/compat/libp2p.v0.11.go
+++ b/ping/go/compat/libp2p.v0.11.go
@@ -6,11 +6,10 @@ package compat
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/libp2p/go-libp2p"
-	"github.com/libp2p/go-libp2p-core/event"
 	"github.com/libp2p/go-libp2p-core/host"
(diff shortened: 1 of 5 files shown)


marten-seemann (commented):
Why is there a race condition? Shouldn't `host.Connect` block until the connection has been established?

MarcoPolo:
A calls `host.Connect` to B and blocks until it has a connection. Then emits the "connected" event.

B waits for the "connected" event and then continues. At the next line in B, we may not have the connection yet on B's side event though A sees that it has a connection to B.

marten-seemann (approved):
Makes sense. This is subtle!

I hope we can rid of this synchronization logic by simplifying the ping test case (#94).

MarcoPolo:
You don’t actually need the connected event. Each side can simply wait to have enough connections before continuing.

PR #97, opened 2023-01-06
Add fast multidimensional interop tests

MarcoPolo:
This is an alternative version to #85 . This uses https://compose-spec.io to define the tests and `docker compose` to run the test. 

On my machine this is 10x faster than testground. On CI, this is about 3.5x faster (I've done no optimizations on either local or CI, but we could probably parallelize the tests more if the CI instances were beefier).

The basic idea here:

1. Read the `versions.ts`
2. Generate the combination of tests we want to run (A dials B using X Y Z).
3. Generate a compose specification from that (compose.yaml)
4. Ask `docker compose` to run this.
5. Collect the results and render them.

The entry point for everything is `testplans.ts` and you can run this locally by doing `npm run test` (assuming you have docker up).

This is not a replacement of testground. Instead this is using common tools to run these interop tests faster. I make no claim here that this replaces testground. That being said, I found working completely outside of testground a nicer and faster experience. Compose is a well trodden tool with lots of documentation available. I found `compose.yaml` spec good and simple enough for what I need here.

### Compared to #85

* Doesn't use testground
* No DSL. I think the https://compose-spec.io is simple and good enough for this. And more widely understood than a custom DSL.
* Rust version not implemented (I wanted to get the tests running first to compare this with #85)
* Faster
* Simpler. While there are a couple steps, each step stands alone. You can inspect a generated `compose.yaml` and inspect it or manually copy it and tweak it. No custom sync service, just a plain redis store and redis client for synchronization.
* You can even run the tests manually locally without docker. Doing this with a testground test is fairly hard.
* This tests that A dials B and B dials A separately.
* Tests that A can dial A.

---

Don't let the 13k diff scare you. Most of it is lock files `go.sum` or `package-lock.json`. The important bits to review here are the Typescript files in `multidim-interop` and the Go implementation of ping.


.github/actions/setup-testground/action.yml
```diff
@@ -10,6 +10,9 @@ inputs:
   testground_ref:
     required: false
     default: 'edge'
+  start_testground:
```
MarcoPolo, line 13: I'll remove this and revert changes to this file.

multidim-interop/go/v0.23/main.go
```diff
@@ -0,0 +1,146 @@
...
+	rClient := redis.NewClient(&redis.Options{
+		Addr:     redis_addr,
+		Password: "",
+		DB:       0,
+	})
+	defer rClient.Close()
+
+	for {
+		// Wait for redis to be ready
+		_, err := rClient.Ping(ctx).Result()
```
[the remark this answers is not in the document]
MarcoPolo: Cool! I’ll use that. Thanks

multidim-interop/go/v0.24/main.go
```diff
@@ -0,0 +1,149 @@
...
+
+		fmt.Println("Ping successful: ", res.RTT)
+
+		rClient.RPush(ctx, "dialerDone", "").Result()
+	} else {
+		_, err := rClient.RPush(ctx, "listenerAddr", host.Addrs()[0].Encapsulate(ma.StringCast("/p2p/"+host.ID().String())).String()).Result()
+		if err != nil {
+			panic("Failed to send listener address")
+		}
+		rClient.BLPop(ctx, 4*time.Second, "dialerDone").Result()
```
jxs, line 147:
```suggestion
		_, err := rClient.BLPop(ctx, 4*time.Second, "dialerDone").Result()
		if err != nil {
			panic("Failed to wait for dialer conclusion")
		}
```

multidim-interop/go/v0.23/main.go
```diff
@@ -0,0 +1,146 @@
...
+
+		fmt.Println("Ping successful: ", res.RTT)
+
+		rClient.RPush(ctx, "dialerDone", "").Result()
+	} else {
+		_, err := rClient.RPush(ctx, "listenerAddr", host.Addrs()[0].Encapsulate(ma.StringCast("/p2p/"+host.ID().String())).String()).Result()
+		if err != nil {
+			panic("Failed to send listener address")
+		}
+		rClient.BLPop(ctx, 4*time.Second, "dialerDone").Result()
```
jxs, line 144:
```suggestion
		_, err := rClient.BLPop(ctx, 4*time.Second, "dialerDone").Result()
		if err != nil {
			panic("Failed to wait for dialer conclusion")
		}
```

multidim-interop/go/v0.22/main.go
```diff
@@ -0,0 +1,142 @@
...
+
+		fmt.Println("Ping successful: ", res.RTT)
+
+		rClient.RPush(ctx, "dialerDone", "").Result()
+	} else {
+		_, err := rClient.RPush(ctx, "listenerAddr", host.Addrs()[0].Encapsulate(ma.StringCast("/p2p/"+host.ID().String())).String()).Result()
+		if err != nil {
+			panic("Failed to send listener address")
+		}
+		rClient.BLPop(ctx, 4*time.Second, "dialerDone").Result()
```
jxs, line 140:
```suggestion
		_, err := rClient.BLPop(ctx, 4*time.Second, "dialerDone").Result()
		if err != nil {
			panic("Failed to wait for dialer conclusion")
		}
		
```

multidim-interop/go/v0.22/main.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/libp2p/go-libp2p/p2p/transport/websocket"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+func main() {
+	var (
+		transport     = os.Getenv("transport")
+		secureChannel = os.Getenv("security")
+		muxer         = os.Getenv("muxer")
+		is_dialer_str = os.Getenv("is_dialer")
```
marten-seemann, line 28: Nit: No underscores in Go variable names.

multidim-interop/go/v0.22/main.go
```diff
@@ -0,0 +1,142 @@
...
+		ip            = os.Getenv("ip")
+		redis_addr    = os.Getenv("REDIS_ADDR")
+	)
+
+	if redis_addr == "" {
+		redis_addr = "redis:6379"
```
marten-seemann, line 34: I assume this is for when you run this locally, without Docker?

multidim-interop/src/compose-runner.ts
```diff
@@ -0,0 +1,59 @@
...
+        renewAnonVolumes?: boolean
+    }
+}
+
+export type RunFailure = any
+
+export async function run(namespace: string, compose: ComposeSpecification, opts: RunOpts): Promise<RunFailure | null> {
+    // sanitize namespace
+    namespace = namespace.replace(/[^a-zA-Z0-9]/g, "-")
+    const dir = path.join(tmpdir(), "compose-runner", namespace)
```
marten-seemann, line 24: If I understand correctly `tmpdir()` only returns the OS' default tmp directory. Is there any function to create a new temporary directory, like `os.MkdirTemp` in Go? Then you could also remove the check if the directory exists below.

multidim-interop/go/v0.22/main.go
MarcoPolo: yup. Give the option to change with a good default

multidim-interop/src/compose-runner.ts
MarcoPolo: There might be, but this is nice because you make sure to clean up old things in case of a crash that didn't get time to cleanup.

GlenDC:
Added JS support for this branch in https://github.com/libp2p/test-plans/pull/98/files. Ready for review @MarcoPolo .

mxinden (commented):
Summary: I suggest we continue without Testground, that is continue on this pull request.

What does Testground offer that WO-Testground doesn't?

- Testground provides network traffic shaping out of the box, but we may be able to achieve the same using Docker and the Compose specification. Additionally, our current Ping tests only ensure that the Ping still makes it across the wire, without thoroughly testing the impact of network shaping.
- Testground offers orchestration across multiple machines via Kubernetes, but the Compose specification may be able to run a single Compose YAML across multiple machines as well.

Highlighting my main arguments for WO-Testground

- It is significantly easier to use, as described in the main Pull Request description.
- It is well-specified and supported, as it builds on top of Compose.

Miscellaneous
- As proposed by @thomaseizinger, I would like the testing logic to be located in the same repository as the code it is testing in the long term. For example, the testplan_0500.rs code should be in the github.com/libp2p/rust-libp2p repository on the v0.50 branch. I don't see any problems with either Testground or WO-Testground in this regard.

Wish for next steps

- I would like to minimize the timeframe in which we use both Testground and WO-Testground. In other words, I would prefer a quick transition from Testground to WO-Testground. I expect maintaining both over a longer period of time to be a high maintenance burden. If I understand WO-Testground correctly, it already includes all the features that we currently use in Testground (excluding network shaping, which I am fine with dropping). See also the discussion on the multidimensional Testground pull request at https://github.com/libp2p/test-plans/pull/85#discussion_r1061725999.
- Proceeding in multiple pull requests is fine by me. Great to see a JS pull request already open https://github.com/libp2p/test-plans/pull/98.
- @marcopolo just a wish. You know best how to proceed.

@MarcoPolo let me know how I can be helpful here. If I understand correctly @jxs will help with the Rust side for now? Is that fine @jxs and @MarcoPolo?

Thanks @marcopolo for pushing this forward!

mxinden (approved):
My preference would be for this pull request to atomically replace all other tests, e.g. the existing Ping tests. That said, this is not a strong preference. I suggest to go with whatever strategy @MarcoPolo proposes.

I am fine proceeding here, i.e. merging here and doing the remaining work (e.g. porting all rust-libp2p versions, updating the CI scripts) in a follow-up pull request. (As discussed, please remove v0.49.0 before merging as it is still using Testground.)

Do I understand correctly that we have consensus to proceed without Testground? @MarcoPolo in case you merge here today, have you talked to everyone involved to give folks a chance to include their arguments in this discussion?

MarcoPolo:
> Do I understand correctly that we have consensus to proceed without Testground? @MarcoPolo in case you merge here today, have you talked to everyone involved to give folks a chance to include their arguments in this discussion?

Yes. I think most of the libp2p team is aligned here. I don't know of anyone who is against this. I've talked with Laurent and he agrees this makes sense for this test.

@p-shahi suggested to make a small writeup to summarize the reasoning here for historical purposes. I'm happy to do that in a standalone issue after merging this PR.

PR #99, opened 2023-01-10
Support other repos using the multidim interop tests

MarcoPolo:
Features:
- Filter tests with a flag
- Load container image from runtime param.
- Add extra versions with a runtime parameter. This allows other repos to define their "head" version to test against.
- Add emit only option, which always users to emit the compose spec that they can then manually run.

How to use this with other repos:
1. In a job, build a container image. [Example](https://github.com/libp2p/go-libp2p/pull/1991/files#diff-f6e386535c20db77aa925c6a8ed443534c486443f73ca333f77ef75493698128R15)
2. Upload the container image as an artifact (this is the only way to pass data between jobs in GH Actions). [Example](https://github.com/libp2p/go-libp2p/pull/1991/files#diff-f6e386535c20db77aa925c6a8ed443534c486443f73ca333f77ef75493698128R23)
3. Run this workflow from the other repo as another job that depends on (i.e. `needs`) the previous job. And pass in the proper input parameters. [Example](https://github.com/libp2p/go-libp2p/pull/1991/files#diff-f6e386535c20db77aa925c6a8ed443534c486443f73ca333f77ef75493698128R28)

Example run that tests across versions and impl for go-libp2p-head: https://github.com/libp2p/go-libp2p/actions/runs/3894748269


MarcoPolo:
tagging either @mxinden or @marten-seemann for review please :)

.github/actions/setup-testground/action.yml
```diff
@@ -1,82 +0,0 @@
-name: start testground
```
marten-seemann, line 1: 🎉

.github/workflows/run-testplans.yml
```diff
@@ -7,6 +7,24 @@ on:
...
         required: true
         type: string
+      extra-versions:
+        description: artifact name for the extra-versions.json file
+        required: false
+        type: string
+      image-tar:
+        description: artifact name for the image.tar(s) to import
+        required: false
+        type: string
```
marten-seemann, line 17: Similar to my comment on the go-libp2p PR, could this be merged into a single option if we make use of tags?

MarcoPolo: No. This was a misunderstanding.

mxinden:
> Add emit only option, which always users to emit the compose spec that they can then manually run.

:rocket:

mxinden (approved):
Steps and code looks good to me. That said, I am not a GitHub action expert. I would assume Laurent and Piotr would be better as reviewers.

@jxs would you mind doing the changes of https://github.com/libp2p/go-libp2p/pull/1991/ for rust-libp2p?

@MarcoPolo I don't have an opinion on whether to block on https://github.com/libp2p/go-libp2p/pull/1991/ in rust-libp2p, or merge early here.

Great to see this happening. Thanks @MarcoPolo.

MarcoPolo:
@galargh do you mind reviewing the GitHub workflow here?

galargh (commented):
I reviewed `.github/workflows/run-testplans.yml` only. One note on `download-artifact` usage. Otherwise, looks good.

.github/workflows/run-testplans.yml
```diff
@@ -18,15 +36,30 @@ jobs:
...
+      - uses: actions/download-artifact@v3
+        if: ${{ inputs.extra-versions != '' }}
+        with:
+          name: ${{ inputs.extra-versions }}
+          path: /tmp/extra-versions
+      - uses: actions/download-artifact@v3
+        if: ${{ inputs.image-tar != '' }}
+        with:
+          name: ${{ inputs.image-tar }}
+          path: /tmp/images/
```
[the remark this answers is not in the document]
MarcoPolo: Cool tip! Not needed for this PR, but I can imagine it being useful in other cases. I guess you have to keep in mind the 90 day limit of artifacts.

thomaseizinger:
> @jxs would you mind doing the changes of [libp2p/go-libp2p#1991](https://github.com/libp2p/go-libp2p/pull/1991) for rust-libp2p?
> 
> @MarcoPolo I don't have an opinion on whether to block on [libp2p/go-libp2p#1991](https://github.com/libp2p/go-libp2p/pull/1991) in rust-libp2p, or merge early here.

Currently, we can't merge any PRs until this is fixed. See for example: https://github.com/libp2p/rust-libp2p/pull/3324

PR #100, opened 2023-01-12
Add webrtc transport multidim interop

jxs:


mxinden:
:rocket: let me know once this is ready for a review.

jxs:
@mxinden thanks, ready! @MarcoPolo the tests are failing because of the ["js-v0.41.0 x js-v0.41.0 (ws, noise, yamux)"](https://github.com/libp2p/test-plans/actions/runs/3925888356/jobs/6711098581#step:12:136) which this PR doesn't affect afaict.

MarcoPolo:
Can you rebase please. The ping succeeded, but the test failed. I think it’s because it failed when closing the Redis client (which is fine).

multidim-interop/rust/src/bin/testplan_0500.rs
```diff
@@ -1,70 +1,58 @@
...
     };
 
-    authenticated
-        .multiplex(mux_upgrade)
-        .timeout(Duration::from_secs(5))
-        .boxed()
+    let timeout = Duration::from_secs(5);
+
+    match secure_channel_param {
+        "noise" => builder
```
mxinden, line 41: As an aside, I suggest using `enum`s here long-term instead of passing around strings. `clap` should make this reasonably easy. Neither important nor urgent.

multidim-interop/src/generator.ts
```diff
@@ -68,6 +69,12 @@ export async function buildTestSpecs(versions: Array<Version>): Promise<Array<Co
                      WHERE a.transport == b.transport
                      -- Only webtransport transports
                      AND a.transport == "webtransport";`);
+    const webrtcQueryResults =
+        await db.all(`SELECT DISTINCT a.id as id1, b.id as id2, a.transport
+                     FROM transports a, transports b
+                     WHERE a.transport == b.transport
+                     -- Only webtransport transports
```
mxinden, line 76:
```suggestion
                     -- Only webrtc transports
```

multidim-interop/versions.ts
```diff
@@ -16,7 +16,7 @@ export const versions: Array<Version> = [
     {
         id: "rust-v0.50.0",
         containerImageID: rustv050.imageID,
-        transports: ["ws", "tcp", "quic-v1"],
+        transports: ["ws", "tcp", "quic-v1", "webrtc"],
```
mxinden, line 19: :rocket:

mxinden (approved):
Minor comments. Good to merge from my end.

jxs:
CC @thomaseizinger you also probably want to take a look, as this builds on your work on the previous interop tests

multidim-interop/rust/src/lib.rs
```diff
@@ -1,13 +1,44 @@
...
 
+/// Supported transports by rust-libp2p.
+#[derive(Clone, Debug, EnumString)]
+#[strum(serialize_all = "kebab-case")]
+pub enum Transport {
+    Tcp,
+    QuicV1,
+    Webrtc,
+    Ws,
+}
```
mxinden, line 19: With `clap` you could do something along the lines of:

``` rust
#[derive(clap::ValueEnum, Clone)]
pub enum Transport {
    Tcp,
    QuicV1,
    Webrtc,
    Ws,
}

#[derive(Debug, Parser)]
#[clap(name = "rust-libp2p Interop Tests")]
struct Opts {
    /// The transport to listen on.
    #[clap(long)]
    transport: Transport,
}

fn main() {
    let opts = Opts::parse();
}
```

(Haven't actually tried the above.)

We are using `clap` in many places anyways, thus consistent across the ecosystem. It gives you nice parser errors. It gives us a `--help` text for free.

@jxs what do you think? I don't have a strong opinion on it. Feel free to ignore entirely or delay for another pull request.

jxs: thanks for the input Max, and sorry I should have probably justified the decision. 
I actually started by following your suggestion with `clap`. But the thing is, your example assumes the variables are passed by command line options, but in our case are passed by environment variables. `clap` allows to address that with the `env` feature, but then `clap` expects the name of the variable to be uppercase, whereas in our case (transport) it's in lower case. 
I tried using [`rename_all`](https://docs.rs/clap/4.1.1/clap/_derive/index.html#arg-attributes) but it seems it only applies to the variable value. 

Then as I noticed in our case we just need to convert from `String`s to `Enum`s I resorted to `strum`.
If you prefer I can try to see if using the `clap` builder we can try to achieve the same.

multidim-interop/src/generator.ts
```diff
@@ -79,14 +86,27 @@ export async function buildTestSpecs(versions: Array<Version>): Promise<Array<Co
...
+        quicQueryResults
+        .concat(quicV1QueryResults)
+        .concat(webtransportQueryResults)
+        .map((test): ComposeSpecification => buildSpec(containerImages, {
+            name: `${test.id1} x ${test.id2} (${test.transport})`,
+            dialerID: test.id1,
+            listenerID: test.id2,
+            transport: test.transport,
+            muxer: "quic",
+            security: "tls",
```
jxs, line 99: @MarcoPolo what do you think of this change? It doesn't break Go `quic` tests, and allows Rust to have one less Variant in the enum. Plus I'd say it makes sense semantically.

MarcoPolo: I don’t think it makes sense. In every other context tls means add a tls encryption layer on top of the transport.

I know quic uses tls underneath and that isn’t a hardcoded decision (some different encryption layer could be possible in the future). But for now I think we should keep tls meaning add a tls security layer to the transport.

multidim-interop/rust/src/bin/testplan_0500.rs
```diff
@@ -1,144 +1,148 @@
...
+
+    let secure_channel_param: testplan::SecProtocol = env::var("security")
+        .context("security environment variable is not set")?
+        .parse()
+        .context("unsupported secure channel")?;
+
+    let muxer_param: Muxer = env::var("muxer")
+        .context("muxer environment variable is not set")?
+        .parse()
+        .context("unsupported muxer")?;
```
thomaseizinger, line 72: Until we move the test to our repository (where we deduplicate the test using Git revisions), I don't think it makes much sense to be typed here because we have to repeat it in every test.

multidim-interop/src/generator.ts
```diff
@@ -79,14 +86,27 @@ export async function buildTestSpecs(versions: Array<Version>): Promise<Array<Co
             muxer: test.muxer,
             security: test.sec,
         })
-    )).concat(quicQueryResults.concat(quicV1QueryResults).concat(webtransportQueryResults).map((test): ComposeSpecification => buildSpec(containerImages, {
-        name: `${test.id1} x ${test.id2} (${test.transport})`,
-        dialerID: test.id1,
-        listenerID: test.id2,
-        transport: test.transport,
-        muxer: "quic",
-        security: "quic",
```
thomaseizinger, line 88: This is a bit odd, I think leaving these blank would make the most sense. It requires the tests to parse the variables in steps but I don't see why that is necessarily a problem.

multidim-interop/rust/src/bin/testplan_0500.rs
```diff
@@ -1,144 +1,148 @@
...
-use libp2pv0500::swarm::SwarmEvent;
-use libp2pv0500::websocket::WsConfig;
-use libp2pv0500::*;
-use std::collections::HashSet;
-use std::env;
-use std::time::Duration;
-use testplan::{run_ping_redis, PingSwarm};
+use testplan::{run_ping, Muxer, PingSwarm, SecProtocol, Transport};
 
 fn build_builder<T, C>(
```
thomaseizinger, line 19: This function doesn't have much value because the type constraints are so insane. You are likely better off by just inlining things and accepting the duplication between the different invocations.

multidim-interop/rust/src/lib.rs
```diff
@@ -1,14 +1,44 @@
...
+
+/// Supported stream multiplexers by rust-libp2p.
+#[derive(Clone, Debug, EnumString)]
+#[strum(serialize_all = "kebab-case")]
+pub enum Muxer {
+    Mplex,
+    Yamux,
+    Quic,
+    Webrtc,
```
thomaseizinger, line 28: These aren't strictly muxers. I think we should leave the muxer variable empty when we don't explicitly set one. IMO, that will be cleaner implementation.

thomaseizinger (commented):
Some thoughts, otherwise LGTM.

multidim-interop/rust/src/lib.rs
jxs: yeah makes sense, thanks Thomas addressed :)

multidim-interop/rust/src/bin/testplan_0500.rs
jxs: I still feel it's more lines of code and repetition, but feel free to submit a PR addressing that Thomas

multidim-interop/rust/src/bin/testplan_0500.rs
jxs: it's now all on the testplans lib, which we can depend on the `rust-libp2p` repo and use from there wdyt?

multidim-interop/rust/src/lib.rs
jxs: created a helper function `from_env` on `lib.rs` which basically made me feel that we are re-writing `clap` so I'll look after if `clap` can actually make this more elegant.

multidim-interop/src/generator.ts
jxs: yeah you are right Marco, I reverted it int this PR so we can merge and address the `rust-libp2p` interop. But I also agree with @thomaseizinger on https://github.com/libp2p/test-plans/pull/100#discussion_r1071735120 if you agree we can submit a PR not using muxer nor transport for `quic` and `webrtc` cases

multidim-interop/src/generator.ts
jxs: Yeah agree Thomas, commented on https://github.com/libp2p/test-plans/pull/100#discussion_r1072194829

multidim-interop/rust/src/lib.rs
mxinden: Let's delay this :+1:

multidim-interop/src/generator.ts
MarcoPolo: I think this makes sense to me. We can make this change in a future PR so we do it across all impls

multidim-interop/src/generator.ts
```diff
@@ -79,14 +86,27 @@ export async function buildTestSpecs(versions: Array<Version>): Promise<Array<Co
...
+        .map((test): ComposeSpecification => buildSpec(containerImages, {
+            name: `${test.id1} x ${test.id2} (${test.transport})`,
+            dialerID: test.id1,
+            listenerID: test.id2,
+            transport: test.transport,
+            muxer: "quic",
+            security: "quic",
+        })))
+        .concat(webrtcQueryResults
+        .map((test): ComposeSpecification => buildSpec(containerImages, {
```
MarcoPolo, line 102: Can we indent this line? The map applies to `webrtcQueryResults` not the `concat(...)` expression like above.

MarcoPolo (approved):

multidim-interop/src/generator.ts
jxs: yup makes sense Marco, updated :)

mxinden:
@jxs two approvals, all checks are green. Please merge once this is ready from your end.

PR #102, opened 2023-01-13
Draft: support ping test for both NodeJS and Browser

GlenDC:
A previous (already merged) PR added support for the new wo-testground way of going for the ping test case for the JS Runtime. This was however only so for NodeJS.

This PR extends that support to Browser.

NOTE: for now this PR is broken as the Docker support is still to be fixed with the new dual support.
Once we agree on the currently taken approach for the dual support (NodeJS + Browser) I'll finish of the docker support,
as part of this PR.

Goal is to have the same test file for both NodeJS and Browser, and to be able to avoid having to bother for our JS tests to write anything non-test specific (e.g. setup browser).


GlenDC:
Replaced by https://github.com/libp2p/test-plans/pull/104, as it makes more sense to go for the aegir based approach. More sense as that is also the testing strategy chosen for all other js-libp2p codebase(s).

PR #104, opened 2023-01-14
Use aegir for testing js-libp2p

MarcoPolo:
This also adds browser support.

This doesn't actually let the browser dial the other node with websockets because the other node doesn't have a valid SSL cert. This can be fixed by passing in [this option](https://playwright.dev/docs/api/class-browser#browser-new-context-option-ignore-https-errors) when we create the page in `playwright-test` (this will require an upstream change).

This would work with webtransport.

Code needs a bit of clean up. The main files are .aegir.js which handles the setup/tear down and the redis proxy server, and ping.spec.ts which does the ping test.


multidim-interop/js/v0.41/test/ping.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+import { webSockets } from '@libp2p/websockets'
+import { noise } from '@chainsafe/libp2p-noise'
+import { mplex } from '@libp2p/mplex'
+import { yamux } from '@chainsafe/libp2p-yamux'
+import { multiaddr } from '@multiformats/multiaddr'
+
+describe('ping test', () => {
+  it('should ping', async () => {
+    let nodeFetchImport = "node-fetch"
+    let fetch = (typeof window === "undefined") ?
```
GlenDC, line 18: I believe stuff like https://www.npmjs.com/package/cross-fetch can handle this for you

multidim-interop/js/v0.41/test/ping.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    const MUXER = process.env.muxer
+    const IS_DIALER_STR = process.env.is_dialer
+    const IP = process.env.ip
+
+    const isDialer = IS_DIALER_STR === 'true'
+    const options: Libp2pOptions = {
+      start: true
+    }
+
+    switch (TRANSPORT) {
```
GlenDC, line 33: Cleaner I think might be to turn this into an if else and disable tcp case if browser. And the regular import is also cleaner I think with the package.json to just turn @libp2p/tcp to false (so an empty import), this keeps all imports at top.

multidim-interop/js/v0.41/test/ping.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+      (await import(nodeFetchImport)).default :
+      window.fetch
+
+    const TRANSPORT = process.env.transport
+    const SECURE_CHANNEL = process.env.security
+    const MUXER = process.env.muxer
+    const IS_DIALER_STR = process.env.is_dialer
+    const IP = process.env.ip
+
+    const isDialer = IS_DIALER_STR === 'true'
```
GlenDC, line 28: might force this to true for browser (I didn't do this,  but makes sense to me now)

multidim-interop/js/v0.41/test/ping.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    try {
+
+      if (isDialer) {
+        const otherMa = process.env.otherMa
+        if (otherMa === undefined) {
+          throw new Error("Failed to wait for listener")
+        }
+        console.log(`node ${node.peerId} pings: ${otherMa}`)
+        const rtt = await node.ping(multiaddr(otherMa))
```
GlenDC, line 82: this steps fails for me on my own approach due to what I discussed in the slack message. Something with the address being invalid for the transport in the browser. Does this work for you?

MarcoPolo: The browser filters out undialable addresses with: https://github.com/libp2p/js-libp2p-websockets/blob/master/src/filters.ts#L24. So you need a dns + wss multiaddr.

multidim-interop/js/v0.41/test/ping.spec.ts
MarcoPolo: Is this supported by all bundlers? Is that part of the package.json spec? https://docs.npmjs.com/cli/v9/configuring-npm/package-json#browser doesn't say anything about that. https://github.com/defunctzombie/package-browser-field-spec#ignore-a-module This does, but I'm not sure if that's well supported.

multidim-interop/js/v0.41/test/ping.spec.ts
MarcoPolo: webrtc can listen, no?

GlenDC: Can it? I was under the impression that this only will be possible once we support browser-to-browser, which we do not yet, do we?

multidim-interop/js/v0.41/test/ping.spec.ts
GlenDC: Fairly certain it should be yes. I always use webpack, def works for that. I imagine stuff like Vite as well. I mean it is really made for bundlers so it would be odd if a bundler does not support it. It's also pretty nice as i works for any dependency as well, no matter how deep in the chain.

multidim-interop/js/v0.41/test/ping.spec.ts
achingbrain: Node 18 has fetch available as a global, no need to add extra deps if you're running on that.

multidim-interop/js/v0.41/test/ping.spec.ts
achingbrain: The `browser` field isn't documented by npm as it's a webpack convention that was adopted by other bundlers - the defunctzombie spec linked to above is probably as good as we're ever going to get..

MarcoPolo: In that case, I feel like it’s better to do the dynamic import here which is part of the ES spec.

achingbrain: Using the regular `import` keyword and the `browser` field to mark the import `false` usually has fewer hidden surprises as the keyword is static code which is easily analysed by bundlers - dynamic imports sometimes cause problems as they are evaluated at runtime by which point the bundler isn't involved any more.

MarcoPolo:
This is ready for review. Adds one browser -> go (webtransport) test case. I'll add the websockets one in a future PR when I update `playwright-test`.

MarcoPolo:
FYI @achingbrain, for js-libp2p head testing:

1. We should have an implementation like this (or exactly this) in the js-libp2p repo.
2. Have js-libp2p call this workflow from its own workflow.

See https://github.com/libp2p/test-plans/pull/99 for more details. I'm happy to add this to js-libp2p since I already have context here. The biggest question for me is how to handle the various transport dependencies since they exist in separate repos. Does this mean that every release will also have to manually update the dependencies in the test?

multidim-interop/js/v0.41/.aegir.js
```diff
@@ -0,0 +1,90 @@
+import { spawn, exec } from "child_process";
+import { existsSync } from "fs";
+import { createClient } from 'redis'
+import http from "http"
+
+const isDialer = process.env.is_dialer === "true"
+const REDIS_ADDR = process.env.REDIS_ADDR || 'redis:6379'
+
+// Used to preinstall the browsers in the docker image
+const initialSetup = process.env.initial_setup === "true"
```
GlenDC, line 10: Seems like an anti pattern. Why does it enter the test at all for this? Not something that requires fixing in Aegir?

MarcoPolo: This is a playwright thing I believe. Maybe there's a way to preinstall chromium before running the test?

multidim-interop/js/v0.41/test/ping.spec.ts
GlenDC: Even if it wouldn't, dynamic imports like this are a bit dirty, no? Cleaner to see them all listed on top, of course one could argue that indirect dependencies are also not visible, but dunno if one should go there.

I have either way also no strong opinion, but it does make sense to me that bundlers do not take this into account, as that would mean to support it for real they would need to pretty much run the entire javascript code to know for sure what is required and what not.

multidim-interop/js/v0.41/test/ping.spec.ts
GlenDC: You might want to add that to the README, that and any other (aegir) info :) I know readme's aren't added a lot so far, but to me they do help and they can help any potential contributor or reader/user to understand how to use the different tests and related code.

multidim-interop/js/v0.41/.aegir.js
```diff
@@ -0,0 +1,90 @@
...
+            body += data;
+          });
+
+          req.on('end', function () {
+            resolve(JSON.parse(body))
+          });
+        })
+
+        try {
+          const redisRes = await redisClient.sendCommand(requestJSON)
```
GlenDC, line 39: Nice! A lot cleaner then my initial approach :D Well done, hat off.

multidim-interop/js/v0.41/package.json
```diff
@@ -1,30 +1,38 @@
...
+  "name": "multidim-interop",
+  "private": true,
+  "version": "1.0.0",
+  "description": "Multidimension Interop Test",
+  "type": "module",
+  "main": "index.js",
+  "author": "Glen De Cauwsemaecker <glen@littlebearlabs.io>",
+  "license": "MIT",
+  "engines": {
+    "node": ">=16"
```
GlenDC, line 11: might want to change this to the following to ensure fetch works

```suggestion
    "node": ">=18"
```

multidim-interop/js/v0.41/test/ping.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    switch (MUXER) {
+      case 'mplex':
+        options.streamMuxers = [mplex()]
+        break
+      case 'yamux':
+        options.streamMuxers = [yamux()]
+        break
+      case 'quic':
```
GlenDC, line 74: is this a mistake, or what is the point of this?

multidim-interop/js/v0.41/test/ping.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+        }
+        break
+      default:
+        throw new Error(`Unknown transport: ${TRANSPORT}`)
+    }
+
+    switch (SECURE_CHANNEL) {
+      case 'noise':
+        options.connectionEncryption = [noise()]
+        break
```
GlenDC, line 59: why not switch them around and make it fall through?

multidim-interop/js/v0.41/test/ping.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+        await redisProxy(["RPUSH", "listenerAddr", multiaddrs[0]])
+        try {
+          await redisProxy(["BLPOP", "dialerDone", "10"])
+        } catch {
+          throw new Error("Failed to wait for dialer to finish")
+        }
+      }
+    } finally {
+      // sleep for a second
+      await new Promise(resolve => setTimeout(resolve, 1000))
```
GlenDC, line 102: this smells? Why do we need this 1s? I say it smells as such code is rarely 100% reliable, and are often a reason for random CI failures.

multidim-interop/js/v0.41/.aegir.js
GlenDC: Probably there is. But even so, `.aegir` files are not a playwright thing, so given `aegir` is a layer in between the code and playwright, it might as well handle funny stuff like this already, as I do not see a point in users of a testing framework having to take this background detail into account.

Not really related to this PR though, but it still stands out

multidim-interop/js/v0.41/test/ping.spec.ts
GlenDC: Meant more like to a README which would be placed https://github.com/libp2p/test-plans/tree/master/multidim-interop/js/v0.41, usually I like to add readme's or other doc on how to use the code, as I know from experience that not everybody is always an expert in all the different protocols used, and than it can be helpful to see a quick example how to use/run the code. It's next to such a WS Example paragraph that one can then add quick instructions how to run this with a DNS + wss mutliaddr, in all the detail (and nothing more) that is required to be able to run this)).

If this is already well documented, you can next to an example of course just link to that doc with a 1 liner attached to it.

multidim-interop/js/v0.41/test/ping.spec.ts
MarcoPolo: oh, whoops. This was me testing something else.

GlenDC:
All in all I like this code. Good for you for pursuing the Aegir approach. If there's something I can do (in parallel or taking over (again)) in related to this PR or other interop testing stuff, do let me know here or via slack. I'm at your service @MarcoPolo .

multidim-interop/js/v0.41/test/ping.spec.ts
MarcoPolo: Right now we pass "muxer=quic security=quic" when running these tests. I'll remove this in a future PR: https://github.com/libp2p/test-plans/issues/106

MarcoPolo:
WebSockets won't be testable by the browser until we fix https://github.com/hugomrdias/playwright-test/issues/525

multidim-interop/js/v0.41/test/ping.spec.ts
MarcoPolo: We don't do this yet. Hopefully soon

multidim-interop/js/v0.41/test/ping.spec.ts
MarcoPolo: > WS Example paragraph that one can then add quick instructions how to run this with a DNS + wss mutliaddr, in all the detail (and nothing more) that is required to be able to run this)).

We can't run this with DNS + WSS yet because of the SSL error. But once we're able to do it I'll make sure to document and subtleties around it.

multidim-interop/js/v0.41/.aegir.js
MarcoPolo: I can do `playwright install` and get around this.

multidim-interop/js/v0.41/test/ping.spec.ts
MarcoPolo: Personal preference. Fall throughs can be subtle.

MarcoPolo:
I'll merge this tomorrow so I can continue with more browser testing.

PR #105, opened 2023-01-17
feat: multidim-interop: Setup badge workflow

MarcoPolo:
This will update a badge on our README.md to link to the latest interop test results. Fixes #55 .

PR #108, opened 2023-01-18
feat: update README badge automatically

galargh:
Resolves https://github.com/libp2p/test-plans/pull/105#issuecomment-1386296268

@MarcoPolo Here's a working version of the update badge workflow. Tested here - https://github.com/galorgh/test-plans/actions/runs/3947334594 

I think the problem you were facing was caused by the fact that workflows triggered by `workflow_run` events have to be checked in to the default branch.

### DIFF
--- a/.github/workflows/update-badge.yml
+++ b/.github/workflows/update-badge.yml
@@ -1,0 +1,79 @@
+name: Update Badge
+
+on:
+  workflow_run:
+    workflows:
+      - libp2p multidimensional interop test
+    types:
+      - completed
+    branches:
+      - master
+
+defaults:
+  run:
+    shell: bash
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+env:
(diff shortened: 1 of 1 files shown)


MarcoPolo:
>  workflows triggered by workflow_run events have to be checked in to the default branch.

Ah, I missed the note at https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

MarcoPolo:
Thank you!

.github/workflows/update-badge.yml
```diff
@@ -0,0 +1,79 @@
...
+on:
+  workflow_run:
+    workflows:
+      - libp2p multidimensional interop test
+    types:
+      - completed
+    branches:
+      - master
+
+defaults:
```
MarcoPolo, line 12: This isn't needed, is it? I thought this was already the default for ubuntu

MarcoPolo (approved):

Issue #52, opened 2022-10-08
add a test plan that measures the RTT of libp2p handshakes


marten-seemann:
The test plan will test that the libp2p handshake completes in the expected number of roundtrips. For example:
* the QUIC handshake should take 1 RTT
* the TCP handshake should take 4 RTTs (1. 3-way TCP handshake, 2. multistream negotiation of security protocol, 3. security handshake, 4. muxer negotiation)
* when using early muxer negotiation (https://github.com/libp2p/go-libp2p/issues/1789), the TCP handshake only takes 3 RTTs

This can easily extended to test that moving the security protocol into the multiaddr (https://github.com/libp2p/specs/pull/353) saves another RTT.

The test plan will need to build on the cross-version test (or maybe the cross-implementation test). This will allow us to assert that while new go-libp2p versions make use of all the performance improvements and handshaking takes fewer RTTs, we're still remain backwards-compatible.

In the long term, this test will probably replace the ping test plan, as it tests strictly more than the ping test.

marten-seemann:
This effort is currently blocked by https://github.com/testground/testground/issues/1488.

marten-seemann:
With the new wo-testground setup (we really need a better name for that…) this should be not too difficult to implement. I hacked a PoC in the [rtt-test branch](https://github.com/libp2p/test-plans/tree/rtt-test), although that code is by no means clean.

There are 2 main things we need to solve / standardize:
1. We need a launcher script (to be mounted into the container) that runs some setup before the node is started (executing the `tc` command to get `netem` to apply an RTT to the interface).
2. We need to specify the name of a results directory (`/results`?) and a JSON format how we export the measurements. I suggest exporting both the handshake latency and the latency determined via ping.

We can then check that the handshake latency matches what we expect. From my initial test, it looks like with a 100ms RTT, we end up with non-ambiguous results (e.g. a handshake that’s supposed to take 6 RTTs would take 615ms). We’ll need to check that this also holds on CI (reliably!). As long as the value is closer to N than to N+1, we should be on the safe side.

p-shahi:
deferring this in favor of the work in https://github.com/libp2p/test-plans/issues/63

PR #114, opened 2023-01-24
interop: Improve logging level of tests

jxs:
By reducing the `redis` container logs to warning, and changing the target of the `rust` ping test to stdout instead of stderr.

### DIFF
--- a/multidim-interop/rust/src/lib.rs
+++ b/multidim-interop/rust/src/lib.rs
@@ -1,7 +1,7 @@
 use std::{env, str::FromStr, time::Duration};
 
 use anyhow::{Context, Result};
-use env_logger::Env;
+use env_logger::{Env, Target};
 use log::info;
 use redis::{AsyncCommands, Client as Rclient};
 use strum::EnumString;
@@ -78,7 +78,9 @@ where
     let mut conn = client.get_async_connection().await?;
 
     info!("Running ping test: {}", swarm.local_peer_id());
-    env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
+    env_logger::Builder::from_env(Env::default().default_filter_or("info"))
+        .target(Target::Stdout)
+        .init();
 
     info!(
         "Test instance, listening for incoming connections on: {:?}.",
--- a/multidim-interop/rust/src/lib.rs
+++ b/multidim-interop/rust/src/lib.rs
@@ -1,7 +1,7 @@
 use std::{env, str::FromStr, time::Duration};
 
 use anyhow::{Context, Result};
-use env_logger::{Env, Target};
+use env_logger::Env;
 use log::info;
 use redis::{AsyncCommands, Client as Rclient};
 use strum::EnumString;
@@ -78,9 +78,7 @@ where
     let mut conn = client.get_async_connection().await?;
 
     info!("Running ping test: {}", swarm.local_peer_id());
-    env_logger::Builder::from_env(Env::default().default_filter_or("info"))
-        .target(Target::Stdout)
-        .init();
+    env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
 
     info!(
         "Test instance, listening for incoming connections on: {:?}.",
--- a/multidim-interop/src/generator.ts
+++ b/multidim-interop/src/generator.ts
@@ -139,7 +139,12 @@ function buildSpec(containerImages: { [key: string]: string }, { name, dialerID,
                     ip: "0.0.0.0",
                 }
             },
-            redis: { image: "redis/redis-stack", }
+            redis: {
+                image: "redis/redis-stack",
+                environment: {
+                    REDIS_ARGS: "--loglevel warning"
+                }
+            }
         }
     }
 }


MarcoPolo (approved):

multidim-interop/rust/src/lib.rs
```diff
@@ -78,7 +78,9 @@ where
     let mut conn = client.get_async_connection().await?;
 
     info!("Running ping test: {}", swarm.local_peer_id());
-    env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
+    env_logger::Builder::from_env(Env::default().default_filter_or("info"))
+        .target(Target::Stdout)
```
thomaseizinger, line 82: AFAIK, it is good unix behaviour to send logs to stderr. Why change to stdout?

jxs: So as they show on the current interop tests, currently interop tests only show `stdout` see [here](https://github.com/libp2p/test-plans/blob/master/multidim-interop/src/compose-runner.ts#L61).  
I read a couple of discussions like [this one](https://softwareengineering.stackexchange.com/questions/439462/log-levels-and-stdout-vs-stderr) as I also found it surprising that the default for `env_logger` was `stderr` but as I didn't see an industry wide statys quo, so I thought it would be easier to just change the Rust tests, since the Go ones already log to `stdout`.

MarcoPolo: at least the “ping successful” should be on stdout. The rest can be on stderr. https://pubs.opengroup.org/onlinepubs/9699919799/functions/stderr.html

thomaseizinger: If `ping successful` is part of the test contract, we should probably be printing that on `stdout` with `println`. Logs should be on `stderr` IMO.

I am sure you thought about this @MarcoPolo: It would be nice if we could rely on the exit code instead of scanning `stdout` for a string. Would it make sense to push a `listenerDone` to redis and thus synchronise the shutdown between the two?

thomaseizinger: This is untested but have you explored using `depends-on` to express the relationship between `dialer` and `listener`? If we tell `docker compose` that `dialer` depends on `listener` and have `listener` run an endless loop at the end, we may get a cleaner shutdown behaviour where only the dialer exits by itself.

MarcoPolo: > If ping successful is part of the test contract, we should probably be printing that on stdout with println. Logs should be on stderr IMO.

Agreed on principle.

> I am sure you thought about this @MarcoPolo: It would be nice if we could rely on the exit code instead of scanning stdout for a string. Would it make sense to push a listenerDone to redis and thus synchronise the shutdown between the two?

Yes. That's what happens. docker compose runs `up` with `--exit-code-from dialer` and uses that exit code as the result of the test. The unfortunate thing is that option implies `--abort-on-container-exit` which will stop all containers if any container exits. If the listener exits first, then we'll get a non-zero exit code from dialer because it was stopped by compose. Extra synchronization doesn't fix this because the dialer can still be slow in shutting down. This lead to some flakiness in the js-libp2p tests.

We may be able to start these in a detached mode, then wait for the exit of the dialer, then stop the compose instances. A bit more involved but doable.

---

The reason I want the `Ping Successful: <duration>` log output is because then these tests will just work with the round trip counter with no modifications. We would introduce latency in the network link, then parse the duration.

MarcoPolo: > have listener run an endless loop at the end, we may get a cleaner shutdown behaviour where only the dialer exits by itself.

That's also an alternative. Although at first glance, I'm not a fan of the wait endlessly for our process to be terminated.

thomaseizinger: > > have listener run an endless loop at the end, we may get a cleaner shutdown behaviour where only the dialer exits by itself.
> 
> That's also an alternative. Although at first glance, I'm not a fan of the wait endlessly for our process to be terminated.

Would likely be wise to have a timeout around that.

thomaseizinger: > We may be able to start these in a detached mode, then wait for the exit of the dialer, then stop the compose instances. A bit more involved but doable.

Ugh, I can definitely see the appeal in having docker compose manage all of that for us.

> The reason I want the `Ping Successful: <duration>` log output is because then these tests will just work with the round trip counter with no modifications. We would introduce latency in the network link, then parse the duration.

Can you expand on what you mean by "the round trip counter"?

I am not opposed to logging success and failure of the test and parse further results but I'd prefer if we can do that conditionally on the exit code. In other words: Have the exit code be the source of truth for test success / failure and then parse stdout/stderr afterwards if we wanted to get more information. We might want to consider emitting JSON then so we can more easily parse it from our test runner.

marten-seemann: > Can you expand on what you mean by "the round trip counter"?

I assume that would be https://github.com/libp2p/test-plans/issues/52, but maybe the cleaner solution is to output a file, since we probably want to export multiple values?

MarcoPolo: What multiple values?

thinking about it more I like the “leave listener around for a bit”. It removes a redis sync point and lets us use the exit code as source of truth.

MarcoPolo: It would be a little easier to parse JSON. But I don't have a preference here. Matching against output strings is only marginally more work. And it's a little nicer to see "Ping Successful: <duration>" then `{"pingDuration": <duration>}`. But this is purely subjective.

Happy to define the stdout schema to print a JSON result at the end.

marten-seemann: > What multiple values?

1. RTT, as measured by the ping protocol
2. Duration of the handshake

thomaseizinger: > What multiple values?
> 
> thinking about it more I like the “leave listener around for a bit”. It removes a redis sync point and lets us use the exit code as source of truth.

I opened an issue: https://github.com/libp2p/test-plans/issues/115

thomaseizinger: > And it's a little nicer to see "Ping Successful: " then `{"pingDuration": <duration>}`.

Depending on how we define the JSON structure, the test runner can parse it and print something more human-centric :)

mxinden: Following up on the concrete discussion for this pull request I suggest:

- Logging to stderr
- Print `ping successful` to stdout via `println!`.

mxinden (approved):
Comment on logging to stderr, otherwise this looks good to me.

Thanks for the patch for the redis log level. Will make my life a lot easier.

jxs:
Yeah makes sense to me also after reading the conversation, thanks for elaborating and researching @thomaseizinger and @MarcoPolo. Should we then also print stderr from the compose runner? As logs from the rust ping test will not be shown (can be done in another PR)

MarcoPolo:
Is it useful? We print stderr when there’s a failure. But only stdout on success.

thomaseizinger:
> Is it useful? We print stderr when there’s a failure. But only stdout on success.

That is the way to go IMO. Quiet by default, noisy on failure.

marten-seemann:
> > Is it useful? We print stderr when there’s a failure. But only stdout on success.
> 
> That is the way to go IMO. Quiet by default, noisy on failure.

Judging from the experience with the QUIC interop runner, having nodes output (very) detailed logs by default has proved very useful in debugging failures, especially when you're debugging a failure involving an implementation you're not familiar with.

MarcoPolo:
Nodes should always output diagnostic logs. Just on stderr. If there's a failure we'll show stderr. If everything looks good, I'm currently thinking that the node simply output a json string in stdout `{"pingRTT": <duration>}`. This is inline with the purpose of stderr: https://pubs.opengroup.org/onlinepubs/9699919799/functions/stderr.html.

thomaseizinger:
> > > Is it useful? We print stderr when there’s a failure. But only stdout on success.
> > 
> > That is the way to go IMO. Quiet by default, noisy on failure.
> 
> Judging from the experience with the QUIC interop runner, having nodes output (very) detailed logs by default has proved very useful in debugging failures, especially when you're debugging a failure involving an implementation you're not familiar with.

I am not sure whether you are agreeing or disagreeing with the quoted statement? :)

marten-seemann:
My general take on this:
* Logs are useful. Let's encourage implementation to dial up the log level. No particular preference for stdout vs. stderr.
* Not sure why we need the `ping successful` output. We get the success / fail bit from the exit code, no need to parse any output. Parsing stdout seems error prone.
* Any test more sophisticated than a simple ping will want to export more results. These results might be measurements, or even binary files (e.g. a file downloaded via protocol `X`). I expect us to introduce a convention that nodes put all their output into a predefined directory, and the runner to run some validation / evaluation steps based on that output.

thomaseizinger:
> Not sure why we need the `ping successful` output. We get the success / fail bit from the exit code, no need to parse any output. Parsing stdout seems error prone.

See @MarcoPolo's comment here: https://github.com/libp2p/test-plans/pull/114#discussion_r1086089409

> * Any test more sophisticated than a simple ping will want to export more results. These results might be measurements, or even binary files (e.g. a file downloaded via protocol `X`). I expect us to introduce a convention that nodes put all their output into a predefined directory, and the runner to run some validation / evaluation steps based on that output.

Without wanting to go too much OT: Another way of solving this would be to use stdout to output metadata and if a certain tests outputs a binary file, we can emit the path to it via stdout. Having said that, let's see what makes the most sense once we have a concrete requirement at hand.

MarcoPolo:
Cherry picked this commit, so closing this

PR #117, opened 2023-01-27
Add multidim-interop test spec

MarcoPolo:
This defines what the test should do.

Note that none of the tests are currently conforming to this. Go prints to stdout; all the tests have the dialer also `RPUSH`; and none of them emit the structured JSON result.

Once we merge this, I'll update the existing tests.

### DIFF
--- a/multidim-interop/README.md
+++ b/multidim-interop/README.md
@@ -11,15 +11,15 @@ its capabilities along with the id of its container image.
 The implementation is run in a container and is passed parameters via
 environment variables. The current parameters are:
 
-| Name         | Description                                                  | Is Optional                                                     |
-| ------------ | ------------------------------------------------------------ | --------------------------------------------------------------- |
-| REDIS_ADDR   | A different address to connect to redis (default redis:6379) | yes                                                             |
-| ip           | IP address to bind the listener to                           | no                                                              |
-| transport    | The transport to use                                         | no                                                              |
-| muxer        | The muxer to use                                             | no, except when transport is one of quic, quic-v1, webtransport |
-| security     | The security channel to use                                  | no, except when transport is one of quic, quic-v1, webtransport |
-| is_dialer    | Should you dial or listen                                    | no                                                              |
-| test_timeout | Control the timeout of test. In seconds                      | yes, default to 10 seconds                                      |
+| Name                 | Description                                                  | Is Optional                                                     |
+| -------------------- | ------------------------------------------------------------ | --------------------------------------------------------------- |
+| transport            | The transport to use                                         | no                                                              |
+| muxer                | The muxer to use                                             | no, except when transport is one of quic, quic-v1, webtransport |
+| security             | The security channel to use                                  | no, except when transport is one of quic, quic-v1, webtransport |
+| is_dialer            | Should you dial or listen                                    | no                                                              |
+| ip                   | IP address to bind the listener to                           | yes, default to "0.0.0.0"                                       |
+| redis_addr           | A different address to connect to redis (default redis:6379) | yes, default to the `redis` host on port 6379                   |
+| test_timeout_seconds | Control the timeout of test.                                 | yes, default to 180 seconds.                                    |
 
 The test should do two different things depending on if it's the dialer or
 listener.
@@ -32,14 +32,14 @@ string result should be emitted to `stdout`.
 1. Connect to the Redis instance.
 2. Create a libp2p node as defined by the environment variables.
 3. Get the listener's address via Redis' `BLPOP` using the `listenerAddr` key.
-4. Record the current time as `handshakeStart`.
+4. Record the current instant as `handshakeStartInstant`.
 5. Connect to the listener.
-6. Ping the listener, and record the round trip time as `pingRTT`
-8. Record the time since `handshakeStart`. This is `handshakePlusOneRTT`.
-9. Print to `stdout` the JSON formatted string: `{"handshakePlusOneRTT":
-   handshakePlusOneRTT, "pingRTT": pingRTT}`. Durations should be printed in
+6. Ping the listener, and record the round trip duration as `pingRTT`
+7. Record the duration since `handshakeStartInstant`. This is `handshakePlusOneRTT`.
+8. Print to `stdout` the JSON formatted string: `{"handshakePlusOneRTTMillis":
+   handshakePlusOneRTT, "pingRTTMilllis": pingRTT}`. Durations should be printed in
    milliseconds as a float.
-10. Exit with a code zero.
+9.  Exit with a code zero.
 
 On error, the dialer should return a non-zero exit code.
 
@@ -51,7 +51,7 @@ The listener should emit all diagnostic logs to `stderr`.
 2. Create a libp2p node as defined by the environment variables.
 3. Publish the listener's address via Redis' `RPUSH` using the `listenerAddr`
    key.
-4. Sleep for the duration of test_timeout. The test runner will kill this
+4. Sleep for the duration of `test_timeout_seconds`. The test runner will kill this
    process when the dialer finishes.
 5. If the timeout is hit, exit with a non-zero error code.
 
--- a/multidim-interop/README.md
+++ b/multidim-interop/README.md
@@ -1,0 +1,57 @@
+# Interoperability test
+
+This tests that different libp2p implementations can communicate with each other
+on each of their supported capabilites.
+
+Each version of libp2p is defined in `versions.ts`. There the version defines
+its capabilities along with the id of its container image.
+
+# Test spec
+
+The implementation is run in a container and is passed parameters via
+environment variables. The current parameters are:
+
+| Name         | Description                                                  | Is Optional                                                     |
+| ------------ | ------------------------------------------------------------ | --------------------------------------------------------------- |
+| REDIS_ADDR   | A different address to connect to redis (default redis:6379) | yes                                                             |
+| ip           | IP address to bind the listener to                           | no                                                              |
+| transport    | The transport to use                                         | no                                                              |
+| muxer        | The muxer to use                                             | no, except when transport is one of quic, quic-v1, webtransport |
+| security     | The security channel to use                                  | no, except when transport is one of quic, quic-v1, webtransport |
+| is_dialer    | Should you dial or listen                                    | no                                                              |
+| test_timeout | Control the timeout of test. In seconds                      | yes, default to 10 seconds                                      |
+
+The test should do two different things depending on if it's the dialer or
+listener.
+
+## Dialer
+
+The dialer should emit all diagnostic logs to `stderr`. Only the final JSON
+string result should be emitted to `stdout`.
+
+1. Connect to the Redis instance.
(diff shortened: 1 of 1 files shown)


MarcoPolo:
@marten-seemann I think this will give us what we need to measure handshake RTTs. Am I missing something?

multidim-interop/README.md
```diff
@@ -0,0 +1,57 @@
...
+string result should be emitted to `stdout`.
+
+1. Connect to the Redis instance.
+2. Create a libp2p node as defined by the environment variables.
+3. Get the listener's address via Redis' `BLPOP` using the `listenerAddr` key.
+4. Record the current time as `handshakeStart`.
+5. Connect to the listener.
+6. Ping the listener, and record the round trip time as `pingRTT`
+8. Record the time since `handshakeStart`. This is `handshakePlusOneRTT`.
+9. Print to `stdout` the JSON formatted string: `{"handshakePlusOneRTT":
```
marten-seemann, line 39: I assume this is a float in seconds?

multidim-interop/README.md
```diff
@@ -0,0 +1,57 @@
...
+   handshakePlusOneRTT, "pingRTT": pingRTT}`.
+10. Exit with a code zero.
+
+On error, the dialer should return a non zero exit code.
```
marten-seemann, line 43:
```suggestion
On error, the dialer should return a non-zero exit code.
```

multidim-interop/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+The listener should emit all diagnostic logs to `stderr`.
+
+1. Connect to the Redis instance.
+2. Create a libp2p node as defined by the environment variables.
+3. Publish the listener's address via Redis' `RPUSH` using the `listenerAddr`
+   key.
+4. Sleep for the duration of test_timeout. The test runner will kill this
+   process when the dialer finishes.
+5. If the timeout is hit, exit with a non-zero error code.
```
marten-seemann, line 55: Can’t the server just run indefinitely, until it’s killed by the test runner?

marten-seemann: I think this would allow us to get rid of the test timeout env. The timeout would be enforced by the runner, which would simplify the test implementations.

MarcoPolo: > I think this would allow us to get rid of the test timeout env. 

It wouldn't. You still want a timeout in the redis BLPOP command on the dialer side.

mxinden (approved):

multidim-interop/README.md
```diff
@@ -0,0 +1,58 @@
...
+its capabilities along with the id of its container image.
+
+# Test spec
+
+The implementation is run in a container and is passed parameters via
+environment variables. The current parameters are:
+
+| Name         | Description                                                  | Is Optional                                                     |
+| ------------ | ------------------------------------------------------------ | --------------------------------------------------------------- |
+| REDIS_ADDR   | A different address to connect to redis (default redis:6379) | yes                                                             |
```
jxs, line 16:
```suggestion
| REDIS_ADDR   | A different address to connect to redis (default redis:6379) | yes, when not provided implementations should default to the `redis` host on port 6379
|
```

jxs (approved):
LGTM, thanks for this Marco, I can update #114 when this gets merged.

multidim-interop/README.md
marten-seemann: > > I think this would allow us to get rid of the test timeout env.
> 
> It wouldn't. You still want a timeout in the redis BLPOP command on the dialer side.

That timeout doesn’t need to be configurable, does it? There’s no way the redis lookup will take significant time, unless the other host fails to boot.

multidim-interop/README.md
```diff
@@ -0,0 +1,58 @@
...
+| ip           | IP address to bind the listener to                           | no                                                              |
+| transport    | The transport to use                                         | no                                                              |
+| muxer        | The muxer to use                                             | no, except when transport is one of quic, quic-v1, webtransport |
+| security     | The security channel to use                                  | no, except when transport is one of quic, quic-v1, webtransport |
+| is_dialer    | Should you dial or listen                                    | no                                                              |
+| test_timeout | Control the timeout of test. In seconds                      | yes, default to 10 seconds                                      |
```
BigLep, line 22:
```suggestion
| test_timeout_seconds | Control the timeout of test.                     | yes, default to 10 seconds                                      |
```

Make self documenting?

multidim-interop/README.md
```diff
@@ -0,0 +1,58 @@
...
+string result should be emitted to `stdout`.
+
+1. Connect to the Redis instance.
+2. Create a libp2p node as defined by the environment variables.
+3. Get the listener's address via Redis' `BLPOP` using the `listenerAddr` key.
+4. Record the current time as `handshakeStart`.
+5. Connect to the listener.
+6. Ping the listener, and record the round trip time as `pingRTT`
+8. Record the time since `handshakeStart`. This is `handshakePlusOneRTT`.
+9. Print to `stdout` the JSON formatted string: `{"handshakePlusOneRTT":
```
BigLep, line 39:
```suggestion
9. Print to `stdout` the JSON formatted string: `{"handshakePlusOneRoundTripDurationMillis":
```

Going for clarify on this being a duration and the units.

multidim-interop/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Dialer
+
+The dialer should emit all diagnostic logs to `stderr`. Only the final JSON
+string result should be emitted to `stdout`.
+
+1. Connect to the Redis instance.
+2. Create a libp2p node as defined by the environment variables.
+3. Get the listener's address via Redis' `BLPOP` using the `listenerAddr` key.
+4. Record the current time as `handshakeStart`.
```
BigLep, line 35: I assume there is "time" terminology we can standardize on?  (I don't know the latest and greatest.  I assume there are normal conventions on this.  My default is to use the precise terms back from jodatime (https://www.joda.org/joda-time/ ) which I believe are in other languages/libraries.)

```suggestion
4. Record the current instant as `handshakeStartInstant`.
```

Just an idea for clarity.  Your call of course.

multidim-interop/README.md
```diff
@@ -0,0 +1,58 @@
...
+5. Connect to the listener.
+6. Ping the listener, and record the round trip time as `pingRTT`
```
BigLep, line 37:
```suggestion
6. Ping the listener, and record the duration of the round trip as `pingRoundTripDuration`
```

(idea for clarity to avoid the ambiguous term "time".)

multidim-interop/README.md
```diff
@@ -0,0 +1,58 @@
...
+8. Record the time since `handshakeStart`. This is `handshakePlusOneRTT`.
```
BigLep, line 38:
```suggestion
8. Record the duration since `handshakeStartInstance`. This is `handshakePlusOneRoundTripDuration`.
```

multidim-interop/README.md
```diff
@@ -0,0 +1,58 @@
...
+9. Print to `stdout` the JSON formatted string: `{"handshakePlusOneRTT":
+   handshakePlusOneRTT, "pingRTT": pingRTT}`. Durations should be printed in
```
BigLep, line 40:
```suggestion
   handshakePlusOneRoundTripDuration, "pingRoundTripDurationMillis": pingRoundTripDuration}`. Durations should be printed in
```

multidim-interop/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Listener
+
+The listener should emit all diagnostic logs to `stderr`.
+
+1. Connect to the Redis instance.
+2. Create a libp2p node as defined by the environment variables.
+3. Publish the listener's address via Redis' `RPUSH` using the `listenerAddr`
+   key.
+4. Sleep for the duration of test_timeout. The test runner will kill this
```
BigLep, line 54:
```suggestion
4. Sleep for the duration of `test_timeout_seconds`. The test runner will kill this
```

multidim-interop/README.md
```diff
@@ -0,0 +1,58 @@
...
+# Test spec
+
+The implementation is run in a container and is passed parameters via
+environment variables. The current parameters are:
+
+| Name         | Description                                                  | Is Optional                                                     |
+| ------------ | ------------------------------------------------------------ | --------------------------------------------------------------- |
+| REDIS_ADDR   | A different address to connect to redis (default redis:6379) | yes                                                             |
+| ip           | IP address to bind the listener to                           | no                                                              |
+| transport    | The transport to use                                         | no                                                              |
```
BigLep, line 18: Is it useful to define the valid values for "transport"?  (same goes for "muxer" and "security").

Do we need to define the behavior if it gets pass a "transport" that it doesn't have a mapping for?

BigLep (commented):
Thanks for documenting.  I'll leave to other to approve.

multidim-interop/README.md
MarcoPolo: This is for the `BLPOP` command. This one depends on the other host starting up, listening, and then broadcasting their address. Originally this was hardcoded as 10 seconds, because that seemed like a large enough time frame. But in CI with a NodeJS listener that startup can take more than 10 seconds. This is the increase the spurred this change.

Now we can say, let's just set some arbitrarily high timeout here (e.g. 1 hour) then let the test runner kill things after N seconds. But in my experience, it's been helpful to see where the tests fails with a timeout. Did the dialer fail to get the listener's address? Or did the dialer fail to connect and ping?

multidim-interop/README.md
MarcoPolo: RTT is a well know term. I'll keep it.

multidim-interop/README.md
MarcoPolo: Valid values of transport depend on what the implementation says it supports. e.g. if an implementation doesn't support TCP, it'll never be called with a `transport=tcp` param.

thomaseizinger:
One thought but perhaps it is obvious to others: I haven't used redis before but it seems to work really well for such synchronisation requirements. However, if our test runner is now specific to the test, why doesn't it perform the synchronisation by reading and writing to stdin/stdout? We would probably not use compose then but just start the docker containers directly via JavaScript.

Something like:

- Start all nodes
- Wait for listener to print address (in JSON) on stdout
- Tell dialer to dial address
- Wait for dialer to exit
- Kill listener

I am not sure it is overall simpler but it would remove the dependency on redis and compose. One downside is that running the binaries / containers yourself is a little more involved because you have to type into stdin now.

MarcoPolo:
The nice thing about using compose is the test runner doesn't have to have much logic about the test itself. You can run `npm run test -- --name-filter="<some test filter>"  --emit-only` and get the exact compose yaml that you can run manually to run the test. You can replace the image with another one and tweak it.

It's also much easier to debug and develop this with a redis instance in the background that deals with passing information around. There are Redis clients for many languages (even prolog!).

Finally, it's nice that the runner runs compose and that's it. Sure it's looking at stdout once, but that's easy. The annoying part is dealing with lifetimes of containers.

MarcoPolo:
Thanks for all the comments all! Merging this and creating an issue to update the tests.

PR #121, opened 2023-01-31
Update Go & JS tests to conform to the multidim interop test spec.

### LINKED ISSUE #112: Remove Rust ping binary in favor of published docker image
cc @mxinden @jxs 

Tracking the follow-up work from https://github.com/libp2p/rust-libp2p/pull/3331 here.

How many versions do we want to go back?

### LINKED ISSUE #113: BuildKit is flooding our caches
I think this came in with the recent interop tests. I've found a mention of "BuildKit" within the "setup docker builx" step: https://github.com/libp2p/rust-libp2p/actions/runs/3994130715/jobs/6851689643#step:8:146

I don't have a solution but the increased number of caches is a problem for us as we are already operating on the given limit of 10GB with our regular CI: https://github.com/libp2p/rust-libp2p/actions/caches

### LINKED ISSUE #115: Interop test runner: Use exit code of dialer as source of truth
Relevant discussion: https://github.com/libp2p/test-plans/pull/114#discussion_r1085989316

Basic idea:

- On success, the listener loops with a timeout (1min?) and waits to be terminated by docker compose
- On success, the dialer exits with 0
- On failure the dialer exits with 1
- On failure, the listener exits with 1

This should simplify the following code snippets because we can use the exit code of the dialer as the source of truth:

https://github.com/libp2p/test-plans/blob/5c19f0876ffddab67c84f244556e633ac5a85fee/multidim-interop/src/compose-runner.ts#L45-L66

### LINKED ISSUE #120: Update interop test implementations to conform to...
to https://github.com/libp2p/test-plans/blob/master/multidim-interop/README.md

Impls:
- [x] Rust https://github.com/libp2p/rust-libp2p/pull/3423
- [ ] Go https://github.com/libp2p/test-plans/pull/121
- [ ] JS( https://github.com/libp2p/test-plans/pull/121

MarcoPolo:
Addresses #120.

Closes #113, closes #120, closes #115, closes #112.

@jxs could you update the Rust test please?

After this PR the tests in the implementation repos will be broken until we update them as well. @jxs can you take updating the rust-libp2p repo? I'll update go-libp2p.

We could do something fancier where we maintain two versions of this test runner and update the implementation repos only once. I think it's okay to break CI for a little bit to make these changes simpler (The time can be minimized if we don't merge this until {rust,go}-libp2p are ready). If folks feel differently let me know.


jxs:
Hi Marco, yeah no worries I'll start on it Tomorrow :)

thomaseizinger:
If we pin the action to a hash in the repo, we should be able to update in steps, right?

thomaseizinger:
> If we pin the action to a hash in the repo, we should be able to update in steps, right?

We might need to also change the action to not just checkout latest `test-plans` master but the hash it was pinned to.

I have an idea that could work ..

thomaseizinger:
> > If we pin the action to a hash in the repo, we should be able to update in steps, right?
> 
> We might need to also change the action to not just checkout latest `test-plans` master but the hash it was pinned to.

I just saw that we already have `test-plans-ref`: https://github.com/libp2p/test-plans/blob/7fd08bfd806b14939c07b9b0dad8a74bb41ca137/.github/workflows/run-testplans.yml#L23

Specifying that + pinning the action to a hash should allow us a graceful upgrade without having to do things in lock step.

thomaseizinger:
Can we merge my composite action before we do this?

https://github.com/libp2p/test-plans/pull/123

That would allow us to migrate each repository individually without breaking any CI.

jxs:
there's also #114. If you prefer you can cherry-pick 07a7eebfad93ac76c6a60d51a2363a0e87c897c9 here Marco

multidim-interop/rust/v0.50/Makefile
```diff
@@ -0,0 +1,18 @@
+image_name := rust-v0.50
+commitSha = 53e477d4423106bb9f9e744a4b275dfb6213ca0a
```
MarcoPolo, line 2: @thomaseizinger / @jxs

This is how I imagine defining future released versions. You can use a published version instead. I wasn't able to figure out how to get that working locally for me.

Also a note, from testing this manually. The Rust listener I don't think is listening for a sigint. The listener doesn't exit when compose tries to stop it after the dialer finishes.

MarcoPolo: also this isn't for the released version 0.50, it's for master I believe. That needs to be fixed, but the test-to-spec update isn't backported yet. Easy to do, I just haven't done it yet.

thomaseizinger: So if I understand correctly, we download the archive and build the container from the source of the tag?

Or do you expect the test to be defined here and we just reference the downloaded source-code?

thomaseizinger: If the test is still co-located with the source-code, I don't really see the benefit of doing this. Can you lay it out please?

If the test is not co-located but defined in here, then we can just use the package manager to point the released version.

MarcoPolo: The test is colocated with the implementation (in rust-libp2p). 

The code here should be minimal, fetch the repo at the specified hash/tag and then build the container. 

The dockerfile here is simply because I couldn't use the one in the rust-libp2p repo since im on a Mac and haven't setup cross builds. 

This way you can run these tests locally without having access to the registry. 

CI will cache this as well so it shouldn't add much time relative to a registry either.

thomaseizinger: > This way you can run these tests locally without having access to the registry.

Is that the only requirement that makes the two solutions differ? Because you can also do that if you just build the container locally and give it the right tag, then docker won't try to pull it again.

MarcoPolo: Interesting. What's the right tag in this case?

thomaseizinger: > CI will cache this as well so it shouldn't add much time relative to a registry either.

At least with Rust, caching and docker are a tricky business. Need to create a separate layer for the dependencies and any change to the lockfile will invalidate that one again, adding a solid 5min or so to the build. The cache we have on CI on the other hand can handle lockfile changes and only rebuilds what is necessary. Hence the solution of copying the binary in.

thomaseizinger: > Interesting. What's the right tag in this case?

This one: https://github.com/libp2p/rust-libp2p/blob/53e477d4423106bb9f9e744a4b275dfb6213ca0a/.github/workflows/publish-interop-test-images.yml#L48

Expanded example: `ghcr.io/libp2p/rust-libp2p/ping-interop-test:v0.50.0`

Try the following locally:

- `docker run hello-world`
- Disconnect from the internet
- `docker run hello-world`

The first will pull the container but unless you specify `--pull=always` on the second one, it will default to `missing` and because you have it, run your local container.

thomaseizinger: Following from that, once you've run the test-suite once locally, it will work without a network connection.

MarcoPolo:
```
❯ docker run --rm -it ghcr.io/libp2p/rust-libp2p/ping-interop-test:v0.50.0                                                                                            
Unable to find image 'ghcr.io/libp2p/rust-libp2p/ping-interop-test:v0.50.0' locally
docker: Error response from daemon: Head "https://ghcr.io/v2/libp2p/rust-libp2p/ping-interop-test/manifests/v0.50.0": denied.
```

I assume I need to authenticate with the registry somehow?

There are two problems I have with just relying on the tag.
1. I can't checkout this repo and run the test. Now I need to setup a registry to get this working.
2. It makes changes to the test implementation harder. Do I have to have a GH action publish a new container for that change? I can't do what the action does [here](https://github.com/libp2p/rust-libp2p/blob/53e477d4423106bb9f9e744a4b275dfb6213ca0a/.github/workflows/publish-interop-test-images.yml#L48) because I would need to cross compile to linux to create the binary.

My requirements:
1. It should be reproducible from a given test-plans commit. If a commit passes, it should always pass in the future.
2. I should be able to run the tests locally, and be able to tweak tests locally.

I'm happy to cache things all day long. And if we could use the registry as a cache, I'm fine with doing that as well. That's what I thought you meant when you said "Because you can also do that if you just build the container locally and give it the right tag". But now I understand you meant that I could manually build it and tag it myself so it doesn't try to fetch from the registry. This breaks the reproducibility.

thomaseizinger: > ```
> ❯ docker run --rm -it ghcr.io/libp2p/rust-libp2p/ping-interop-test:v0.50.0                                                                                            
> Unable to find image 'ghcr.io/libp2p/rust-libp2p/ping-interop-test:v0.50.0' locally
> docker: Error response from daemon: Head "https://ghcr.io/v2/libp2p/rust-libp2p/ping-interop-test/manifests/v0.50.0": denied.
> ```
> 
> I assume I need to authenticate with the registry somehow?

Nothing is published yet as I wanted to await the final spec of the tests! :)
The branches are ready in rust-libp2p. Just need to click the publish button in the workflow.
 
> There are two problems I have with just relying on the tag.
> 1. I can't checkout this repo and run the test. Now I need to setup a registry to get this working.

Nothing needs to be set up once the images are published. The images will be publicly accessible.

> 2. It makes changes to the test implementation harder. Do I have to have a GH action publish a new container for that change? I can't do what the action does [here](https://github.com/libp2p/rust-libp2p/blob/53e477d4423106bb9f9e744a4b275dfb6213ca0a/.github/workflows/publish-interop-test-images.yml#L48) because I would need to cross compile to linux to create the binary.

The fact that we build the binary outside the container is a simplification (and performance optimisation) that us `rust-libp2p` maintainers accepted because we all run linux. If necessary, we can switch to building inside a container. It has at the end of the day nothing to do with pulling images from a registry or building them locally.

> My requirements:
> 1. It should be reproducible from a given test-plans commit. If a commit passes, it should always pass in the future.

We can (and should) pin the container hash in the `versions.ts` file. That will fetch the right container even if the tag gets overwritten. IMO this is a super strong commitment to reproducibility because we are running bit-for-bit the same binary on every test.

> 2. I should be able to run the tests locally, and be able to tweak tests locally.

Changing the test is tricky because it lives in another source-code repository anyway. For local dev, I'd suggest to build the container locally and add it via `--extra-version`. I don't think we should "cheat" the local setup by overwriting an existing test or container.

MarcoPolo: Let's chat sync. I think that'll be faster than all this back and forth :)

thomaseizinger: @MarcoPolo and I agreed in a sync-chat that using these caches is basically equivalent to a registry, just slightly leaner.

Building a docker container will reuse local / cached layers if they are present. Running an image will pull it unless it is present. If we have a central cache in S3, building and pulling from a registry are essentially equivalent, we are just downloading layers.

The benefit of building and caching over pulling is that we only need to define how to build a container in the implementation repository. Then, any commit (or tag for that matter) can be referenced within the test-plans repo here and used to run the tests.

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -9,9 +9,40 @@ inputs:
...
+    required: false
+    default: ""
+  s3-secret-access-key:
+    description: "S3 secret key id for the cache"
+    required: false
+    default: ""
+  aws-region:
+    description: "Which AWS region to use"
+    required: false
+    default: "us-east-1"
```
thomaseizinger, line 27: Can these be env variables so we don't need to pass them on explicitly?

.github/workflows/multidim-interop.yml
```diff
@@ -9,11 +9,15 @@ name: libp2p multidimensional interop test
...
 jobs:
   run-multidim-interop:
-    uses: "./.github/workflows/run-testplans.yml"
-    with:
-      dir: "multidim-interop"
-  run-multidim-interop-ng:
+    env:
+      TEST_PLAN_DIR: ${{ inputs.dir }}
+      AWS_BUCKET: ${{ vars.S3_BUCKET_NAME }}
+      AWS_REGION: us-east-1
```
thomaseizinger, line 15: Why are we adjusting this? Have all repos migrated to the new action? I'd like to delete this first then (i.e. have smaller, more focused PRs.)

multidim-interop/Makefile
```diff
@@ -1,13 +1,14 @@
 GO_SUBDIRS := $(wildcard go/*/.)
 JS_SUBDIRS := $(wildcard js/*/.)
+RUST_SUBDIRS := $(wildcard rust/*/.)
```
thomaseizinger, line 3: This will all go away, now that we are co-locating the test with the implementations, correct?

multidim-interop/dockerBuildWrapper.sh
```diff
@@ -1,12 +1,13 @@
 #!/usr/bin/env /bin/bash
 
-# If in CI use caching
-if [[ -n "${CI}" ]]; then
+# If in CI and we have a defined cache bucket, use caching
+if [[ -n "${CI}" ]] && [[ -n "${AWS_BUCKET}" ]]; then
```
thomaseizinger, line 4: Why CI? Why not also locally?

multidim-interop/src/generator.ts
```diff
@@ -111,8 +110,6 @@ export async function buildTestSpecs(versions: Array<Version>): Promise<Array<Co
                 dialerID: test.id1,
                 listenerID: test.id2,
                 transport: test.transport,
-                muxer: "quic",
-                security: "quic",
```
thomaseizinger, line 115: Can we do this in a separate PR?

multidim-interop/src/generator.ts
```diff
@@ -139,27 +134,35 @@ function buildSpec(containerImages: { [key: string]: string }, { name, dialerID,
...
                     is_dialer: true,
                     ip: "0.0.0.0",
+                    ...(!!muxer && { muxer }),
+                    ...(!!security && { security }),
                     ...extraEnv,
                 }
             },
             listener: {
+                // Add init process to be PID 1 to proxy signals. Rust doesn't
+                // handle SIGINT without this
```
thomaseizinger, line 146: We could do that. But as far as I know, running init as PID is considered a good practise so I feel a bit uneasy rust-libp2p being called out as the bad actor here :)

.github/actions/run-interop-ping-test/action.yml
MarcoPolo: I picked this since it's explicit this is an input to the test. Env variables aren't as explicit.

.github/workflows/multidim-interop.yml
MarcoPolo: go and rust have. This doesn't affect them anyways. This is only for this repo.

multidim-interop/Makefile
MarcoPolo: No, we still need to define the steps needed to ask the implementation how to build the container.

I'll mark this PR as draft and it'll make more sense then. Still iterating a bit here.

.github/actions/run-interop-ping-test/action.yml
thomaseizinger: I see. Perhaps explicit is better, 4 inputs just felt like a lot of clutter.

multidim-interop/dockerBuildWrapper.sh
MarcoPolo: CI is the ephemeral runners. Locally you'll already have access to your local cache.

I don't want to deal with getting folks access to this cache bucket right now.

multidim-interop/src/generator.ts
MarcoPolo: lol, it is the only one that *needs* this :P

.github/actions/run-interop-ping-test/action.yml
MarcoPolo: That's how much is needed for the s3 cache to work. It'll test without them of course.

multidim-interop/src/generator.ts
MarcoPolo: I thought about it, but there's a lot of overhead for doing all these things in separate PRs. I reorganized the commits so that they can be reviewed commit by commit.

MarcoPolo:
This is ready for a review. It can be reviewed commit by commit. That's easier for me than the overhead of dealing with 6 PRs.

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -9,9 +9,40 @@ inputs:
...
 runs:
   using: "composite"
   steps:
+    - run: |
+        echo "S3_AWS_ACCESS_KEY_ID=${{ inputs.s3-access-key-id }}" >> $GITHUB_ENV
+        echo "S3_AWS_SECRET_ACCESS_KEY=${{ inputs.s3-secret-access-key }}" >> $GITHUB_ENV
+        echo "AWS_BUCKET=${{ inputs.s3-cache-bucket }}" >> $GITHUB_ENV
+        echo "AWS_REGION=${{ inputs.aws-region }}" >> $GITHUB_ENV
```
thomaseizinger, line 35: Do we need all four in the env?

multidim-interop/Makefile
thomaseizinger: Would it make sense to move this logic into the runner and only reference the commit in `versions.ts`?

The logic is the same (download, unzip, run docker build) for all implementations, right? If we agree on a directory path where in the repo this sits, we could save on the Makefiles for each version.

thomaseizinger (approved):
LGTM!

Some non-blocking questions.

I guess we should prepare the commits in `rust-libp2p` properly (i.e. finish the backport) before we merge this?

.github/actions/run-interop-ping-test/action.yml
MarcoPolo: yes

MarcoPolo: We can remove the access info.

MarcoPolo:
> I guess we should prepare the commits in rust-libp2p properly (i.e. finish the backport) before we merge this?

I have the preference of merging this, and then a PR later that updates the commit with the backport. I don't expect that backport to break anything and it would be good to unblock other work here (like rolling out the s3 caching).

thomaseizinger:
> > I guess we should prepare the commits in rust-libp2p properly (i.e. finish the backport) before we merge this?
> 
> I have the preference of merging this, and then a PR later that updates the commit with the backport. I don't expect that backport to break anything and it would be good to unblock other work here (like rolling out the s3 caching).

Sounds good to me!

multidim-interop/Makefile
thomaseizinger: Sounds good, let's see how it pans out! You've dealt with this more than I did :)

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -9,9 +9,38 @@ inputs:
...
+      shell: bash
+
+    - name: Configure AWS credentials for S3 build cache
+      if: ${{ inputs.s3-access-key-id }} != "" && ${{ inputs.s3-secret-access-key }} != ""
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-access-key-id: ${{ inputs.s3-access-key-id }}
+        aws-secret-access-key: ${{ inputs.s3-secret-access-key}}
+        aws-region: ${{ env.AWS_REGION }}
```
mxinden, line 42: rust-libp2p executes this step on each pull request. Some pull requests originate from forks. Those pull requests don't have access to the above secrets. Thus this step fails.

See https://github.com/libp2p/rust-libp2p/actions/runs/4142906998/jobs/7164133938 on https://github.com/libp2p/rust-libp2p/pull/3344 as an example.

I don't yet have the full picture on the caching approach via an S3 bucket. Would mounting the bucket as read-only from forks be an option? Worst case, we don't mount the bucket on forks at all, thus building all images from scratch.

MarcoPolo: > Thus this step fails.

The original plan was that the `if` should have skipped this step. I'll check what that isn't happening.

> Would mounting the bucket as read-only from forks be an option? 

Unfortunately no. The s3 cache needs rw access. Maybe there's a hidden option I'm not seeing, but when I tried a read only key it failed.

My original plan was to have forks build from scratch.

MarcoPolo: I had the `if` statement wrong. #131 fixes it.

thomaseizinger: Did you try omitting the `--cache-to` option when you tried with a read-only key?

According to https://docs.docker.com/build/cache/backends/, those two options operate separately and I am surprised that `--cache-from` by itself would need write permissions.

If we conditionally define `--cache-to`, we should be set our bucket to be publicly readable and have forks use our cache.

Alternatively, I see that this also has a registry cache option which might be interesting to explore.

I really want our CI to always be fast, not just for our own PRs. (I also don't like wasting compute power.)

thomaseizinger: Happy to help with this but it would earliest be in ~5 days as I am off the grid for the next few.

MarcoPolo: > Did you try omitting the --cache-to option when you tried with a read-only key?

I hadn't, that's probably it!

Probably worth figuring out how to make this bucket public and then defaulting to that public bucket if no keys are passed. I'll open an issue.

Issue #111, opened 2023-01-19
NG `run-testplans` action


thomaseizinger:
Currently, this is a workflow but it could just be a composite action which would remove the artifact up and downloading.

Relevant discussion: https://github.com/libp2p/rust-libp2p/pull/3331#discussion_r1073550520

Suggestions on new API design for the action/test-runner: https://github.com/libp2p/rust-libp2p/pull/3331#discussion_r1073556206

In particular:

- Change API to accept a series of JSON files
- Each file describes capabilities of one test binary (i.e. one JSON object, no arrays)

Happy to implement these changes once we agree on them :)

thomaseizinger:
> > Change API to accept a series of JSON files
> 
> Yup. like `npm test -- --extra-version=./foo.json --extra-version=./bar.json` works for me.

I was actually thinking something like:

```
npm test -- ./go-libp2p-v0.22.json ./go-libp2p-head.json ./rust-libp2p-head.json
```

In other words, arguments the same way `mv` takes a list of files to move.

The name of the files doesn't matter. I'd also consider getting rid of `versions.ts` and require the JSON files imported there to be passed on the commandline too.

> > Each file describes capabilities of one test binary (i.e. one JSON object, no arrays)
> 
> sgtm!
> 
> Let me know if you do end up picking this up. I'll do the same if I start on this today :)

Will do!

MarcoPolo:
> I'd also consider getting rid of versions.ts and require the JSON files imported there to be passed on the commandline too.

Getting type checking is very nice. I don't know if I want to give that up. I think of "version.ts" as the source of truth for what supported versions are capable of.

thomaseizinger:
> > I'd also consider getting rid of versions.ts and require the JSON files imported there to be passed on the commandline too.
> 
> Getting type checking is very nice. I don't know if I want to give that up. I think of "version.ts" as the source of truth for what supported versions are capable of.

I just figured that if we pass the "extra" versions on the command-line we might as well pass all JSON files there to simplify the overall design. It seemed like a natural extension but you are right that we'd be losing type-checks on those files. Why are they declared in JSON files to begin with and not just within the array that is defined in `versions.ts`?

MarcoPolo:
> Why are they declared in JSON files to begin with and not just within the array that is defined in versions.ts?

They are defined in the array? https://github.com/libp2p/test-plans/blob/master/multidim-interop/versions.ts#L18

[...]

PR #123, opened 2023-02-01
Introduce composite action for running ping interop tests

thomaseizinger:
This PR implements some of the ideas discussed in #111. It is intended to be fully backwards-compatible and can thus be rolled out incrementally across our repositories.

We do a few things here:

1. GitHub actions needs to already checkout the `test-plans` repository to access the action's source code. We can reuse this checkout to access the rest of the repository. Thus, if we reference this action via a commit hash, downstream repositories don't suddenly break on changes to the test-runner or other things in here.
2. Compared to the previous workflow, we don't cache the NPM dependencies. This is a technical limitation originating from (1) as GitHub refuses to cache things outside of $GITHUB_WORKSPACE. This doesn't really matter though as our runner has very few dependencies which are installed in no time.
3. Because this is now a composite action, we don't need to up/download any artifacts from the source repository. Instead we simply embed this action in a bigger workflow: https://github.com/libp2p/rust-libp2p/blob/test-composite-test-plans-action/.github/workflows/interop-test.yml.
4. `extra-versions` should be a space-separated list of JSON files that each contain a JSON **object** which describes a version to include in the matrix. See the changes to the JSON file in the `rust-libp2p` PR: https://github.com/libp2p/rust-libp2p/pull/3414/files
5. The support for `extra-version` in the test runner is backwards-compatible.
6. When building local docker images, repositories need to make sure that `containerImageID` exists locally. As you can see from the `rust-libp2p` PR, all that is needed is to build and tag the image locally. Because we are within the same workflow, the state of the local docker daemon is preserved.

### DIFF
--- a/multidim-interop/testplans.ts
+++ b/multidim-interop/testplans.ts
@@ -50,9 +50,7 @@ import path from "path";
         }
     }
 
-    const extraVersionPaths = argv.extraVersion as string[] || []
-
-    for (let versionPath of extraVersionPaths.filter(p => p !== "")) {
+    for (let versionPath of argv.extraVersion.filter(p => p !== "")) {
         const contents = await fs.readFile(versionPath);
         extraVersions.push(JSON.parse(contents.toString()))
     }


.github/actions/run-interop-ping-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+    default: ""
+  extra-versions:
+    description: "Space-separated paths to JSON files describing additional images"
+    required: false
+    default: ""
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        WORK_DIR=$(realpath "${{ github.action_path }}/../../../multidim-interop")
```
MarcoPolo, line 16: Does this depend on where the caller checks this out? Or how did you know how many parent dirs to go up to?

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+        echo "WORK_DIR=$WORK_DIR" >> $GITHUB_OUTPUT
+      shell: bash
+      id: find-workdir
+
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 17
```
MarcoPolo, line 23: Might as well use 18 since it's LTS

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+      run: npm ci
+      shell: bash
+
+    - name: Build images
+      working-directory: ${{ steps.find-workdir.outputs.WORK_DIR }}
+      run: make
+      shell: bash
+
+    - name: Run the test
+#      timeout-minutes: 30
```
MarcoPolo, line 43: Uncomment or delete?

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+      working-directory: ${{ steps.find-workdir.outputs.WORK_DIR }}
+      run: WORKER_COUNT=2 npm run test -- --extra-version=${{ inputs.extra-versions }} --name-filter=${{ inputs.test-filter }}
```
MarcoPolo, line 45: Does multiple extra versions work? if I pass in "a.json b.json" does this get interpolated as:
```
 WORKER_COUNT=2 npm run test -- --extra-version=a.json b.json --name-filter=${{ inputs.test-filter }}
 ```
 or 
 ```
  WORKER_COUNT=2 npm run test -- --extra-version="a.json b.json" --name-filter=${{ inputs.test-filter }}
 ```

File: .github/actions/run-interop-ping-test/action.yml (hunk 1 above)
Comment (thomaseizinger) [reply .github/actions/run-interop-ping-test/action.yml:16]: This depends on where this file is within this repository. I am basically walking up from the `.github` directory here.

File: .github/actions/run-interop-ping-test/action.yml (hunk 3 above)
Comment (thomaseizinger) [reply .github/actions/run-interop-ping-test/action.yml:43]: I've left it for now to remind myself that it is a missing feature now. Is deleting fine?

File: .github/actions/run-interop-ping-test/action.yml (hunk 4 above)
Comment (thomaseizinger) [reply .github/actions/run-interop-ping-test/action.yml:45]: According to the yargs docs yes. I'll do some local testing to confirm.

File: .github/actions/run-interop-ping-test/action.yml (hunk 1 above)
Comment (MarcoPolo) [reply .github/actions/run-interop-ping-test/action.yml:16]: Got it. That makes sense. Could we leave a comment here saying that?

Comment (thomaseizinger) [reply .github/actions/run-interop-ping-test/action.yml:16]: Will do! Technically, we could even hardcode this path but then we'd depend on GitHub actions runner internals.

File: .github/actions/run-interop-ping-test/action.yml (hunk 4 above)
Comment (MarcoPolo) [reply .github/actions/run-interop-ping-test/action.yml:45]: Tested this:
```
❯ npm run test -- --extra-version=a.json b.json   

> Multidimensional libp2p interop test@0.0.1 test
> ts-node testplans.ts --extra-version=a.json b.json

Extra version paths:  [ 'a.json', 'b.json' ]
```

I believe the action will run it without the `"`. so this should work.

File: .github/actions/run-interop-ping-test/action.yml [hunk 5]
```diff
@@ -0,0 +1,78 @@
...
+      shell: bash
+
+    - name: Print the results
+      working-directory: ${{ steps.find-workdir.outputs.WORK_DIR }}
+      run: cat results.csv
+      shell: bash
```
MarcoPolo, line 51: We don't need to specify bash everywhere, do we? This is default for linux/mac. I guess we'd need to specify this if this is run in windows (but I bet other things would break).

Can this be specified once at the top? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrun

.github/workflows/multidim-interop.yml
```diff
@@ -12,3 +12,8 @@ jobs:
     uses: "./.github/workflows/run-testplans.yml"
     with:
       dir: "multidim-interop"
+  run-multidim-interop-ng:
```
MarcoPolo, line 15: Should we just replace the above?

.github/actions/run-interop-ping-test/action.yml
MarcoPolo: I added the comment

.github/actions/run-interop-ping-test/action.yml
MarcoPolo: Their example sets shell every time as well: https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

MarcoPolo: This is required: https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsshell.

Nvm

.github/workflows/multidim-interop.yml
MarcoPolo: I'm guessing the intent was to make sure the existing flow didn't break?

MarcoPolo (approved):
This looks good.

After we merge this, I'm guessing the plan is to change {rust,go}-libp2p to use this composite action at the current commit. Then we can rollout the other changes?

thomaseizinger:
> This looks good.
> 
> After we merge this, I'm guessing the plan is to change {rust,go}-libp2p to use this composite action at the current commit. Then we can rollout the other changes?

Yep, that would be the idea!

.github/workflows/multidim-interop.yml
thomaseizinger: Yeah, it is always easy enough to just remove the above later.

.github/actions/run-interop-ping-test/action.yml
thomaseizinger: > we could even hardcode this path

This is wrong because it wouldn't work with forks of this action. See an example here: https://github.com/libp2p/rust-libp2p/actions/runs/4060621646/jobs/6989891784#step:6:10

We have to use the `github.action_path` variable to account for this dynamic element.

thomaseizinger:
> After we merge this, I'm guessing the plan is to change {rust,go}-libp2p to use this composite action at the current commit.

I opened a PR for `go-libp2p` here: https://github.com/libp2p/go-libp2p/pull/2039

PR #125, opened 2023-02-01
Cache builds with s3 buckets

### LINKED ISSUE #113: BuildKit is flooding our caches

MarcoPolo:
Closes #113

### DIFF
--- a/.github/workflows/multidim-interop.yml
+++ b/.github/workflows/multidim-interop.yml
@@ -12,6 +12,7 @@ jobs:
     uses: "./.github/workflows/run-testplans.yml"
     with:
       dir: "multidim-interop"
+    secrets: inherit
   run-multidim-interop-ng:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/run-testplans.yml
+++ b/.github/workflows/run-testplans.yml
@@ -31,6 +31,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       TEST_PLAN_DIR: ${{ inputs.dir }}
+      AWS_BUCKET: ${{ vars.S3_BUCKET_NAME }}
+      AWS_REGION: us-east-1
     defaults:
       run:
         shell: bash
@@ -41,6 +43,12 @@ jobs:
           path: test-plans
           repository: "libp2p/test-plans"
           ref: ${{ inputs.test-plans_ref }}
+      - name: Configure AWS credentials for S3 build cache
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ vars.S3_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.S3_AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ env.AWS_REGION }}
       # Download input data
       - uses: actions/download-artifact@v3
         if: ${{ inputs.extra-versions != '' }}
--- a/multidim-interop/dockerBuildWrapper.sh
+++ b/multidim-interop/dockerBuildWrapper.sh
@@ -1,12 +1,15 @@
 #!/usr/bin/env /bin/bash
 
+# TODO replace bucket name here
+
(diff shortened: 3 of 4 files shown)


marten-seemann (commented):
Why does this PR change the test implementation? Accidentally committed part of the handshake RTT test?

MarcoPolo:
Targeting the wrong base. I updated.

marten-seemann:
Is this cache ever garbage collected?

MarcoPolo:
According to the terraform this is set to [90 days](https://github.com/pl-strflt/tf-aws-bootstrap/blob/main/libp2p.tf#L24). That's probably overkill, but this will eventually be GCd.

thomaseizinger:
This should target the new action now instead of the old workflow.

What happens if the secrets aren't set? Will it fail or just not use the cache?

MarcoPolo:
> This should target the new action now instead of the old workflow.

Yup! I'll rebase the other PR and update this.

> What happens if the secrets aren't set? Will it fail or just not use the cache?

Fails at the start. I'll change this to not use the cache instead. I think it's important to allow other people to be able to run these tests on their forks without having to deal with setting up an S3 cache.

thomaseizinger:
> > What happens if the secrets aren't set? Will it fail or just not use the cache?
> 
> Fails at the start. I'll change this to not use the cache instead. I think it's important to allow other people to be able to run these tests on their forks without having to deal with setting up an S3 cache.

Agreed. Are we going to use org-wide secrets for this? I think it would make sense to share the cache between our implementations.

thomaseizinger:
Is it possible to set up this cache such that we don't need credentials to read from it? It would be nice if I people would benefit from these when they run the tests locally.

MarcoPolo:
> Is it possible to set up this cache such that we don't need credentials to read from it? It would be nice if I people would benefit from these when they run the tests locally.

I think so? But I'm not sure how common that will be. Or if it'll open us up to having to consider mitigating attacks on this (unbounded reads can be expensive).

MarcoPolo:
I'm going to merge this to #121 and continue from there.

thomaseizinger:
> > Is it possible to set up this cache such that we don't need credentials to read from it? It would be nice if I people would benefit from these when they run the tests locally.
> 
> I think so? But I'm not sure how common that will be. Or if it'll open us up to having to consider mitigating attacks on this (unbounded reads can be expensive).

I'd like to avoid having to set up S3 creds on my local machine but still want to benefit from the cache.

MarcoPolo:
> I'd like to avoid having to set up S3 creds on my local machine but still want to benefit from the cache.

just so I understand, the ask is that you’d like to avoid the initial 15-30 minute or so one time cost in building the containers?

thomaseizinger:
> > I'd like to avoid having to set up S3 creds on my local machine but still want to benefit from the cache.
> 
> just so I understand, the ask is that you’d like to avoid the initial 15-30 minute or so one time cost in building the containers?

Not sure how niche this is but I have a "vacuum" script that cleans out all caches (package manager, docker, build dirs, etc) on my harddrive. That is easily dozens of GB. I run that routinely (whenever I need space) so this wouldn't just be "initial".

Having said that, I am okay with not optimizing for this. I hope I have to seldomly run these locally! :)

marten-seemann:
> Not sure how niche this is but I have a "vacuum" script that cleans out all caches (package manager, docker, build dirs, etc) on my harddrive. That is easily dozens of GB. I run that routinely (whenever I need space) so this wouldn't just be "initial" but probably every time I need them.

Sounds pretty niche to me. @thomaseizinger, this also sounds like you need a bigger hard drive ;)

PR #127, opened 2023-02-07
multidim-interop: change output units to seconds

marten-seemann:
Let's use SI units wherever possible.

### DIFF
--- a/multidim-interop/README.md
+++ b/multidim-interop/README.md
@@ -36,10 +36,10 @@ string result should be emitted to `stdout`.
 5. Connect to the listener.
 6. Ping the listener, and record the round trip duration as `pingRTT`
 7. Record the duration since `handshakeStartInstant`. This is `handshakePlusOneRTT`.
-8. Print to `stdout` the JSON formatted string: `{"handshakePlusOneRTTMillis":
-   handshakePlusOneRTT, "pingRTTMilllis": pingRTT}`. Durations should be printed in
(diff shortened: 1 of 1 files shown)


MarcoPolo:
Why seconds? This is subjective, but milliseconds seems to be natural range for this as results will be between between 0.1 and ~100. That reads easier than 0.0001 to 0.100.

marten-seemann:
Any choice will be arbitrary, so I suggest to use SI units. Same reason we measure distances in meters and not in feet 🤓

As a point of reference, Prometheus also has the same convention: https://prometheus.io/docs/practices/naming/

MarcoPolo:
I don't understand. This is is an SI unit with an SI prefix. I really don't want to change this just to avoid an SI prefix.

thomaseizinger (commented):
I am ambivalent. Happy to move forward with this if you own the change through the various repositories @marten-seemann!

Note regarding prometheus: It does recommend a `_seconds` postfix so it would be "pingRTTSeconds".

MarcoPolo:
Overall I'm leaning towards closing this. This seems to be minor and don't make the test any better. The cost is extra time updating and backporting these changes. Unless you feel strongly and want to push these changes @marten-seemann, I suggest closing this.

MarcoPolo:
Closing this as it's been a week. If you're committed to this, feel free to reopen.

PR #128, opened 2023-02-07
chore: update readme and update references to Testground

p-shahi:
Not updating the roadmap yet.


README.md
```diff
@@ -1,11 +1,19 @@
...
 
 [![Made by Protocol Labs](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](http://protocol.ai)
 ![Go version](https://img.shields.io/badge/go-%3E%3D1.14.0-blue.svg?style=flat-square)
 
-This repository contains Testground test plans for libp2p components.
+This repository contains interoperability and end to end tests for libp2p modules across different implementations and versions.
+
+## Specs
+
+Please see the interoperability test specification here: [Interoperability Test Specs](https://github.com/libp2p/test-plans/blob/master/multidim-interop/README.md)
```
thomaseizinger, line 12: As far as I understand, this is not _the_ specification but just _one_ specification. As we add more tests, we will have to write more of these specs for e.g. AutoNAT, relay, etc.

I think that should be captured in the wording /  structure here.

thomaseizinger (approved):
Thank you for this.

README.md
p-shahi: Good point, I've addressed

PR #129, opened 2023-02-07
[DISCUSS] chore: tidy up the repo

p-shahi:
I propose we find a new home for the tests in `dht/` and `pubsub/` as this repo is mainly for e2e tests of libp2p modules and implementations.
Furthermore, they rely on Testground which we don't use anymore.


thomaseizinger:
I think the `ping` directory can go as well. We only run the `multidim-interop` tests now. Those are actually _test_ specific as far as I understand so perhaps that should be renamed to `ping`?

thomaseizinger:
The README is also a bit dated I tink.

thomaseizinger:
> The README is also a bit dated I think

Sorry, only just saw https://github.com/libp2p/test-plans/pull/128.

p-shahi:
> I think the ping directory can go as well. We only run the multidim-interop tests now.
Thanks I wasn't sure, removed it now.

> Those are actually test specific as far as I understand so perhaps that should be renamed to ping?

Should the folder be renamed or files?

For files, I prefer having descriptive names to make it clear what the test is doing from a glance:
* testplan_0500.rs -> ping_0500.rs
* main.go -> ping_0220.go
* ping.spec.ts -> ping0411.spec.ts

But maybe we can use this opportunity to move these tests under a ping/ directory too

i.e. move `multidim-interop/rust`, `multidim-interop/go`, `multidim-interop/js` to `multidim-interop/ping/rust/` `multidim-interop/ping/go`, `multidim-interop/ping/js`

thomaseizinger:
> > Those are actually test specific as far as I understand so perhaps that should be renamed to ping?
> 
> Should the folder be renamed or files?

As far as I understood @MarcoPolo , "multidim-interop" is **not** a generic test runner.

As we add more tests like AutoNAT, we will create new runners.

The reason is that asserting the correct result of the test is test specific: number of RTTs, latency, etc

Perhaps we will carve out some reusable bits at some point, or maybe even a generic runner but until then, the entire directory is "just one test".

To answer your question: Yes I think the directory should be renamed!

thomaseizinger:
> > Those are actually test specific as far as I understand so perhaps that should be renamed to ping?
> 
> Should the folder be renamed or files?

As far as I understood @MarcoPolo , "multidim-interop" is **not** a generic test runner.

As we add more tests like AutoNAT, we will create new runners.

The reason is that asserting the correct result of the test is test specific: number of RTTs, latency, etc

Perhaps we will carve out some reusable bits at some point, or maybe even a generic runner but until then, the entire directory is "just one test".

To answer your question: Yes I think the directory should be renamed!

MarcoPolo:
Looks good. Thanks for doing this, it's been on my list so glad to see it done!

MarcoPolo (approved):

PR #131, opened 2023-02-10
Test workflow without cache keys

MarcoPolo:
Might make sense to have this as a permanent workflow so that we make sure this doesn't break.

Draft for now as I fiddle with CI


MarcoPolo:
Leaning towards not including the no-cache workflow. If this fails without cache we'll know from the failed CI run, so skipping it to avoid the 20+min CI runs.

MarcoPolo:
Changed my mind. We can skip the actual tests, and the check to make sure the images build without cache should take roughly the same amount of time. And this only happens to changes in this repo. If it's annoying we can remove it later, but for now it makes me feel a little better to know we are testing this code path.

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -34,7 +34,7 @@ runs:
       shell: bash
 
     - name: Configure AWS credentials for S3 build cache
-      if: ${{ inputs.s3-access-key-id }} != "" && ${{ inputs.s3-secret-access-key }} != ""
+      if: inputs.s3-access-key-id != '' && inputs.s3-secret-access-key != ''
```
thomaseizinger, line 37: GitHub actions syntax is so subtle it is really annoying. Maybe I should pick up my idea of writing a linter for this again ...

My first guess at what is wrong here would have been:


```suggestion
      if: ${{ inputs.s3-access-key-id != '' }} && ${{ inputs.s3-secret-access-key != '' }}
```

.github/workflows/no-cache-multidim-interop.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  run-multidim-interop:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      # Purposely not using cache to replicate how forks will behave.
+      - uses: ./.github/actions/run-interop-ping-test
+        with:
+          # It's okay to not run the tests, we only care to check if the tests build without cache.
+          test-filter: "no test matches this, skip all"
```
thomaseizinger, line 19: Clever!

.github/workflows/no-cache-multidim-interop.yml
```diff
@@ -0,0 +1,19 @@
+on:
+  workflow_dispatch:
```
thomaseizinger, line 2: Do we need this?


```suggestion
```

.github/workflows/no-cache-multidim-interop.yml
```diff
@@ -0,0 +1,19 @@
...
+  pull_request:
+  push:
+    branches:
+      - "master"
```
thomaseizinger, line 6: We share the same triggers with the other workflow right? Why not have it in the other file as a separate job?

multidim-interop/dockerBuildWrapper.sh
```diff
@@ -9,6 +9,8 @@ if [[ -n "${CI}" ]] && [[ -n "${AWS_BUCKET}" ]]; then
     --cache-from type=s3,mode=max,bucket=$AWS_BUCKET,region=$AWS_REGION,prefix=buildCache,name=$IMAGE_NAME \
     "$@"
 else
-	docker build -t $IMAGE_NAME "$@"
+    docker buildx build \
+    --load \
+	-t $IMAGE_NAME "$@"
```
thomaseizinger, line 14: Would it make sense to do something like:


```suggestion
    docker buildx build \
    --load \
    $CACHING_OPTIONS \
	-t $IMAGE_NAME "$@"
```

Where `$CACHING_OPTIONS` is conditionally defined instead of the entire command?

thomaseizinger (approved):
Thanks! Some nits, nothing blocking :)

.github/workflows/no-cache-multidim-interop.yml
MarcoPolo: I wanted it to run in parallel and isolated. Is that the case with a separate job?

.github/workflows/no-cache-multidim-interop.yml
MarcoPolo, line 19:
```suggestion
          test-filter: '"no test matches this, skip all"'
```

.github/workflows/no-cache-multidim-interop.yml
thomaseizinger: Yes jobs run in parallel in a workflow unless you declare a dependency with `needs:`

PR #136, opened 2023-02-13
fix(aws): Only set bucket var when creds are available

mxinden:
The `dockerBuildWrapper` script enables caching based on the `AWS_BUCKET` environment variable. When calling the `libp2p ping interop test` with a bucket name but no AWS credentials, this leads to the script enabling caching even though the workflow does not have access to AWS.

With this commit the environment variable `AWS_BUCKET` is only set when credentials are available and thus the `dockerBuildWrapper` script does not enable caching then workflow does nothave access to AWS.

Not deeply familiar with the setup. Let me know in case there are cleaner solutions.

See also https://github.com/libp2p/rust-libp2p/pull/3462 and the corresponding failure on its first commit.

### DIFF
--- a/.github/actions/run-interop-ping-test/action.yml
+++ b/.github/actions/run-interop-ping-test/action.yml
@@ -28,7 +28,8 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - run: |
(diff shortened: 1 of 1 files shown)


MarcoPolo (approved):

mxinden:
:pray: thanks for the quick help @MarcoPolo!

Issue #132, opened 2023-02-10
Make build cache bucket public


MarcoPolo:
@galargh could we make the build cache publicly readable? This would allow forks to still benefit from the cache.

MarcoPolo:
Context here: https://github.com/libp2p/test-plans/pull/121#discussion_r1103314329

thomaseizinger:
Should we then set it unconditionally or still always pass the name and region in?

In other words, if you pass credentials, write to the cache, otherwise just read from it.

MarcoPolo:
default to the public name/region but allow you to override.

galargh:
Done through https://github.com/pl-strflt/tf-aws-bootstrap/pull/1

BTW, these are the options for global bucket ACLs - https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl

I set it to `public-read`

thomaseizinger:
What is the name of the bucket? I'd like to set it as the default for our action to have PRs also use this cache.

galargh:
It's https://github.com/pl-strflt/tf-aws-bootstrap/blob/cfe94aae3e3b7f7ea099f14f834dc1e5aac626df/libp2p.tf#L2

galargh:
@MarcoPolo I updated the policy again - https://github.com/pl-strflt/tf-aws-bootstrap/commit/d31fd08a4574bbe1d3104ccc81a72c994141c248. Apparently, the ready-made one doesn't involve Get on objects 🤷

thomaseizinger:
> http://s3.amazonaws.com/libp2p-by-tf-aws-bootstrap/buildCacheblobs/sha256:ffcc0b88f0be9d496e2dffcdb451a8c0e42b8f31b6807b31985570a532c5ecbd gives me access denied.

Just tried, this one works now!

MarcoPolo:
ty! I think we can close this.

PR #138, opened 2023-02-15
Nim interop support

### LINKED ISSUE #137: Add nim-libp2p test to interop tester
The implementation is similar to the ping test and should follow: https://github.com/libp2p/test-plans/blob/master/multidim-interop/README.md.

This has been on my todo list for a while since it might be fun to try Nim, but haven't gotten around to it. Anyone should feel free to take this.

cc @Menduist

Menduist:
cc @MarcoPolo 

Just translated the Go script to nim, didn't check if it works as I have no idea how to do that :D
Would appreciate some guidance

closes #137


mxinden (commented):
:rocket: first step should be to add nim here:

https://github.com/libp2p/test-plans/blob/2b40c181de24689f2ff29f2a595e46601c32b828/multidim-interop/Makefile#L1-L11

Menduist:
:green_circle: , though I would like to pin dependencies better before merging

MarcoPolo:
Nice work!

Fyi you can emit the compose file this runs by doing `npm test -- --name-filter="nim" --emit-only`. Then save one of those entries as a compose.yml and run it with `docker compose up`.

You can also run the listener and dialer directly with a local redis node for easier debugging. Or some combination of local/compose setup.

Also you may want to include this in the `nim-libp2p` repo so that you test changes against all released versions (including nim-libp2p 1.0). It should look something like this (Go example) https://github.com/libp2p/go-libp2p/tree/master/test-plans and the GHA workflow (https://github.com/libp2p/go-libp2p/blob/master/.github/workflows/interop-test.yml). This is very similar to what you have here.

Then, once you release versions in the future, the version here in this repo can simply reference the published commit along with how to build it (a simple Makefile) e.g. https://github.com/libp2p/test-plans/tree/master/multidim-interop/go/v0.25 for the last release of Go-libp2p.

If you want these builds to be faster you can use an S3 bucket to cache the layers. We might make our cache read-public in the future, follow along here: https://github.com/libp2p/test-plans/issues/132.

>  though I would like to pin dependencies better before merging

:+1: Tests here should be reproducible from a given commit. These dependencies should be pinned.

Menduist:
Yup thank you, after the 7th failed commit I figured out how to run it locally :D

I did something slightly different to what you proposed, since nim-libp2p is in v1, the test file should work for all v1.X versions, so I've left it here if that's OK

MarcoPolo:
> I did something slightly different to what you proposed, since nim-libp2p is in v1, the test file should work for all v1.X versions, so I've left it here if that's OK

It's okay for it to be the same test file. But the build should be reproducible from a given commit. If I build this today, then build it in a month from the same commit I should have the same result. Is that the case?

Menduist:
Yes, everything should be pinned (libp2p & its dependencies are installed from the libp2p folder which has a lockfile)

MarcoPolo (approved):
\o/ Thank you!

PR #146, opened 2023-02-24
Switch to redis:7-alpine

MarcoPolo:
It's 10x smaller at 30MiB vs >500 MiB. We don't need or want the full "redis-stack" (comparison: https://redis.io/download/)


marten-seemann (approved):

thomaseizinger (approved):
Thanks for making CI faster ❤️

PR #150, opened 2023-03-09
Add ws dialing test in browser

MarcoPolo:
This adds a websocket (unsecure) interop test. I'm not sure how useful this is, since in practice you will probably have to use wss. So I'll keep this as a draft for now.


MarcoPolo:
Closing this for now unless we find a need or desire for this.

PR #151, opened 2023-03-09
Add go-libp2p v0.26

MarcoPolo:
At what point should we start removing older versions?


marten-seemann:
> At what point should we start removing older versions?

Good question. Lotus is still running on v0.23.4, so it would be nice to still keep that one.

multidim-interop/go/v0.26/Makefile
```diff
@@ -0,0 +1,20 @@
+image_name := go-v0.26
+commitSha := 59a14cf3194d5d057c45cb1dbc7b1af3a116bc7a
```
marten-seemann, line 2: Any way we can use the tag name here?

marten-seemann (approved):

multidim-interop/go/v0.26/Makefile
MarcoPolo: I believe tags are mutable. The goal is for this always to work. If we could use a combination tag and sha I would be okay with that. Wdyt?

marten-seemann: On the Git level they are, but when using Go, they're not. The Go proxy makes sure that you can effectively never update a tag.

marten-seemann: So it should be fine to just use tags here, as it makes it more readable. No strong feelings though, we can update this in a later PR.

thomaseizinger: > On the Git level they are, but when using Go, they're not. The Go proxy makes sure that you can effectively never update a tag. 

This hash is used to download a Git repo though. How is the Go proxy involved here? What am I missing?

MarcoPolo: Yeah, but that doesn't guarantee that using a tag is immutable here. What I can do is use a tag and then compare the sha256 output of the artifact with some expected value. That way we get the tag semantics and the pinning semantics.

thomaseizinger:
> > > At what point should we start removing older versions?
> > 
> > Anything older than a year perhaps? I think going after time and not number of versions would be good.
> 
> Will that make the runtime of our CI explode? If I had to guess, I'd say we're releasing around 10 versions per year for go-libp2p.

Is that 10 minor versions or does that include patch releases?

PR #154, opened 2023-03-17
feat: run multidim-interop.yml on self-hosted runners

galargh:
This PR revives the proposal from https://github.com/libp2p/test-plans/pull/92 (https://github.com/libp2p/test-plans/commit/c53be4c4ac2f190acbd7d855d9560b9660570cdd).

Please note that the first build on self-hosted runners had to populate the cache https://github.com/libp2p/test-plans/actions/runs/4447196185/attempts/1. This suggests that some part of the machine config influences the cache keys. It might need some tweaks to make it interoperable with hosted GitHub Actions runners; or it might be a good idea to have another job which populates the cache for hosted GitHub Actions runners.

The subsequent runs were using cache as expected:
- https://github.com/libp2p/test-plans/actions/runs/4447196185/attempts/2
-  https://github.com/libp2p/test-plans/actions/runs/4447742213

I did not try to increase worker count nor modify docker run parameters to accommodate the more powerful environment.

This is the self-hosted runner definition: https://github.com/pl-strflt/tf-aws-gh-runner/blob/main/runners.tf#L28
This is the packer config the AMI was built from: https://github.com/pl-strflt/tf-aws-gh-runner/blob/main/images/ubuntu-jammy/kubo.pkrvars.hcl

We can modify the runners as you wish of course.


.github/actions/run-interop-ping-test/action.yml
```diff
@@ -65,12 +68,21 @@ runs:
...
       shell: bash
 
     - name: Run the test
       working-directory: ${{ steps.find-workdir.outputs.WORK_DIR }}
-      run: WORKER_COUNT=2 npm run test -- --extra-version=${{ inputs.extra-versions }} --name-filter=${{ inputs.test-filter }}
+      env:
+        WORKER_COUNT: ${{ inputs.worker-count }}
+        EXTRA_VERSION: ${{ inputs.extra-versions }}
+        NAME_FILTER: ${{ inputs.test-filter }}
+      run: npm run test -- --extra-version=$EXTRA_VERSION --name-filter=$NAME_FILTER
```
galargh, line 85: It's a good practice not to use template expansion inside `run` blocks.

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -65,12 +68,21 @@ runs:
 
     - name: Build images
       working-directory: ${{ steps.find-workdir.outputs.WORK_DIR }}
+      env:
+        AWS_BUCKET: ${{ inputs.s3-cache-bucket }}
+        AWS_REGION: ${{ inputs.aws-region }}
+        AWS_ACCESS_KEY_ID: ${{ inputs.s3-access-key-id }}
+        AWS_SECRET_ACCESS_KEY: ${{ inputs.s3-secret-access-key }}
```
galargh, line 75: This way, we modify the env only for this step.

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -54,9 +44,22 @@ runs:
...
+          echo "config=/etc/buildkit/buildkitd.toml" >> $GITHUB_OUTPUT
+        fi
+      shell: bash
+      id: buildkit
+
     - name: Set up Docker Buildx
       id: buildx
       uses: docker/setup-buildx-action@v2
+      with:
+        config: ${{ steps.buildkit.outputs.config }}
```
galargh, line 62: On standard, hosted GitHub Actions runners this will be an empty string. Which is the default.

thomaseizinger:
Half an hour is still pretty slow. This job usually finishes in 15min on PRs for us? Is that because we filter?

Will this mean we will also be able to have a cache for the PR docker builds that use `RUN --type=cache` (forgot the exact syntax).

MarcoPolo:
m5.large is 2 cpus and 8GB of ram. Isn't that roughly the same as the hosted runners? https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

galargh:
I moved you to c5.4xlarge, but, ultimately, you can pick any instance type that suits your workload characteristic the best - https://instances.vantage.sh/. 

I requested the job that runs on self-hosted (the one with cache) to use 16 workers here - https://github.com/libp2p/test-plans/pull/154/commits/255238ea7cd2ab9126d3edc6519bf09ade8f0d39

I also had to request the build part to remain sequential - https://github.com/libp2p/test-plans/pull/154/commits/0626f93a20f40bf8dc121d3134f07da6f582a658

The job - https://github.com/libp2p/test-plans/actions/runs/4465794780/jobs/7843205397?pr=154 - took 20 minutes to finish. It spent ~7 minutes building stuff and then ~13 minutes running the tests.

thomaseizinger:
> The job - [libp2p/test-plans/actions/runs/4465794780/jobs/7843205397?pr=154](https://github.com/libp2p/test-plans/actions/runs/4465794780/jobs/7843205397?pr=154) - took 20 minutes to finish. It spent ~7 minutes building stuff and then ~13 minutes running the tests.

That is an amazing improvement! With the filtering that we do in the pull-request workflows, that is likely to be even quicker!

thomaseizinger (approved):

thomaseizinger:
Is anything blocking this from merging? I'd be great to improve the runtime of this job.

MarcoPolo:
defer until after ipfs thing

.github/actions/run-interop-ping-test/action.yml
thomaseizinger: Could we land this refactoring in a separate PR perhaps? I think it is valuable and doesn't have to be blocked by the decision on how we configure the workers.

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -60,29 +44,56 @@ runs:
...
+        AWS_SECRET_ACCESS_KEY: ${{ inputs.s3-secret-access-key }}
       working-directory: ${{ steps.find-workdir.outputs.WORK_DIR }}
       run: npm run cache -- push
       shell: bash
 
     - name: Run the test
       working-directory: ${{ steps.find-workdir.outputs.WORK_DIR }}
-      run: WORKER_COUNT=2 npm run test -- --extra-version=${{ inputs.extra-versions }} --name-filter=${{ inputs.test-filter }}
+      env:
+        WORKER_COUNT: ${{ inputs.worker-count }}
```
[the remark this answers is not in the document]
thomaseizinger: @MarcoPolo What part of the stack is actually reading the `WORKER_COUNT` variable?

thomaseizinger (commented):
Thanks for the work here! I'd suggest we split this into two PRs so we can land parts of it quicker and discuss if we can perhaps avoid introducing the "input".

.github/actions/run-interop-ping-test/action.yml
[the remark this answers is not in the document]
thomaseizinger: The other thing I was just wondering is, would it make sense to set this to higher than the number of threads. If I understand correctly, then a lot of time is spent in the tests in docker booting up the containers. Is this actually a CPU bound task or could we perhaps run 4 or more of these with 2 threads as well?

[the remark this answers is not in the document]
galargh: Keeping worker count as an input will make testing this easier. Auto-detection sounds like a good idea too but I'd prefer not to add it to this PR. 

@MarcoPolo Would you mind giving it a review?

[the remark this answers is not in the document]
MarcoPolo: worker count is used by the test runner:
https://github.com/libp2p/test-plans/blob/master/multidim-interop/testplans.ts#L78

It's worth testing if more workers than CPUs would help here.

MarcoPolo:
rebased, fyi

MarcoPolo:
26 minutes on this branch versus 54 minutes on master. @galargh What's the cost to running this? It's nice it's twice as fast, but it's also currently free.

MarcoPolo (commented):
One open question around cost, but changes look good. Please request another review when the question has been answered.

MarcoPolo:
@galargh Friendly ping on my open question :)

galargh:
I totally missed the question about cost! Let me do a quick calculation for you.

The monthly cost of moving on with this should be around $38/month (see the AWS resource calculation - https://calculator.aws/#/estimate?id=9a5a201262a70d9f491015a5aa81a4aeaf1ae966). I based this on:
- the current configuration of [4xlarge runner](https://github.com/pl-strflt/tf-aws-gh-runner/blob/main/runners.tf)
- 66 `libp2p multidimensional interop test` workflow runs in the past 30 days
- 36 minutes `run-multidim-interop` job took to complete on this PR
- ~3GB of new data in the S3 bucket in the past 30 days

thomaseizinger:
With the recently added implementations, the interop test are now by far our longest CI job at > 20 minutes: https://github.com/libp2p/rust-libp2p/actions/runs/5387398505/jobs/9778664726

What is the process around deciding whether the spend is worth it?

.github/workflows/multidim-interop.yml
```diff
@@ -9,14 +9,15 @@ name: libp2p multidimensional interop test
...
-    runs-on: ubuntu-latest
+    runs-on: ['self-hosted', 'linux', 'x64', '4xlarge'] # https://github.com/pl-strflt/tf-aws-gh-runner/blob/main/runners.tf
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/run-interop-ping-test
         with:
           s3-cache-bucket: libp2p-by-tf-aws-bootstrap
           s3-access-key-id: ${{ vars.S3_AWS_ACCESS_KEY_ID }}
           s3-secret-access-key: ${{ secrets.S3_AWS_SECRET_ACCESS_KEY }}
+          worker-count: 16
```
galargh, line 20: If we want to stick to hosted in this repo, let's apply this suggestion. I'd still want to have this PR merged because it'd enable other repos to modify worker count and fully take advantage of self-hosted config.

```suggestion
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - uses: ./.github/actions/run-interop-ping-test
        with:
          s3-cache-bucket: libp2p-by-tf-aws-bootstrap
          s3-access-key-id: ${{ vars.S3_AWS_ACCESS_KEY_ID }}
          s3-secret-access-key: ${{ secrets.S3_AWS_SECRET_ACCESS_KEY }}
```

thomaseizinger: I am in favor of that.

Although I'd assume that the traffic on this repo in terms of PRs is negligible compared to our repos. That being said, I don't really care about how long CI takes in here so I think it makes sense to stick to hosted runners in this repo.

MarcoPolo (approved):

MarcoPolo:
Will merge after CI passes. Thank you so much @galargh! ❤️

PR #159, opened 2023-03-24
Add perf dashboard

mxinden:
Opening your branch @MarcoPolo as a draft pull request in order to comment.


perf-dashboard/README.md
```diff
@@ -0,0 +1,62 @@
...
+# Benchmark binary
+
+This is per implementation. It's a binary that accepts the following flags:
+
+| Flag                 | description                                            |
+| -------------------- | ------------------------------------------------------ |
+| parallel_connections | number of parallel connections to use                  |
+| upload_bytes         | number of bytes to upload to server per connection     |
+| download_bytes       | number of bytes to download from server per connection |
+| n_times              | number of times to do the perf round trip              |
```
mxinden, line 52: Is this the number of times one executes a libp2p perf round trip (open stream, sent # bytes to receive, sent bytes, receive bytes) per connection? Or is this the number of times one should execute the whole benchmark?

In case of the former, we still need a parameter on how often the binary should execute the whole test, otherwise it can not emit aggregates like avg, min, max or p95.

perf-dashboard/benchmark-data-schema.json
```diff
@@ -0,0 +1,45 @@
...
+                        "type": "string"
+                    }
+                },
+                "required": [
+                    "name",
+                    "unit",
+                    "result",
+                    "implementation",
+                    "stack",
+                    "version"
```
mxinden, line 36: If I am not mistaken, these are missing a property definition above.

mxinden:
Status update: [Latest iteration of rust-libp2p perf client](https://github.com/libp2p/rust-libp2p/pull/3646) now prints the following JSON:

``` json
{
  "benchmarks": [
    {
      "name": "Single Connection throughput – Upload",
      "result": "1270317.353942249",
      "unit": "bits/s"
    },
    {
      "name": "Single Connection 1 byte round trip latency 100 runs 95th percentile",
      "result": "0.138944185",
      "unit": "s"
    }
  ]
}
```

perf-dashboard/README.md
MarcoPolo: > this the number of times one should execute the whole benchmark

Yes. I was thinking this would be shared amongst all connections. So you could say have n_times = 100, and parallel_connections = 10 and have an average of 10 roundtrips per conn, but some may have more or less in practice.

MarcoPolo: Here are all the params the msquic tool accepts: https://github.com/microsoft/msquic/tree/main/src/perf

perf-dashboard/benchmark-data-schema.json
MarcoPolo: Oh this is the wrong schema, sorry. Use perf-dashboard/benchmarks.schema.json instead.

MarcoPolo:
Sorry for the confusion in having two schema files. The correct one is `perf-dashboard/benchmarks.schema.json`. The output above is for the wrong schema.

perf-dashboard/benchmark-data-schema.json
MarcoPolo: Deleted the wrong one.

perf-dashboard/example-data.json
```diff
@@ -0,0 +1,161 @@
+{
+    "$schema": "./benchmarks.schema.json",
+    "benchmarks": [
+        {
+            "name": "Single Connection throughput – Upload",
+            "unit": "bits/s",
+            "comparisons": [
+                {
+                    "name": "http",
```
BigLep, line 9: I assume https instead of http?

Also, do we want to be more clear than just "https"?  Should we specify whether this is using go stdlibrary, rust stdlibrary, curl, etc?  (and if so, I assume the version of go, rust, curl, etc. matters).  At which point is a "comparison" very different than a result?

perf-dashboard/example-data.json
```diff
@@ -0,0 +1,161 @@
+{
+    "$schema": "./benchmarks.schema.json",
+    "benchmarks": [
+        {
+            "name": "Single Connection throughput – Upload",
```
BigLep, line 5: Where do I see n, p0, p95,p100, and avg ?

perf-dashboard/README.md
```diff
@@ -0,0 +1,62 @@
...
+| close_connection     | bool. Close the connection after a perf round trip     |
+
+The binary should report the following:
+
+* Stats on the length of the run:
```
BigLep, line 57:
```suggestion
* Stats on the durations of the `n_times` runs:
```

I assume that is correct (and I think makes it more clear)

perf-dashboard/README.md
```diff
@@ -0,0 +1,62 @@
...
+  * Total, Avg, Min, Max, p95
```
BigLep, line 58: Rather than special-casing Min / Max, just do p0 and p100 ?

perf-dashboard/README.md
```diff
@@ -0,0 +1,62 @@
...
+* Round trips per second:
```
BigLep, line 59: I'm trying to follow what this means.  

I think we probably needs some vocabulary here.  
I'm open to different terms, but I'm assuming we have a "test run" which will do "n_times" round trips.

I assume this means that at the end of the "test run" we eimit:
avg number of round trips per second that were completed
p0 number of round trips per second
p100 number of round trips per second
p95 number of round trips per second
etc.

This means that if the benchmark runner does a "test run" that lasts for ~10 seconds, it will be tracking for each of those seconds the number of round trips that complete each second so that it can then compute the avg/p0/p95/p100 stats.  

Is that right?

BigLep (commented):
I think this PR is really useful for defining the end state and making sure we agree on the interface.

It was clear from me in the schema's where we define the "stat" (e.g., avg/p0/p95/p100).

PR #161, opened 2023-03-31
Cache final images instead of build layers

MarcoPolo:
I think this will make @thomaseizinger very happy ;)

In an effort to speed things up, this changes the caching strategy from relying on Docker's somewhat opaque rules to simply caching the generated image. Here's how it works:

1. We define some cache key. Generally this is either the commit we're building from or the hash of the input files. Up to the implementation to decide.
2. We check s3 directly to see if we have a file named `$IMAGE_NAME-$CACHE_KEY-$ARCH.tar.gz`, if we do import that image.
3. No? Build as normal.
4. If we have the `PUSH_CACHE` env we can push the built image to s3.

Q: Why not a GH registry?
A: Then we have to deal with GC. This lets us rely on the s3 behavior of a lifetime for objs.

Q: Why not an S3 registry?
A: That requires a bit more work to set up. You need a registry container running in the background. This is simpler. It also supports public read only access to the cache.

### Benefits

1. Faster CI runs for everyone. Less time spent building, even if only cached (I noticed one js-libp2p PR took 8 minutes before starting the test).
2. More predictable caching. Caching strategy is dead simple here so it's easy to see why a cache missed happened.
3. Anonymous S3 Reads. Everyone can use the cache
4. Faster downloads of images. If the final image of the container-mage is small (it should be), then this is significantly less data than getting every single layer and rebuilding.

### Why didn't we do this from the start?

One step at a time. The build layer caching strategy was convenient and it was correct. I was hoping it would be better, but here we are.

Note that this solution is different (and imo better) than each implementation managing its own images and pushing to a registry. I say better since only this repo needs to worry about pushing images and GC'ing them.

### Why now?

@achingbrain blocked my [PR](https://github.com/libp2p/js-libp2p/pull/1615) until I make interop tests go vroom.


multidim-interop/dockerBuildWrapper.sh
```diff
@@ -1,13 +1,45 @@
 #!/usr/bin/env /bin/bash
+set -eou pipefail
+set -x
 
+# Print the hash of our current working directory.
+# We'll use this to store the build cache in S3.
+CWD_HASH=$(tar --sort=name \
+      --mtime="1970-01-01 00:00Z" \
+      --owner=0 --group=0 \
+      -cvf - . | sha256sum | sed 's/.$//' | xargs) # Sed removes the trailing dash, xargs removes the spaces
```
thomaseizinger, line 10: This hashes the contents of all files, correct?

multidim-interop/src/compose-runner.ts
```diff
@@ -43,14 +44,11 @@ export async function run(namespace: string, compose: ComposeSpecification, opts
...
-        const { stdout, stderr } =
-            (await Promise.race([
-                exec(`docker compose -f ${path.join(dir, "compose.yaml")} up ${upFlags.join(" ")}`),
-                // Timeout - uses any type because this will only reject the promise.
-                new Promise<any>((resolve, reject) => { timeoutId = setTimeout(() => reject("Timeout"), 1000 * timeoutSecs) })
-            ]))
+        const timeoutId = setTimeout(() => controller.abort(), 1000 * timeoutSecs)
+
+        const controller = new AbortController();
+        const { signal } = controller;
```
thomaseizinger, line 50: I think you can merge these.

thomaseizinger (commented):
Nice idea. We are getting pretty close to a registry 😁

The fingerprinting might have bugs. As we all know, cache invalidation is hard :)

Does S3 support a PUT-like operation? We could use that to have the user supply the cache key, much like we would supply the name of the docker image for a registry.

What do you think?

multidim-interop/dockerBuildWrapper.sh
MarcoPolo: Yes. And the structure

MarcoPolo:
> The fingerprinting might have bugs

I think if we limit the context to the directory (or subdirectory) where the build script is called from and the builds are reproducible, then this shouldn’t be a problem since we are hashing all the inputs. Is there some bug you had in mind?

Note this is an input hash addressed scheme rather than content hash addressed scheme. There are trade offs here, but this is generally a bit more straightforward.

MarcoPolo:
Ah, I think I understand what you mean. Instead of this being part of the wrapper, we have the caller give us a key to use. We can expose helpers to do the correct thing. But this could be useful if the caller wants to hash one file instead of a directory because they know the build is reproducible from that file. e.g. The caller can pass in a commit hash and Dockerfile hash as the key since it'll be downloading the repo at that commit anyways.

thomaseizinger:
More or less. I was thinking that if we just supply a key, we could basically emulate a registry. With a registry, I'd push a new version of the `rust-libp2p-ping-interop-test` image every time `master` changes. I don't really care about the hash of all inputs at this point. That fact that we pushed to master has authority.

Perhaps I am too influenced from how we do caching in `rust-libp2p`: We treat all caches for PRs as read-only and always update all caches on pushes to master.

MarcoPolo:
I don't want to rebuild and republish unnecessarily.

One change I'm trying to make to speed things up is to check if we get a cache hit before downloading the source to build from. This will give us a bit of a speed up if we do hit since we don't need to fetch the repo source.

thomaseizinger:
> I don't want to rebuild and republish unnecessarily.
> 
> One change I'm trying to make to speed things up is to check if we get a cache hit before downloading the source to build from. This will give us a bit of a speed up if we do hit since we don't need to fetch the repo source.

That is getting to a point where referencing a tagged image from a registry [here](https://github.com/libp2p/test-plans/blob/master/multidim-interop/versions.ts#L28) would give you all of that without any code in here.

What if we combine two solutions:

1. For images of released versions, push them to a registry. They never change unless we make patch releases and publish upon those can be automated.
2. For the latest head, we pass in an image hash from an image we just built in the PR.

thomaseizinger:
With "they never change", I meant that there is never a need to update things in here (IIRC, that updates in here require releasing an image somewhere else was part of the criticism back when we discussed this).

We would release a new image even for the patch release, meaning the old container would always be the same. We can pin its hash too if we want to be absolutely sure.

A registry still breaks constraint 2 but it can never fulfill that. It also doesn't have to IMO. If we go as far as "not downloading the code if the cache doesn't exist" then we are just almost building a registry in S3 but without the support of having docker automatically pull it.

.github/actions/run-interop-ping-test/action.yml
```diff
@@ -65,12 +65,12 @@ runs:
 
     - name: Build images
       working-directory: ${{ steps.find-workdir.outputs.WORK_DIR }}
-      run: make
+      run: make -j 2
```
thomaseizinger, line 68: Is there a reason to still use make here? We could as well just use typescript for this step.

Going even further, we could make it a "proper" action that is published to the GitHub marketplace. That would allow us to move away from a `composite` one to either Docker (and allow us to write it in go or Rust) or JavaScript.

MarcoPolo:
This makes the checks go from 40+ min to ~25 min.

.github/actions/run-interop-ping-test/action.yml
MarcoPolo: To add another point, this is the root directory of these tests. Make here does nothing but call make in the implementation folders. I wouldn't want to remove this make. It's a good interface point such that implementations know they have to provide a Makefile that generates a image.json. The rest can be done however they want.

multidim-interop/rust/v0.48/Makefile
```diff
@@ -1,12 +1,26 @@
...
 commitSha := 7b3047d6d05d599f11f05938d4257e70de66ac12
+cacheKey := ${commitSha}
+shouldUseCache := ${shell IMAGE_NAME=${imageName} CACHE_KEY=${cacheKey} ../../helpers/shouldUseCache.sh}
 
 all: image.json
 
-image.json: rust-libp2p-${commitSha}
-	cd rust-libp2p-${commitSha} && IMAGE_NAME=${image_name} ../../../dockerBuildWrapper.sh -f interop-tests/Dockerfile .
-	docker image inspect ${image_name} -f "{{.Id}}" | \
+ifneq (${shouldUseCache},)
```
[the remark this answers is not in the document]
MarcoPolo: Each version is considered a separate thing. There's no dependency between them. We can remove duplication at the expense of interlinking these things. I don't think that's worth it.

This code is also relatively straightforward. But at a high level, I agree make and bash is not the best solution. However, rewriting all this in X will give us no real benefit at the end of the day. If you want to make this better, I'd suggest looking into how to improve the test runner speed, each test takes about 2s. Can we speed that up? Updating that will give us the most bang for our buck.

MarcoPolo:
> A registry still breaks constraint 2 but it can never fulfill that. It also doesn't have to IMO. 

If you break constraint 2 like that you'll also break constraint 1. You aren't building from a commit. You're building from a commit + the state of the registry.

MarcoPolo:
Also this should significantly speed up the Nim-libp2p and Zig-libp2p since they'll now be able to take advantage of the public cache. Should shave 25min from those runs. (Previously builds would take 30min, but now everyone can use the cache which means setup should only be <5min).

MarcoPolo:
@thomaseizinger / @mxinden This system also benefits from a minimal final image size. It would be great if one of you could update the Dockerfile to use multistage builds and only emit the minimal image.

multidim-interop/go/v0.23/Makefile
```diff
@@ -1,11 +1,27 @@
-image_name := go-v0.23
+SHELL := /usr/bin/env bash
+imageName := go-v0.23
 
+cacheKey=${shell find . -type f | sort | grep -v image.json  | tr "\n" " " | xargs ../../helpers/hashFiles.sh}
```
marten-seemann, line 4: This is a lot of duplicated, non-trivial code. Would it make sense to split out the common parts into a separate script?

marten-seemann (commented):
This is a massive amount of code changes that are not well-contained. There’s a lot of non-obvious steps that have to inserted into every build file of every image. 

Are we certain that massively complicating our setup is worth the speed up?
Specifically, can’t we get most of the benefits by using normal Docker caching when we use multistage builds for the implementations that output a binary?

multidim-interop/go/v0.23/Makefile
MarcoPolo: Happy to make that change in a follow up PR. But I don’t think this is that complicated. This doesn’t actually need the sort call since that’s handled by the helper.

We could also have the helper read from stdin instead of the args and that would remove the `tr`.

Long term though this will go away as all the future versions are defined in the implementation repos and the cache key becomes the commit.

MarcoPolo:
> This is a massive amount of code changes that are not well-contained. There’s a lot of non-obvious steps that have to inserted into every build file of every image.

Disagree. Yes, there’s a lot of duplication, but that’s because each thing is contained. The steps don’t have to be inserted. The worst that happens is the image isn’t cached.

Could this be changed? Sure. Maybe rewriting it in a simpler scripting language would help and keep things DRY. Makefiles aren’t that flexible unfortunately. I would be open to that change in a future PR. For now this gives us a pretty big speed improvement that we want for getting js-libp2p test.

> Are we certain that massively complicating our setup is worth the speed up?

Yes. This shaves of 10 min for every run. Even more if you didn’t have the aws s3 keys setup (e.g. nim).

> Specifically, can’t we get most of the benefits by using normal Docker caching when we use multistage builds for the implementations that output a binary?

I don’t follow. We were using docker layer caching before. Could you expand?

This is blocking getting js-libp2p tests in, which is blocking getting the latest js-libp2p version tested, which is blocking getting js-libp2p-webrtc and browser-to-browser tests in. I’m happy to iterate on this in the future, but we kind of need this sooner rather than later.

marten-seemann:
> > Specifically, can’t we get most of the benefits by using normal Docker caching when we use multistage builds for the implementations that output a binary?
> 
> I don’t follow. We were using docker layer caching before. Could you expand?

We haven't used multi-stage builds though. Would using a multi-stage build speed up restoring from the cache, as Docker now presumably would only need to download the layers from the second stage, and can skip all builder stage layers?

MarcoPolo:
> > > Specifically, can’t we get most of the benefits by using normal Docker caching when we use multistage builds for the implementations that output a binary?
> > 
> > 
> > I don’t follow. We were using docker layer caching before. Could you expand?
> 
> We haven't used multi-stage builds though. Would using a multi-stage build speed up restoring from the cache, as Docker now presumably would only need to download the layers from the second stage, and can skip all builder stage layers?

I don’t think so. 

I have a small change I can make to reduce a lot of the code here. I’ll make that soon

thomaseizinger:
> @thomaseizinger / @mxinden This system also benefits from a minimal final image size. It would be great if one of you could update the Dockerfile to use multistage builds and only emit the minimal image.

I'll open an issue on the `rust-libp2p` repository.

PR #162, opened 2023-04-01
Fix: Correctly abort process on timeout

MarcoPolo:


multidim-interop/src/compose-runner.ts
```diff
@@ -43,14 +44,11 @@ export async function run(namespace: string, compose: ComposeSpecification, opts
...
-        let timeoutId
-        const { stdout, stderr } =
-            (await Promise.race([
-                exec(`docker compose -f ${path.join(dir, "compose.yaml")} up ${upFlags.join(" ")}`),
-                // Timeout - uses any type because this will only reject the promise.
-                new Promise<any>((resolve, reject) => { timeoutId = setTimeout(() => reject("Timeout"), 1000 * timeoutSecs) })
-            ]))
+        const timeoutId = setTimeout(() => controller.abort(), 1000 * timeoutSecs)
+
+        const controller = new AbortController();
```
thomaseizinger, line 49: Isn't `controller` undefined in the closure? Even if this works, can we declare `controller` first please?

multidim-interop/src/compose-runner.ts
```diff
@@ -43,14 +44,11 @@ export async function run(namespace: string, compose: ComposeSpecification, opts
...
+        const { signal } = controller;
+        const { stdout, stderr } = await exec(`docker compose -f ${path.join(dir, "compose.yaml")} up ${upFlags.join(" ")}`, { signal })
```
thomaseizinger, line 51: If `exec` needs an object that has a `signal` property, can't you just pass `controller` directly?

thomaseizinger (commented):
Thanks for splitting this out into a separate PR.

multidim-interop/src/compose-runner.ts
MarcoPolo: This avoids passing in other things that are in the controller. This is also how you do it per the docs: https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback

multidim-interop/src/compose-runner.ts
MarcoPolo: It's tricky, but L49 will always happen before the closure. I'll update this anyways.

thomaseizinger (approved):

Issue #63, opened 2022-10-07
F. ⏱ Future-proof Benchmarking


p-shahi:

BigLep:
@mxinden : I know you're doing blog drafting at https://observablehq.com/@mxinden-workspace/libp2p-perf -  thanks!

A couple of things that came to mind while looking at it:
1. Is this just where we're drafting, or something we expect to point users to?
2. If pointing users to it, can we create a libp2p Observable namespace?
3. What is going to be the presence on the libp2p blog about this?

(no pressure to respond now - I don't mean to be catching you in-progress.  If this has been though through and you just haven't gotten to any of this yet, that is fine.  I am just leaving notes now while it's coming to mind.)

mxinden:
**Status Update**

https://github.com/libp2p/test-plans/pull/184 and thus a first iteration of the libp2p performance benchmark setup is merged. Follow-up / future work in tracked in the description of this issue.

mxinden:
Crossposting a Slack message here, for those following through GitHub only.

> *First iteration of the libp2p performance benchmarking setup merged*
> 
> More specifically https://github.com/libp2p/test-plans/pull/184.
> 
> Let's try it out. We recently released a new version of rust-libp2p, v0.52.0. Now we can add it to the benchmarking setup, see how well it performs.
> 
> 1. Create a pull request against libp2p/test-plans adding the new version https://github.com/libp2p/test-plans/pull/202. (Note that in this particular case we are updating an existing version, as we have been already testing a pre-release of v0.52.0.)
> 2. Trigger the _libp2p perf test_ GitHub workflow for the branch of the pull request https://github.com/libp2p/test-plans/actions/workflows/perf.yml
> 3. Follow the workflow run, which eventually pushes the benchmark results to the branch https://github.com/libp2p/test-plans/actions/runs/5351995316/jobs/9706612810
> 4. See the new commit on your pull request https://github.com/libp2p/test-plans/pull/202 https://github.com/libp2p/test-plans/pull/202/commits/b94960f74d0a2d85b429c311295721bc6d3f6e9f
> 5. Visualize the benchmark results via our ObservableHQ dashboard. Just make sure to choose the right branch, in this case `perf-rust-libp2p-v0.52` in the `Branch` input https://observablehq.com/@mxinden-workspace/libp2p-performance-dashboard
> 
[...]

PR #163, opened 2023-04-04
feat(perf): add (provision, build, run) tooling 

mxinden:
See `perf/README.md` and overarching tracking issue https://github.com/libp2p/test-plans/issues/63.

Outstanding work:

- [x] Remove libp2p specific concepts (e.g. multiaddr) e.g. to no longer require the https example to parse a multiaddr.
- [x] Rename public ssh key
- [x] Remove `--n-times` flag from binaries. Potentially introduce `--n-parallel-requests` flag in the future. Multiple _sequential_ runs are now done in the runner implementation.
- [x] Transfer and run binaries instead of docker images / containers. Or build on the machines.
- [x] Configurable Server port. See https://github.com/libp2p/test-plans/pull/163/#discussion_r1183459190.
- [ ] Remove the need for static peer IDs, i.e. remove `--secret-key-seed`.
- [ ] Resource utilization benchmark?
- [x] Consider adding benchmark parameters to output json file. See https://github.com/libp2p/test-plans/pull/163#pullrequestreview-1382950014.
- [ ] Add js-libp2p https://github.com/libp2p/js-libp2p/pull/1604
- [ ] Add back zig-libp2p
- [x] Move to TLS only (for now). Easier to compare with HTTPs and no additional security protocol negotiation.
- [ ] Support testing other security protocols, i.e. Noise.


mxinden:
Test run between server in us-west-1 and client in us-east-1:

```
➜  perf git:(perf-terraform) ✗ ssh ec2-user@$(terraform output -raw client_public_ip) sudo docker run --tty --rm --entrypoint perf-client mxinden/libp2p-perf --server-address /ip4/$(terraform output -raw server_public_ip)/tcp/4001

[2023-04-06T18:34:17Z INFO  perf_client] Start benchmark: round-trip-time latency
[2023-04-06T18:34:48Z INFO  perf_client] Finished: 56 pings in 30.3030s
[2023-04-06T18:34:48Z INFO  perf_client] - 0.5399 s median
[2023-04-06T18:34:48Z INFO  perf_client] - 0.5499 s 95th percentile
    
[2023-04-06T18:34:48Z INFO  perf_client] Start benchmark: single connection single channel throughput
[2023-04-06T18:35:15Z INFO  perf_client] Finished: sent 10.00 MiB in 13.56 s and received 10.00 MiB in 13.66 s
[2023-04-06T18:35:15Z INFO  perf_client] - 5.90 MiBit/s up
[2023-04-06T18:35:15Z INFO  perf_client] - 5.86 MiBit/s down
    
[2023-04-06T18:35:15Z INFO  perf_client] Start benchmark: single connection parallel requests per second
[2023-04-06T18:35:16Z INFO  perf_client] Finished: sent 1000 1 bytes requests with 1 bytes response each within 0.40 s
[2023-04-06T18:35:16Z INFO  perf_client] - 2493.58 req/s
    
[2023-04-06T18:35:16Z INFO  perf_client] Start benchmark: sequential connections with single request per second
[2023-04-06T18:35:47Z INFO  perf_client] Finished: established 39 connections with one 1 bytes request and one 1 bytes response within 30.81 s
[2023-04-06T18:35:47Z INFO  perf_client] - 0.2488 s 95th percentile connection establishment
[2023-04-06T18:35:47Z INFO  perf_client] - 0.8114 s 95th percentile connection establishment + one request

{"benchmarks":[{"comparisons":[],"name":"Single Connection throughput – Upload","results":[{"implementation":"rust-libp2p","result":773013.3502842287,"transportStack":"TODO","version":"TODO"}],"unit":"bits/s"},{"comparisons":[],"name":"Single Connection 1 byte round trip latency 95th percentile","results":[{"implementation":"rust-libp2p","result":0.811350517,"transportStack":"TODO","version":"TODO"}],"unit":"s"}]}
```

mxinden:
Recent commits add a runner implementation, running a Rust Perf docker image on the AWS client once with TCP, once with QUIC:

```
➜  runner git:(perf-terraform) npm run start -- --client-public-ip $(terraform output -raw -state ../terraform/terraform.tfstate client_public_ip) --server-public-ip $(terraform output -raw -state ../terraform/terraform.tfstate server_public_ip)

> runner@1.0.0 start
> node ./dist/index.js --client-public-ip xxx --server-public-ip xxx

[2023-04-08T18:42:55Z INFO  perf_client] Start benchmark: custom
[2023-04-08T18:43:02Z INFO  perf_client] Finished: Established 10 connections uploading 1 and download 1 bytes each
[2023-04-08T18:43:07Z INFO  perf_client] Start benchmark: custom
[2023-04-08T18:43:09Z INFO  perf_client] Finished: Established 10 connections uploading 1 and download 1 bytes each
{
  "benchmarks": [
    {
      "name": "",
      "unit": "s",
      "results": [
        {
          "result": [
            0.53530147,
            0.538841711,
            0.536516464,
            0.537959156,
            0.53936546,
            0.543814705,
            0.537634717,
            0.539665182,
            0.536461729,
            0.541753673
          ],
          "implementation": "",
          "version": "rust-master",
          "transportStack": ""
        },
        {
          "result": [
            0.12290573,
            0.125007226,
            0.123338546,
            0.122678248,
            0.123330097,
            0.12530641,
            0.120919996,
            0.125090132,
            0.122547649,
            0.121178493
          ],
          "implementation": "",
          "version": "rust-master",
          "transportStack": ""
        }
      ],
      "comparisons": []
    }
  ]
}
```

perf/runner/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+            break;
+        default:
+            console.error("Unsupported transport stack ${args.transportStack}");
+            process.exit(1);
+
+    }
+
+    const binFlags = `--server-address ${serverAddress} --upload-bytes ${args.uploadBytes} --download-bytes ${args.downloadBytes} --n-times ${args.nTimes}`
+    // TODO Take docker hub repository from version.ts
+    const dockerCMD = `docker run --rm --entrypoint perf-client mxinden/libp2p-perf@sha256:${args.dockerImageId} ${binFlags}`
```
MarcoPolo, line 118: I assume we don't want the entrypoint or the `mxinden/libp2p-perf` here, right?

MarcoPolo:
Where is the rust implementation?

mxinden:
> Where is the rust implementation?

Rust implementation is here:

https://github.com/libp2p/rust-libp2p/pull/3646

perf/runner/src/index.ts
mxinden: Nit pick, the entrypoint is `perf-client`. Either way, yes, I will change the Rust image to use `perf-client` by default. In other words, for the JS and Go image, no `--entrypoint` should be needed.

perf/runner/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+            serverAddress = `/ip4/${args.serverPublicIP}/udp/4001/quic-v1`;
+            break;
+        default:
+            console.error("Unsupported transport stack ${args.transportStack}");
+            process.exit(1);
+
+    }
+
+    const binFlags = `--server-address ${serverAddress} --upload-bytes ${args.uploadBytes} --download-bytes ${args.downloadBytes} --n-times ${args.nTimes}`
+    // TODO Take docker hub repository from version.ts
```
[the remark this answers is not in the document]
mxinden: > We should assume we have the image here. And not load anything here.

Can you expand on what benefit not-loading would have here? Why have a separate step for loading.

> building things

Note that nothing here is building things.

MarcoPolo:
Pushed a Go implementation, could you try it out? 

I'm curious if we'll see the `2023/04/12 04:27:38 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/quic-go/quic-go/wiki/UDP-Receive-Buffer-Size for details.` message. That would be a problem.

mxinden:
Thanks @MarcoPolo. WIll update here once I have it integrated.

MarcoPolo:
Here's the zig-libp2p version: https://github.com/MarcoPolo/zig-libp2p/tree/marco/perf-runner/examples/perf. I don't expect the perf to be amazing here because it's using the test setup rather than a perf oriented setup, and I have one outstanding perf bug that I haven't tracked down yet.

I haven't set up the environment yet, but it's on my list of things to do :)

Also after playing around with this some more, I think we should add another flag for `--run-server` that runs the server mode of this instead of mucking around with the entrypoints.

mxinden:
> Here's the zig-libp2p version: https://github.com/MarcoPolo/zig-libp2p/tree/marco/perf-runner/examples/perf. 

Great. 

> Also after playing around with this some more, I think we should add another flag for `--run-server` that runs the server mode of this instead of mucking around with the entrypoints.

Fine by me. I can merge the client and server implementation into one binary on the Rust side.

---

Order on my TODO list today:

1. Deploy image to both client and server in order to test Go against Go and Rust against Rust.
2. Change machine and AMI types.
3. Load output into Observable HQ.
4. Integrate Zig libp2p.

I don't think I will get to (3) today. Will continue where I left off tomorrow.

Sounds good @MarcoPolo?

mxinden:
@MarcoPolo since we talked about it, https://github.com/libp2p/test-plans/pull/163/commits/24333385107a7bc7b6fb1df833bb5be8e6fbf7c8 addresses the UDP receive buffer being to small.

```
2023/04/13 17:37:19 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/quic-go/quic-go/wiki/UDP-Receive-Buffer-Size for details.
```

mxinden:
Planning to experiment with `m5n.8xlarge` with 32 vcores, 128 GB memory and 25 GBit network bandwidth next.

https://aws.amazon.com/ec2/instance-types/

perf/impl/go/v0.27/main.go
```diff
@@ -0,0 +1,109 @@
...
+	var opts []libp2p.Option
+	if *runServer  {
+		opts = append(opts, libp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/4001", "/ip4/0.0.0.0/udp/4001/quic-v1"))
+
+		// TODO: Fake identity. For testing only.
+		priv, _, err := crypto.GenerateEd25519Key(&simpleReader{seed:uint8(*secretKeySeed)})
+		if err != nil {
+			panic(err)
+		}
+		opts = append(opts, libp2p.Identity(priv))
```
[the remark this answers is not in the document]
MarcoPolo: Is this so that you can keep the same peer id on the server?

MarcoPolo:
> Explore comparison options (e.g. ping, iperf or http benchmark)

We need to make sure this is HTTPS or TCP/TLS. We don't want to compare ourselves against a non-crypto benchmark since that's completely different.

MarcoPolo:
I merged master in here fyi

perf/impl/go/v0.27/main.go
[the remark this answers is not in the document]
MarcoPolo: Can we have the rust server print out the multiaddr(s) on start and parse that from stdout. I'd much rather do that then copy this logic to zig and js

MarcoPolo:
Added some small changes. Also updated the plot to process the data: https://observablehq.com/@realmarcopolo/libp2p-perf (only the 10MiB throughput)

mxinden:
Latest result with a `m5n.8xlarge` for both server and client:

``` json
{
  "benchmarks": [
    {
      "name": "Single Connection throughput – Upload 10 MiB",
      "unit": "s",
      "results": [
        {
          "result": [
            13.026514164
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "tcp"
        },
        {
          "result": [
            0.707346464
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            0.8061137
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "tcp"
        },
        {
          "result": [
            0.61057115
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "quic-v1"
        }
      ],
      "comparisons": []
    },
    {
      "name": "Single Connection throughput – Download 10 MiB",
      "unit": "s",
      "results": [
        {
          "result": [
            13.991809147
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "tcp"
        },
        {
          "result": [
            6.114455213
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            0.82152283
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "tcp"
        },
        {
          "result": [
            0.6158309
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "quic-v1"
        }
      ],
      "comparisons": []
    },
    {
      "name": "Single Connection 1 byte round trip latency",
      "unit": "s",
      "results": [
        {
          "result": [
            0.554760062,
            0.558321847,
            0.555003099,
            0.562800442,
            0.555066646,
            0.580909281,
            0.539751891,
            0.559987914,
            0.582082051,
            0.558601806
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "tcp"
        },
        {
          "result": [
            0.12941641,
            0.125653117,
            0.1291954,
            0.121426498,
            0.125761626,
            0.123739958,
            0.126287167,
            0.122107052,
            0.119937935,
            0.126166054
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            0.12355031,
            0.062096734,
            0.06215763,
            0.062082987,
            0.062174562,
            0.062082175,
            0.06209195,
            0.062078677,
            0.062102236,
            0.062094487
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "tcp"
        },
        {
          "result": [
            0.06488921,
            0.064933814,
            0.06498787,
            0.064782694,
            0.0647692,
            0.064765766,
            0.064823076,
            0.0647876,
            0.06480806,
            0.06482354
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "quic-v1"
        }
      ],
      "comparisons": []
    }
  ]
}
```

perf/terraform/region/main.tf
```diff
@@ -78,15 +78,15 @@ resource "aws_security_group" "restricted_inbound" {
   }
 
   ingress {
-    from_port   = 4001
-    to_port     = 4001
+    from_port   = 1
+    to_port     = 65535
```
mxinden: Why change the ingress port @MarcoPolo? For your zig server you only need egress, no?

perf/impl/go/v0.27/perf.go
```diff
@@ -0,0 +1,128 @@
...
+func (ps *PerfService) sendBytes(ctx context.Context, s network.Stream, bytesToSend uint64, buf []byte) error {
+	for bytesToSend > 0 {
+		toSend := buf
+		if bytesToSend < BlockSize {
+			toSend = buf[:bytesToSend]
+		}
+
+		n, err := s.Write(toSend)
+		if err != nil {
+			return nil
```
mxinden, line 73:
```suggestion
			return err
```

Right @MarcoPolo?

mxinden: Would be resolved with 2cceb89.

mxinden:
Latest results. Though caveat, "Single connection test" in rust is a new connection each time.

``` json
{
  "benchmarks": [
    {
      "name": "Single Connection throughput – Upload 100 MiB",
      "unit": "s",
      "results": [
        {
          "result": [
            36.107754525
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "tcp"
        },
        {
          "result": [
            25.158733018
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            3.0486078
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "tcp"
        },
        {
          "result": [
            1.3905061
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            1.12445139
          ],
          "implementation": "zig-libp2p",
          "version": "zig-libp2p-v0.0.1",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            2.864837
          ],
          "implementation": "go-https",
          "version": "go-https-v0.0.1",
          "transportStack": "tcp"
        }
      ],
      "comparisons": []
    },
    {
      "name": "Single Connection throughput – Download 100 MiB",
      "unit": "s",
      "results": [
        {
          "result": [
            35.891990685
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "tcp"
        },
        {
          "result": [
            17.889019488
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            3.0845697
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "tcp"
        },
        {
          "result": [
            1.3131636
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            1.10859072
          ],
          "implementation": "zig-libp2p",
          "version": "zig-libp2p-v0.0.1",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            3.3248453
          ],
          "implementation": "go-https",
          "version": "go-https-v0.0.1",
          "transportStack": "tcp"
        }
      ],
      "comparisons": []
    },
    {
      "name": "Single Connection 1 byte round trip latency",
      "unit": "s",
      "results": [
        {
          "result": [
            0.121320514,
            0.127613735,
            0.128823818,
            0.126729274,
            0.126257158,
            0.120123706,
            0.124415892,
            0.122991401,
            0.119875156,
            0.125521931
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "tcp"
        },
        {
          "result": [
            0.121860941,
            0.125438857,
            0.122788572,
            0.127739452,
            0.119284307,
            0.123368308,
            0.120416946,
            0.124157598,
            0.124542512,
            0.125584374
          ],
          "implementation": "rust-libp2p",
          "version": "rust-master",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            0.06413981,
            0.06411159,
            0.06402883,
            0.064040005,
            0.06408656,
            0.06401954,
            0.064131595,
            0.063996695,
            0.06401078,
            0.06400767
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "tcp"
        },
        {
          "result": [
            0.064188026,
            0.06442312,
            0.06403817,
            0.06405531,
            0.06401896,
            0.06409879,
            0.064039245,
            0.064041995,
            0.06404696,
            0.06405316
          ],
          "implementation": "go-libp2p",
          "version": "go-v0.27.0",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            0.128355666,
            0.0624025724,
            0.0623786114,
            0.0623729676,
            0.062541835,
            0.0623633749,
            0.0623705536,
            0.0623646713,
            0.0623618885,
            0.0623838268
          ],
          "implementation": "zig-libp2p",
          "version": "zig-libp2p-v0.0.1",
          "transportStack": "quic-v1"
        },
        {
          "result": [
            0.1880148,
            0.061682936,
            0.06158356,
            0.0615392,
            0.06163392,
            0.061580177,
            0.061594333,
            0.061534688,
            0.061547145,
            0.061544057
          ],
          "implementation": "go-https",
          "version": "go-https-v0.0.1",
          "transportStack": "tcp"
        }
      ],
      "comparisons": []
    }
  ]
}
```

mxinden:
On the matter of jumbo frames, since we talked about it at IPFS Thing @MarcoPolo and @marten-seemann, the path between the two machines has an MTU of 1500, i.e. does not allow for ping payloads beyond `1472` (IP and ICMP headers sum to 28 bytes):

```
ssh ec2-user@$(terraform output -raw -state ../terraform/terraform.tfstate server_public_ip) ping -M do -s 8972 $(terraform output -raw -state ../terraform/terraform.tfstate client_public_ip) 
PING xxx (xxx) 8972(9000) bytes of data.
From xxx icmp_seq=1 Frag needed and DF set (mtu = 1500)
ping: local error: message too long, mtu=1500
ping: local error: message too long, mtu=1500
ping: local error: message too long, mtu=1500

ssh ec2-user@$(terraform output -raw -state ../terraform/terraform.tfstate server_public_ip) ping -M do -s 1473 $(terraform output -raw -state ../terraform/terraform.tfstate client_public_ip)
PING xxx (xxx) 1473(1501) bytes of data.
ping: local error: message too long, mtu=1500
ping: local error: message too long, mtu=1500
ping: local error: message too long, mtu=1500
ping: local error: message too long, mtu=1500

ssh ec2-user@$(terraform output -raw -state ../terraform/terraform.tfstate server_public_ip) ping -M do -s 1472 $(terraform output -raw -state ../terraform/terraform.tfstate client_public_ip)
PING xxx (xxx) 1472(1500) bytes of data.
1480 bytes from 54.227.46.50: icmp_seq=1 ttl=103 time=61.0 ms
1480 bytes from 54.227.46.50: icmp_seq=2 ttl=103 time=60.9 ms
1480 bytes from 54.227.46.50: icmp_seq=3 ttl=103 time=60.9 ms
1480 bytes from 54.227.46.50: icmp_seq=4 ttl=103 time=60.9 ms
```

perf/impl/https/main.go
```diff
@@ -0,0 +1,241 @@
...
+
+		durations[i] = time.Since(startTime)
+	}
+
+	return durations, nil
+}
+
+func generateEphemeralCertificate() (tls.Certificate, error) {
+	// Generate a private key
+	privKey, err := rsa.GenerateKey(rand.Reader, 2048)
```
marten-seemann, line 99: Please don't use RSA keys.

perf/terraform/region/main.tf
```diff
@@ -0,0 +1,143 @@
...
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  tags = merge(var.common_tags, {
+    Name = "restricted_inbound_sg"
+  })
+}
+
+resource "aws_key_pair" "mxinden" {
```
marten-seemann, line 114: I'm not familiar with Terraform. How does this work? It doesn't seem correct to hardcode a user name here.

perf/dashboard/README.md
```diff
@@ -0,0 +1,62 @@
...
+| download_bytes       | number of bytes to download from server per connection |
+| n_times              | number of times to do the perf round trip              |
+| close_connection     | bool. Close the connection after a perf round trip     |
+
+The binary should report the following:
+
+* Stats on the length of the run:
+  * Total, Avg, Min, Max, p95
+* Round trips per second:
+  * Avg, Min, Max, p95
```
BigLep, line 60: Feedback from [before](https://github.com/libp2p/test-plans/pull/159#discussion_r1148428159):

Rather than special-casing Min / Max, just do p0 and p100 ?

BigLep, line 60: Feedback from [before](https://github.com/libp2p/test-plans/pull/159#discussion_r1148428982):

I'm trying to follow what this means.

I think we probably needs some vocabulary here.
I'm open to different terms, but I'm assuming we have a "test run" which will do "n_times" round trips.

I assume this means that at the end of the "test run" we emit:
avg number of round trips per second that were completed
p0 number of round trips per second
p100 number of round trips per second
p95 number of round trips per second
etc.

This means that if the benchmark runner does a "test run" that lasts for ~10 seconds, it will be tracking for each of those seconds the number of round trips that complete each second so that it can then compute the avg/p0/p95/p100 stats.

Is that right?

Also, I assume the benchmark doesn't care how many concurrent connections there are. It just tracks how many round trips complete each second so it can compute the stats?

perf/dashboard/README.md
```diff
@@ -0,0 +1,62 @@
...
+| -------------------- | ------------------------------------------------------ |
+| parallel_connections | number of parallel connections to use                  |
+| upload_bytes         | number of bytes to upload to server per connection     |
+| download_bytes       | number of bytes to download from server per connection |
+| n_times              | number of times to do the perf round trip              |
+| close_connection     | bool. Close the connection after a perf round trip     |
+
+The binary should report the following:
+
+* Stats on the length of the run:
```
BigLep, line 57:
```suggestion
* Stats on the durations of the `n_times` runs:
```

I assume that is correct (and I think makes it more clear)

BigLep (commented):
I copied a couple of comments over from https://github.com/libp2p/test-plans/pull/159

I think it would be useful to show what command arguments were passed to the binary to generate the output in https://github.com/libp2p/test-plans/pull/163#issuecomment-1509725274

mxinden:
Sorry @BigLep. I did not forget about your comments in https://github.com/libp2p/test-plans/pull/159. I should have made that more explicit before closing.

I don't think it makes sense to review the current state of the pull request. Converted to _Draft_ for now.

perf/impl/https/main.go
```diff
@@ -0,0 +1,208 @@
...
+	return cert, nil
+}
+
+type Latencies struct {
+	Latencies []float32 `json:"latencies"`
+}
+
+func main() {
+	runServer := flag.Bool("run-server", false, "Should run as server")
+	serverIPAddr := flag.String("server-ip-address", "", "Server address")
```
marten-seemann, line 146: Can we not hard-code the port? `IP:port` would be an easy way to be more general.

perf/impl/https/main.go
```diff
@@ -0,0 +1,208 @@
...
+	_ = flag.String("transport", "", "Transport to use")
+	_ = flag.Uint64("secret-key-seed", 0, "Server secret key seed")
+	uploadBytes := flag.Int("upload-bytes", 0, "Upload bytes")
+	downloadBytes := flag.Int("download-bytes", 0, "Download bytes")
```
marten-seemann, line 150: Nit: these should be uint64s.

perf/impl/https/main.go
```diff
@@ -0,0 +1,208 @@
...
+		// Run the client and print the results
+		durations, err := runClient(fmt.Sprintf("%s:%d", *serverIPAddr, 4001), *uploadBytes, *downloadBytes, *nTimes)
+		if err != nil {
+			panic(err)
+		}
+
+		// Convert durations to seconds and marshal as JSON
+		timesS := make([]float32, 0, len(durations))
+		for _, d := range durations {
+			timesS = append(timesS, float32(d.Nanoseconds())/1_000_000_000)
```
marten-seemann, line 194: Why not `d.Seconds()`?

perf/impl/https/main.go
mxinden: What do you need different ports for @marten-seemann?

marten-seemann: I don’t see any reason to hardcode this. I’ve already run into the problem that I want to use a different port (since this is the Kubo default port), and this seems super trivial to make configurable.

mxinden:
@marten-seemann with the latest commit:

- the usage of multiaddr is removed
- binaries are built on the remote machines, thus making MacOS user lifes easier
- targets are run as bare binaries, outside of docker, thus removing another layer of abstraction

This should unblock your work testing quic-go with this setup. I still need to improve documentation and land a couple of refactorings. In case you want to get started already anyways, I would recommend using `perf/impl/https/v0.1` as an example.

(Haven't forgotten about your review above.)

perf/impl/https/main.go
mxinden: Thanks. Done in d1b7d00.

perf/impl/https/main.go
mxinden: Thanks. Fixed in 1f87937.

perf/impl/https/main.go
mxinden: > and this seems super trivial to make configurable.

Not super trivial. Thus far the *Runner* does not have to take any information from the *Server* to the *Client*, e.g. by reading the *Server's* stdout. In case we want to make the port dynamic at runtime, we would need some mechanism from server->runner-> client. None of this is very difficult. Though I will not prioritize it e.g. over proper visualizations unless we have a direct use-case for it.

mxinden: Added to the outstanding list in the pull request description.

marten-seemann: I think we have a misunderstanding here. I wasn't asking for the server picking a random port, and communicating that to the client. I don't think we need that at all.

Instead, I was expecting that the runner passes the port to the server:
```
go run main.go -run-server=true -server-address=0.0.0.0:1234
```

fwiw, that's how I implemented it in quic-go's perf runner: https://github.com/quic-go/perf/blob/master/cmd/main.go

Does that make sense?

perf/dashboard/README.md
mxinden: We no longer aggregate data points in the implementations, nor the runner, but instead provide everything up to the visualization layer, thus giving users full control over how they want to aggregate.

perf/runner/src/index.ts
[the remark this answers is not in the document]
mxinden: Binaries are build and run on the remote machines now. Implementations can choose whether they want to build or run with Docker. All current implementations (rust, go, https) build but don't run inside of Docker. Resolving here.

perf/terraform/region/main.tf
mxinden: Done in c2831cc.

mxinden:
> I think it would be useful to show what command arguments were passed to the binary to generate the output in [#163 (comment)](https://github.com/libp2p/test-plans/pull/163#issuecomment-1509725274)

Added to the list of potential improvements, though not prioritizing it. I expect high-level users to interact with the ObservableHQ dashboard only. Let me know in case you think differently about priority.

perf/dashboard/README.md
mxinden: Things have changed a lot since:

- Each iteration is a single connection only.
- `--n-times` parameter is unused and likely to be removed. In other words, first iteration will only do a single stream per connection.
- No aggregation happening in implementations nor runner. Aggregation is happening at very last, in ObservableHQ, thus giving users more options to aggregate.

Resolving here for now. Let's add more complex benchmarks in future iterations.

mxinden:
With [9ed7f1a](https://github.com/libp2p/test-plans/pull/163/commits/9ed7f1ab07759e4d18618ddbcfed73d0d04cefc9) we are now also testing quic-go via https://github.com/quic-go/perf. Note that there are a couple of minor patches that I still need to upstream to https://github.com/quic-go/perf, see https://github.com/mxinden/perf/tree/irgnore-unknown.

The latest commit also updates the checked-in benchmark results that I ran just now. https://observablehq.com/@mxinden-workspace/libp2p-perf visualizes the latest results automatically.

//CC @marten-seemann

BigLep:
I'm not saying it needs to be tracked in this PR, but for calling this effort "done-done-done" we need to have docs that are easily discoverable where someone can learn what we mean by performance and how they can inspect it.   This is likely some libp2p docs and a blog entry.  (They can certainly point to this readme).

mxinden:
@BigLep yes, tracked in https://github.com/libp2p/test-plans/issues/63.

mxinden:
Rough estimate for a bandwidth delay product between the two machines.

```
[ec2-user@ip-xxx]$ ping  xxx 
PING xxx (xxx) 56(84) bytes of data.
64 bytes from xxx: icmp_seq=1 ttl=97 time=60.5 ms
64 bytes from xxx: icmp_seq=2 ttl=97 time=60.4 ms
64 bytes from xxx: icmp_seq=3 ttl=97 time=60.4 ms
64 bytes from xxx: icmp_seq=4 ttl=97 time=60.4 ms
64 bytes from xxx: icmp_seq=5 ttl=97 time=60.4 ms
64 bytes from xxx: icmp_seq=6 ttl=97 time=60.4 ms
64 bytes from xxx: icmp_seq=7 ttl=97 time=60.4 ms
64 bytes from xxx: icmp_seq=8 ttl=97 time=60.4 ms
64 bytes from xxx: icmp_seq=9 ttl=97 time=60.4 ms
64 bytes from xxx: icmp_seq=10 ttl=97 time=60.5 ms
```
```
[ec2-user@ip-xxx]$ iperf3 -c xxx  -u -b 25g
Connecting to host xxx, port 5201
[  5] local xxx port 55791 connected to xxx port 5201
[ ID] Interval           Transfer     Bitrate         Total Datagrams
[  5]   0.00-1.00   sec   405 MBytes  3.39 Gbits/sec  292936  
[  5]   1.00-2.00   sec   405 MBytes  3.40 Gbits/sec  293077  
[  5]   2.00-3.00   sec   404 MBytes  3.39 Gbits/sec  292818  
[  5]   3.00-4.00   sec   404 MBytes  3.39 Gbits/sec  292730  
[  5]   4.00-5.00   sec   405 MBytes  3.40 Gbits/sec  293246  
[  5]   5.00-6.00   sec   405 MBytes  3.40 Gbits/sec  293318  
[  5]   6.00-7.00   sec   404 MBytes  3.39 Gbits/sec  292567  
[  5]   7.00-8.00   sec   405 MBytes  3.40 Gbits/sec  293088  
[  5]   8.00-9.00   sec   405 MBytes  3.39 Gbits/sec  292946  
[  5]   9.00-10.00  sec   404 MBytes  3.39 Gbits/sec  292876  
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[  5]   0.00-10.00  sec  3.95 GBytes  3.39 Gbits/sec  0.000 ms  0/2929602 (0%)  sender
[  5]   0.00-10.06  sec  3.95 GBytes  3.37 Gbits/sec  0.002 ms  976/2929291 (0.033%)  receiver

iperf Done.
```

mxinden:
With [eff32de](https://github.com/libp2p/test-plans/pull/163/commits/eff32dec610b416a85b94a88af7cec49bad7e179) we now also measure via `ping` and `iperf3` and commit the results to the benchmark JSON.

I have a rough outline of how to use the new data in https://observablehq.com/@mxinden-workspace/libp2p-perf.

BigLep:
@mxinden:

> Move to TLS only (for now). Easier to compare with HTTPs and no additional security protocol negotiation.

Agreed we should have something very comparable to HTTPs, but can we also add other security protocol negotiation so we can see how much that impacts performance?

perf/impl/go-libp2p/v0.27/main.go
```diff
@@ -0,0 +1,127 @@
...
+
+	start := time.Now()
+	h.Peerstore().AddAddrs(serverInfo.ID, serverInfo.Addrs, peerstore.TempAddrTTL)
+	// Use h.Network().DialPeer() instead of h.Connect to skip waiting for
+	// identify protocol to finish.
+	_, err = h.Network().DialPeer(context.Background(), serverInfo.ID)
+	if err != nil {
+		panic(err)
+	}
+	connectionEstablished := time.Since(start)
```
mxinden, line 93: Skipping the identify wait in `Connect`.

perf/impl/go-libp2p/v0.27/perf.go
```diff
@@ -0,0 +1,171 @@
...
+	s, err := ps.Host.Network().NewStream(network.WithNoDial(ctx, "already dialed"), p)
+	if err != nil {
+		return 0, 0, err
+	}
+	s.SetProtocol(ID)
+	lzcon := msmux.NewMSSelect(s, ID)
+	s = &streamWrapper{
+		Stream: s,
+		rw:     lzcon,
+	}
```
mxinden, line 110: Skipping the identify wait in `NewStream`.

mxinden (commented):
@MarcoPolo @marten-seemann as discussed before I added patches to the go-libp2p perf implementation to not wait for the libp2p identify exchange when (1) establishing the connection and (2) opening the perf stream.

There is likely a more go-libp2p idiomatic way of doing this. Feedback welcome.

mxinden:
The above brings us one step closer to reaching https latency performance. The vanilla Go https implementation needs 3RTT, both rust-libp2p and with the above now also go-libp2p with TCP need 4 RTT.

mxinden:
> @mxinden:
> 
> > Move to TLS only (for now). Easier to compare with HTTPs and no additional security protocol negotiation.
> 
> Agreed we should have something very comparable to HTTPs, but can we also add other security protocol negotiation so we can see how much that impacts performance?

We can. I suggest doing so in future iterations. Added to the list of potential next steps in the pull request description.

perf/impl/go-libp2p/v0.27/main.go
mxinden: @marten-seemann @MarcoPolo @sukunrt any idea why the connection establishment with the code above takes 3 network round-trips? I would expect it to take 2 RTT.

My assumption:
- 1 RTT for TCP.
- I only enable TLS, thus no security protocol negotiation, more specifically optimistic 0RTT multistream-select.
- 1 RTT for TLS. (Assuming TLS 1.3 is the default.)
- Using default muxers, which should be only Yamux, thus no muxer negotiation, more specifically optimistic 0RTT multistream-select.
- Not waiting for the additional identify exchange.

mxinden:
I changed the base of the pull request to the `perf` branch on the upstream `libp2p/test-plans` repository. Will ease collaboration with @galargh on https://github.com/mxinden/test-plans/pull/2.

**Let's continue on https://github.com/libp2p/test-plans/pull/184.**

perf/impl/go-libp2p/v0.27/main.go
sukunrt: go-libp2p will spend 1RTT negotiating the security. 
go-libp2p doesn't do 0RTT select for security. It does 0RTT select for protocols when it knows the peer supports the given protocol(mostly via identify) otherwise it'll still do the 1RTT select.

mxinden: > go-libp2p will spend 1RTT negotiating the security protocol.

Given that the client only supports one security protocol, i.e. TLS, shouldn't it be using the lazy mechanism, in other words optimistically use TLS @sukunrt?

https://github.com/multiformats/go-multistream/blob/master/lazyClient.go

sukunrt: I think we can change this. But currently this is not what's happening. 
https://github.com/libp2p/go-libp2p/blob/master/p2p/net/upgrader/upgrader.go#L319-L330
In security negotiation we don't use lazyClient. 
It uses this function which does wait for the reply. https://github.com/multiformats/go-multistream/blob/master/client.go#L117

mxinden: Thank you @sukunrt. I forgot that go-libp2p supports [the multistream-select simultaneous-open extension](https://github.com/libp2p/specs/blob/master/connections/simopen.md) and thus can not do an optimistic multistream-select negotiation when negotiating a single protocol (here TLS) only.

The extension is not needed in libp2p's hole punching as roles (i.e. dialer and listener) are derived from the roles on the relayed connection, see [DCUtR protocol section](https://github.com/libp2p/specs/blob/master/relay/DCUtR.md#the-protocol). As far as I can tell the simultaneous-open extension is only needed for the case where two nodes randomly connect via TCP at the same time.

In my eyes the probability for two nodes to randomly, i.e. not coordinated via DCUtR, connect via TCP **at the same time** is low and only leads to a connection failure. Thus we don't support it in rust-libp2p.

@sukunrt @MarcoPolo @marten-seemann what do you think of removing the simultaneous-open extension in go-libp2p and thus enabling the optimistic security protocol negotiation when negotiation a single protocol only? This would allow go-libp2p to be able to compete with vanilla TCP+TLS+HTTP when it comes to connection establishment latency. In other words both go-lib2p with TLS+Yamux and vanilla TCP+TLS+HTTP would take 3 RTT (2 RTT connection establishment + 1 RTT request/response).

marten-seemann: I don’t think we should make code changes just to look better in benchmarks. In practice, most configurations (IPFS and Lotus) use TLS and Noise.

@vyzo had numbers on simultaneous opens happening in the wild. Would you mind sharing them here?

Menduist: FWIW, in nim-libp2p, we don't bind when we are public, which avoids uncoordinated simultaneous opens in the wild

We don't have enough data yet to see if it causes issues, but so far the only scenario where it doesn't work is if someone made a manual port mapping to a different port than the internal one, and didn't specify it to libp2p. In this case, we can't discover our public port, and we think we are private (since we don't bind)

mxinden: Thank you for the input everyone.

I documented the above as well as past discussions in https://github.com/libp2p/go-libp2p/issues/2330. Let's continue the discussion over there in case there is interest.

PR #166, opened 2023-04-12
fix(interop): rename `webrtc` to `webrtc-direct`

mxinden:
See https://github.com/multiformats/multiaddr/pull/150#issuecomment-1468791586 for context on discussion.

- Renames the `webrtc` transport identifier to `webrtc-direct`.
- Re-enables `webrtc-direct` support for rust-libp2p v0.51. Previously disabled in https://github.com/libp2p/test-plans/pull/160. See https://github.com/libp2p/rust-libp2p/pull/3781 for corresponding change on the rust-libp2p side.
- Leaves JS v0.41 and v0.42 untouched. To be done in a follow-up alongside JS dependency updates.

### DIFF
--- a/multidim-interop/src/generator.ts
+++ b/multidim-interop/src/generator.ts
@@ -61,7 +61,7 @@ export async function buildTestSpecs(versions: Array<Version>): Promise<Array<Co
                      AND ma.muxer == mb.muxer
                      -- quic only uses its own muxer/securechannel
                      AND a.transport != "webtransport"
-                     AND a.transport != "webrtc"
+                     AND a.transport != "webrtc-direct"
                      AND a.transport != "quic"
                      AND a.transport != "quic-v1";`);
     const quicQueryResults =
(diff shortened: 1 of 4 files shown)


thomaseizinger (approved):

multidim-interop/src/generator.ts
```diff
@@ -61,7 +61,7 @@ export async function buildTestSpecs(versions: Array<Version>): Promise<Array<Co
                      AND ma.muxer == mb.muxer
                      -- quic only uses its own muxer/securechannel
                      AND a.transport != "webtransport"
-                     AND a.transport != "webrtc"
```
MarcoPolo, line 64: we want both here, right? but not needed right now

MarcoPolo (approved):

multidim-interop/src/generator.ts
MarcoPolo: We'll add this when we add webrtc tests here :)

PR #181, opened 2023-05-15
feat: update Rust tests to use smaller docker container

thomaseizinger:
This is the follow-up PR to https://github.com/libp2p/rust-libp2p/issues/3881. We update the hash to the merge commits of the 3 backport PRs for these versions:

- https://github.com/libp2p/rust-libp2p/pull/3943
- https://github.com/libp2p/rust-libp2p/pull/3944
- https://github.com/libp2p/rust-libp2p/pull/3945

With this patch, the final docker image that needs to be downloaded is only 50MB.

### DIFF
--- a/multidim-interop/impl/rust/v0.49/Makefile
+++ b/multidim-interop/impl/rust/v0.49/Makefile
@@ -1,5 +1,5 @@
 image_name := rust-v0.49
-commitSha := 582c84043050eb0dd6e52d1bd0527175181d41cb
+commitSha := 45e3f9c61ad6de18e3e28ca2810a65668d6bba8d
 
 all: image.json
 
--- a/multidim-interop/impl/rust/v0.50/Makefile
+++ b/multidim-interop/impl/rust/v0.50/Makefile
@@ -1,5 +1,5 @@
 image_name := rust-v0.50
-commitSha := beb66b5832384d9b813fcbf1f0fa01e6a64a9c5f
+commitSha := 97071b9a8e0dbd2d2526400169eaa30f317e9d62
 
 all: image.json
 
--- a/multidim-interop/impl/rust/v0.51/Makefile
+++ b/multidim-interop/impl/rust/v0.51/Makefile
@@ -1,5 +1,5 @@
 image_name := rust-v0.51
-commitSha := 0d5cac0cb595702567e50221fed9ae525b4c6f20
+commitSha := 8c26c617b94d992c5fbd91efc73049c838740df1
 
 all: image.json
 


MarcoPolo (approved):

thomaseizinger:
I am rerunning the tests to get some numbers on the improvement. I don't expect them to be huge because GitHub normally has pretty solid network connections but it is interesting nonetheless.

mxinden (approved):
Thanks Thomas!

thomaseizinger:
At the moment, this is 10min slower than runs on master.

Looking at the logs, we don't seem to be using a cachs for the new images, not even on the 2nd run. Is that expected @MarcoPolo ? Shouldn't we save the cache after the first run?

MarcoPolo:
> Looking at the logs, we don't seem to be using a cache for the new images, not even on the 2nd run. Is that expected @MarcoPolo ? Shouldn't we save the cache after the first run?

This branch is on your fork so you aren’t running CI with the same enviornment. Specifically the secrets that allow CI to write to the the cache.

In the future, we should use a branch so that it pushes to the cache. For now, we can just go ahead and merge this.

thomaseizinger:
We can see from here onwards that download the new Rust (> 0.49) images takes only about 1 second whereas the old[^1] one (0.48) takes about 20: https://github.com/libp2p/test-plans/actions/runs/4993522255/jobs/8946282869#step:3:612

[^1]: I didn't port 0.48 because we are going to remove it soon.

MarcoPolo:
Nice :)

MarcoPolo:
This probably does make the tests run faster as well since extracting the layer and mounting the image is faster.

thomaseizinger:
Doesn't seem to be the case but looking at previous runs we don't seem to have very reliable numbers anyway: https://github.com/libp2p/test-plans/actions/workflows/multidim-interop.yml?query=branch%3Amaster

MarcoPolo:
On my machine testing rust-libp2p v0.49 before and after:

small image 13M
```
❮ time docker run --rm -it sha256:37f6afd68ed35d937b7f90781983caab7b7f26623e69cef3899da71e0086b78f
//...
docker run --rm -it   0.05s user 0.03s system 18% cpu 0.407 total
```

big image 532M
```
time docker run --rm -it  sha256:e3222b68a9f5319d22892558607cc71197875678a1eb6134e2db38107333685b
//...
docker run --rm -it   0.05s user 0.03s system 16% cpu 0.509 total
```

Run several times to have a warm FS cache. 100ms difference. Yeah pretty small.

What would really help speed all this up is if we could start these containers instantly instead of taking 400ms to start up. Even the `hello-world` container for me takes 300 ms to startup. There must be some strategies to speed this up, since I wouldn't expect serverless compute to be okay with a 300ms startup latency.

Issue #183, opened 2023-05-25
Automate performance benchmarking


mxinden:
Overarching tracking issue: https://github.com/libp2p/test-plans/issues/63

https://github.com/libp2p/test-plans/pull/163/ allows running the various perf protocol implementations (https://github.com/libp2p/specs/pull/478) on cloud instances.

Ideally we have a CI setup, that allows users to trigger the above. This removes the need for users to run it on their local machine.

### Workflow

Starting point: libp2p implementation maintainer wants to test a new release candidate.

3. They open a pull request to `libp2p/test-plans`, adding their release candidate to `perf/impl`.
4. They trigger the automation via a GitHub comment.
5. CI runs:
    1.  Generate an ssh keypair and set the public key at `perf/terraform/user.pub`.
    2. Run `cd perf/terraform && terraform apply`.
    3. Run `cd perf/runner && npm run start -- --client-public-ip $(terraform output -raw -state ../terraform/terraform.tfstate client_public_ip) --server-public-ip $(terraform output -raw -state ../terraform/terraform.tfstate server_public_ip)`. This will write the benchmark results to `perf/runner/benchmark-results.json`.
    4. Run `cd perf/terraform && terraform destroy`.
    6. Push `perf/runner/benchmark-results.json` to the pull request.
5. Depending on the benchmark results:
    - In case the automation catches a regression, the maintainer can cut another patch release and update the `libp2p/test-plans` pull request and retrigger the automation.
    - In case the results are fine, the maintainer can cut the release. Once released, they update the `libp2p/test-plans` pull request, retrigger the automation and merge. Thus `libp2p/testplans` `master` branch always contains the latest benchmarking results.

mxinden:
//CC @galargh I extracted the automation section from https://github.com/libp2p/test-plans/pull/163/ to this new issue.

galargh:
Thanks :) Here's the first pass at the automation - https://github.com/mxinden/test-plans/pull/2

mxinden:
First successful automation run on `libp2p/test-plans` happened today. See https://github.com/libp2p/test-plans/issues/63#issuecomment-1603683902. Thus merging here.

Thank you @galargh!

PR #184, opened 2023-05-26
feat(perf): add (automation, provision, build, run) tooling

mxinden:
### Description

> # libp2p performance benchmarking
> 
> This project includes the following components:
> 
> - `terraform/`: a Terraform scripts to provision infrastructure
> - `impl/`: implementations of the [libp2p perf protocol](https://github.com/libp2p/specs/pull/478) running on top of e.g. go-libp2p, rust-libp2p or Go's std-library https stack
> - `runner/`: a set of scripts building and running the above implementations on the above infrastructure, reporting the results in `benchmark-results.json`
> 
> Benchmark results can be visualized with https://observablehq.com/@mxinden-workspace/libp2p-performance-dashboard.

See `perf/README.md` and overarching tracking issue https://github.com/libp2p/test-plans/issues/63.

Continuation of https://github.com/libp2p/test-plans/pull/163.

### Outstanding work in this pull request:

- [x] Remove libp2p specific concepts (e.g. multiaddr) e.g. to no longer require the https example to parse a multiaddr.
- [x] Rename public ssh key
- [x] Remove `--n-times` flag from binaries. Potentially introduce `--n-parallel-requests` flag in the future. Multiple _sequential_ runs are now done in the runner implementation.
- [x] Transfer and run binaries instead of docker images / containers. Or build on the machines.
- [x] Configurable Server port. See https://github.com/libp2p/test-plans/pull/163/#discussion_r1183459190.
- [x] Consider adding benchmark parameters to output json file. See https://github.com/libp2p/test-plans/pull/163#pullrequestreview-1382950014.
- [x] Move to TLS only (for now). Easier to compare with HTTPs and no additional security protocol negotiation.
- [x] Remove the need for static peer IDs, i.e. remove `--secret-key-seed`.

### Outstanding work for future pull requests:
- [ ] Only use a single seconds measurement. Don't split in connection establishment, upload and download. These are difficult to differentiate.
- [ ] Add js-libp2p https://github.com/libp2p/js-libp2p/pull/1604
- [ ] Add back zig-libp2p
- [ ] Run iperf on both UDP *and* TCP
- [ ] Add vanilla Go http in order to measure overhead of TLS
- [ ] Support testing other security protocols, i.e. Noise
- [ ] Resource utilization benchmark?
- [ ] Automation see https://github.com/libp2p/test-plans/issues/183
- [ ] Don't blindly trust server and clients ssh key (mitm attack)
- [ ] ... see tracking issue https://github.com/libp2p/test-plans/issues/63


mxinden:
This pull request is ready for technical review. I would appreciate input from everyone interested in the libp2p perf work.

mxinden:
> This pull request is ready for technical review. I would appreciate input from everyone interested in the libp2p perf work.

Friendly ping @marten-seemann @MarcoPolo @sukunrt and @thomaseizinger.

Unless there are any objections I will proceed here and merge sometime early next week. I am happy to do follow-ups after the merge. Also, in case anyone raises strong objections after the merge, we can still revert.

marten-seemann:
I was planning to review, but this is a lot of code.

mxinden:
> I was planning to review, but this is a lot of code.

True. Big project at this point. Let me know if there is something I can do to help you review @marten-seemann. Also let me know if you want me to block merging on your review.

MarcoPolo:
I'd like to review as well before merging.

sukunrt:
I'll have a look this week. I have no issues with merging this before my review.

perf/terraform/modules/long_lived/main.tf
```diff
@@ -0,0 +1,111 @@
...
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+}
+
+resource "aws_launch_template" "perf" {
+  name          = "perf-node"
+  image_id      = var.ami
+  instance_type = "m5n.8xlarge"
```
mxinden, line 97: Currently this pull request uses [`m5n.8xlarge` machines with 25 Gibps](https://instances.vantage.sh/aws/ec2/m5n.8xlarge?selected=m5n.8xlarge&region=us-east-1&os=linux&cost_duration=hourly&reserved_term=Standard.noUpfront).

A couple of considerations:

- AWS offers either exactly XX or "up to" XX network performance. For what this guarantee is worth, I suggest, for the sake of reproducibility to choose the exactly XX types. (In case someone has more expertise here, please speak up.)
- No matter what instance type we choose, it will not be representative of the most libp2p deployments. At this point libp2p covers the whole space from a small laptop to huge bare metal machines.
- Thus far I have chosen an instance type such that the underlying hardware is not the bottleneck of the benchmark. In other words that the bottleneck shows up to be in the libp2p implementations itself.
- When it comes to network bandwidth this is a double-edged sword. I.e. high bandwidth requires large send and receive buffers. Large send and receive buffers make us vulnerable to DOS attacks.
- We could as well use the [general purpose `m5n.8xlarge` (note the missing `n`) instance type](https://instances.vantage.sh/aws/ec2/m5a.8xlarge). This would give us 10 Gibps instead of 25 Gibps at a slightly lower price.

With the above in mind, I opted to keep the instance type (`m5n.8xlarge`) as is.

perf/runner/src/index.ts
```diff
@@ -0,0 +1,238 @@
...
+    // Save results to benchmark-results.json
+    fs.writeFileSync('./benchmark-results.json', JSON.stringify(benchmarkResults, null, 2));
+
+    console.error("== done");
+}
+
+function runPing(clientPublicIP: string, serverPublicIP: string): PingResults {
+    console.error(`= run 100 pings from client to server`);
+
+    let cmd = `ssh -o StrictHostKeyChecking=no ec2-user@${clientPublicIP} 'ping -c 100 ${serverPublicIP}'`;
```
marten-seemann, line 59: Not a priority, but it would be nice if we didn't make ourselves vulnerable to MITM attacks here. Is there any (easy) way to get the SSH fingerprint? I imagine this is a very common problem that people have when they programmatically boot up servers.

marten-seemann (commented):
@mxinden took the time to walk me through the design earlier today. Thank you for that.

This is fantastic work! I'm super happy how clean this turned out. I especially like how the recursive Makefiles contain all the build logic to the respective implementation.

I reviewed the high-level design and the Go code. I won't be able to give meaningful feedback on the Terraform setup, nor on the Rust code.

I've added quite a few comments here, most of them minor suggestions how to improve the Go code.

Design-wise, here's what I suggest changing:
1. We should probably run `iperf` on TCP and on UDP. It's far from guaranteed that we're getting the same result here.
2. We should benchmark both TLS and Noise. The majority of TCP traffic on the IPFS and the Ethereum network uses Noise, so this is probably the more relevant metric for users. Plus, being able to compare TLS and Noise will likely inform future users' decision which one(s) to use.
3. The Go code currently "cheats" to achieve good results, by using code paths that no user would use. See https://github.com/libp2p/test-plans/pull/184#discussion_r1219324445.

marten-seemann (commented):
@mxinden took the time to walk me through the design earlier today. Thank you for that.

This is fantastic work! I'm super happy how clean this turned out. I especially like how the recursive Makefiles contain all the build logic to the respective implementation.

I reviewed the high-level design and the Go code. I won't be able to give meaningful feedback on the Terraform setup, nor on the Rust code.

I've added quite a few comments here, most of them minor suggestions how to improve the Go code.

Design-wise, here's what I suggest changing:
1. We should probably run `iperf` on TCP and on UDP. It's far from guaranteed that we're getting the same result here.
2. We should benchmark both TLS and Noise. The majority of TCP traffic on the IPFS and the Ethereum network uses Noise, so this is probably the more relevant metric for users. Plus, being able to compare TLS and Noise will likely inform future users' decision which one(s) to use.
3. The Go code currently "cheats" to achieve good results, by using code paths that no user would use.

perf/impl/go-libp2p/v0.27/main.go
```diff
@@ -0,0 +1,143 @@
...
+
+func main() {
+	runServer := flag.Bool("run-server", false, "Should run as server")
+	serverAddr := flag.String("server-address", "", "Server address")
+	transport := flag.String("transport", "tcp", "Transport to use")
+	uploadBytes := flag.Uint64("upload-bytes", 0, "Upload bytes")
+	downloadBytes := flag.Uint64("download-bytes", 0, "Download bytes")
+	flag.Parse()
+
+	ipPort := strings.Split(*serverAddr, ":")
```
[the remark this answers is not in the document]
mxinden: :pray: 5f1a4ba

perf/impl/go-libp2p/v0.27/main.go
```diff
@@ -0,0 +1,143 @@
...
+	if len(ipPort) != 2 {
+		fmt.Println("Invalid server address format. Expected format: 'ip:port'")
```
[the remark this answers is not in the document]
mxinden: :+1: 5f1a4ba

perf/impl/go-libp2p/v0.27/main.go
```diff
@@ -0,0 +1,143 @@
...
+		libp2p.DefaultPrometheusRegisterer,
+	}
+
+	if *runServer {
+		opts = append(opts, libp2p.ListenAddrStrings(tcpMultiAddrStr, quicMultiAddrStr))
+
+		// Generate fake identity.
+		priv, _, err := crypto.GenerateEd25519Key(&simpleReader{seed: 0})
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
mxinden: Thanks. See e5cb56b.

perf/impl/go-libp2p/v0.27/main.go
```diff
@@ -0,0 +1,143 @@
...
+		libp2p.DefaultConnectionManager,
+		libp2p.DefaultMultiaddrResolver,
+		libp2p.DefaultPrometheusRegisterer,
+	}
+
+	if *runServer {
+		opts = append(opts, libp2p.ListenAddrStrings(tcpMultiAddrStr, quicMultiAddrStr))
+
+		// Generate fake identity.
+		priv, _, err := crypto.GenerateEd25519Key(&simpleReader{seed: 0})
```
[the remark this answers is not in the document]
mxinden: Added documentation in e28351d

perf/impl/go-libp2p/v0.27/perf.go
```diff
@@ -0,0 +1,171 @@
...
+	err = p.sendBytes(context.Background(), s, bytesToSend, buf)
+	if err != nil {
+		log.Errorw("err", err)
+		s.Reset()
+		return
+	}
+
+}
+
+func (ps *PerfService) sendBytes(ctx context.Context, s network.Stream, bytesToSend uint64, buf []byte) error {
```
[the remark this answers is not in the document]
mxinden: Fine by me. Done in a9cd2a2.

perf/impl/go-libp2p/v0.27/perf.go
```diff
@@ -0,0 +1,171 @@
...
+			return err
+		}
+		bytesToSend -= uint64(n)
+	}
+	s.CloseWrite()
+
+	return nil
+}
+
+func (ps *PerfService) drainStream(ctx context.Context, s network.Stream, buf []byte) (uint64, error) {
```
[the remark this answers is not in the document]
mxinden: Cool! a9cd2a2.

perf/impl/go-libp2p/v0.27/perf.go
```diff
@@ -0,0 +1,171 @@
...
+			s.Reset()
+			return recvd, err
+		}
+	}
+}
+
+func (ps *PerfService) RunPerf(ctx context.Context, p peer.ID, bytesToSend uint64, bytesToRecv uint64) (time.Duration, time.Duration, error) {
+	// Use ps.Host.Network().NewStream() instead of ps.Host.NewStream() to
+	// skip waiting for identify protocol to finish.
+	s, err := ps.Host.Network().NewStream(network.WithNoDial(ctx, "already dialed"), p)
```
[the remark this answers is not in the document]
mxinden: Fine by me. I did the following reverts, thus negotiating both Noise and TLS (no additional roundtrip due to https://github.com/libp2p/go-libp2p/issues/2330 preventing optimistic negotiation) and waiting for the identify exchange on connection establishment and stream creation. Now, it works just like a usual Kubo setup.

- 13e244b perf Revert "fix(go-libp2p): use tls only"
- a8ded97 Revert "fix(go-libp2p): use NewWithoutDefaults to disable Noise"
- 88b6f34 Revert "fix(go-libp2p): don't wait for identify round trip"
- 6e0fc1d Revert "fix(go-libp2p): don't wait for identify when dialing"

//CC @MarcoPolo since we discussed not waiting on identify here.

> I'd argue that we should do the latter, and take this as an opportunity to change the default behavior (which I've wanted to do for a long time). Maybe putting actual numbers on it, and being to claim credit for an improvement in our new metrics page, will make prioritizing this fix more attractive :)

:+1: That makes sense to me @marten-seemann.

perf/runner/src/index.ts
mxinden: Added to "Outstanding work for future pull requests:". Agree that this is not a priority.

perf/impl/go-libp2p/v0.27/main.go
```diff
@@ -0,0 +1,143 @@
...
+
+		libp2p.DefaultListenAddrs,
+		libp2p.Transport(tcp.NewTCPTransport),
+		libp2p.Transport(quic.NewTransport),
+		libp2p.DefaultMuxers,
+		libp2p.DefaultPeerstore,
+		libp2p.DefaultResourceManager,
+		libp2p.DefaultConnectionManager,
+		libp2p.DefaultMultiaddrResolver,
+		libp2p.DefaultPrometheusRegisterer,
```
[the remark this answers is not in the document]
mxinden: As far as I can tell, given that I use `NewWithoutDefaults`, the above is necessary.

Either way, I removed the "optimization" of using TLS only, thus I can just depend on the defaults and use `libp2p.New`.

See https://github.com/libp2p/test-plans/pull/184#discussion_r1222456427.

perf/impl/go-libp2p/v0.27/main.go
```diff
@@ -0,0 +1,143 @@
...
+
+		// Generate fake identity.
+		priv, _, err := crypto.GenerateEd25519Key(&simpleReader{seed: 0})
+		if err != nil {
+			panic(err)
+		}
+		opts = append(opts, libp2p.Identity(priv))
+	} else {
+		opts = append(opts, libp2p.RandomIdentity)
+	}
```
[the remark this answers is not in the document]
mxinden: Resolved with https://github.com/libp2p/test-plans/pull/184#discussion_r1222456427.

perf/impl/go-libp2p/v0.27/perf.go
```diff
@@ -0,0 +1,171 @@
...
+	recvDuration := time.Since(recvStart)
+
+	if recvd != bytesToRecv {
+		return sendDuration, recvDuration, fmt.Errorf("expected to recv %d bytes, got %d", bytesToRecv, recvd)
+	}
+
+	return sendDuration, recvDuration, nil
+}
+
+type streamWrapper struct {
```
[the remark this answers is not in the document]
mxinden: Resolved with https://github.com/libp2p/test-plans/pull/184#discussion_r1222456427.

mxinden:
> I've added quite a few comments here, most of them minor suggestions how to improve the Go code.

Thank you. All comments addressed with the recent commits. Mind taking another look @marten-seemann?

> We should probably run `iperf` on TCP and on UDP. It's far from guaranteed that we're getting the same result here.

Good point. I don't think we need to block this pull request on it. Added to the list of follow-up items. Let me know what you think.

> We should benchmark both TLS and Noise. The majority of TCP traffic on the IPFS and the Ethereum network uses Noise, so this is probably the more relevant metric for users. Plus, being able to compare TLS and Noise will likely inform future users' decision which one(s) to use.

Yes. This is on the list for follow-up pull requests. I am especially curious how they compare on the bandwidth benchmark.

> The Go code currently "cheats" to achieve good results, by using code paths that no user would use.

Removed with https://github.com/libp2p/test-plans/pull/184#discussion_r1222456427.

thomaseizinger (approved):
Great work! Added some minor comments. Can't really comment on the terraform stuff.

mxinden:
Friendly ping @marten-seemann. Any additional follow-ups you would like to see?

Friendly ping @MarcoPolo. Can you foresee when you have time to review?

> > > We should probably run `iperf` on TCP and on UDP. It's far from guaranteed that we're getting the same result here.
> > 
> > 
> > Good point. I don't think we need to block this pull request on it. Added to the list of follow-up items. Let me know what you think.
> 
> Where's that list? I think this should be resolved rather sooner than later (let's say within the next 2 weeks). Don't care if it's in this PR, as long as it makes it in soon.

See pull request description @marten-seemann.

marten-seemann:
> > > > We should probably run `iperf` on TCP and on UDP. It's far from guaranteed that we're getting the same result here.
> > > 
> > > 
> > > Good point. I don't think we need to block this pull request on it. Added to the list of follow-up items. Let me know what you think.
> > 
> > 
> > Where's that list? I think this should be resolved rather sooner than later (let's say within the next 2 weeks). Don't care if it's in this PR, as long as it makes it in soon.
> 
> See pull request description @marten-seemann.

Missed that. Would probably make sense to move to a new issue, so it doesn't get lost when this PR is merged.

perf/impl/go-libp2p/v0.27/main.go
```diff
@@ -0,0 +1,119 @@
...
+	case "tcp":
+		multiAddrStr = tcpMultiAddrStr
+	case "quic-v1":
+		multiAddrStr = quicMultiAddrStr
+	default:
+		fmt.Println("Invalid transport. Accepted values: 'tcp' or 'quic-v1'")
+		return
+	}
+	// Peer ID corresponds to the above fake identity.
+	multiAddrStr = multiAddrStr + "/p2p/12D3KooWDpJ7As7BWAwRMfu1VU2WCqNjvq387JEYKDBj4kx6nXTN"
```
MarcoPolo, line 74: I think this should be passed in as part of the multiaddr for the server, I don't see the benefit of hard coding this identity. We are already printing it anyways. And then you can remove all the code around generating keys from a seed.

perf/impl/go-libp2p/v0.27/perf.go
```diff
@@ -0,0 +1,126 @@
...
+		bytesToSend -= uint64(n)
+	}
+	s.CloseWrite()
+
+	return nil
+}
+
+func (ps *PerfService) drainStream(ctx context.Context, s network.Stream) (uint64, error) {
+	var recvd int64
+	recvd, err := io.Copy(io.Discard, s)
```
MarcoPolo, line 85: nit: Using `io.CopyBuffer` with a 64k buffer might be nicer than allocating one everytime.

perf/impl/go-libp2p/v0.27/perf.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		log.Errorw("err", err)
+		s.Reset()
+		return
+	}
+
+}
+
+func (ps *PerfService) sendBytes(ctx context.Context, s network.Stream, bytesToSend uint64) error {
+	buf := pool.Get(BlockSize)
```
MarcoPolo, line 63: nit: You can stack allocate this buf in the `RunPerf` function and use it instead of needing a sync.Pool.

perf/impl/https/v0.1/main.go
```diff
@@ -0,0 +1,199 @@
...
+		http.Error(w, "failed to read uint64 value", http.StatusBadRequest)
+		return
+	}
+
+	// Read and discard the remaining bytes in the request
+	io.Copy(io.Discard, r.Body)
+
+	buf := make([]byte, BlockSize)
+
+	for bytesToSend > 0 {
```
MarcoPolo, line 42: Please set the `content-type` and `content-length` headers and then call `WriteHeader(OK)` such that the Go stdlib doesn't do extra work here (https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/net/http/server.go;drc=ea4631cc0cf301c824bd665a7980c13289ab5c9d;l=120).

perf/impl/https/v0.1/main.go
```diff
@@ -0,0 +1,199 @@
...
+func runClient(serverAddr string, uploadBytes, downloadBytes uint64) (time.Duration, time.Duration, error) {
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		},
+	}
+
+	reqBody := make([]byte, 8+uploadBytes)
```
MarcoPolo, line 66: ? You're allocating all the bytes at once here? We could be uploading gigs. Please make a new reader that will read until `uploadBytes` have been read.

perf/impl/https/v0.1/main.go
```diff
@@ -0,0 +1,199 @@
...
+	binary.BigEndian.PutUint64(reqBody, uint64(downloadBytes))
+
+	startTime := time.Now()
+	resp, err := client.Post(fmt.Sprintf("https://%s/", serverAddr), "application/octet-stream", bytes.NewReader(reqBody))
+	if err != nil {
+		return 0, 0, err
+	}
+	uploadDoneTime := time.Now()
```
MarcoPolo, line 74: Do you know if this is actually the time when the bytes have been flushed?

MarcoPolo: I think it's worth just measuring the whole thing from start to finish rather than trying to measure the parts. It'll be more accurate since we know for sure when we're done versus when we've finished the upload.

perf/README.md
```diff
@@ -0,0 +1,54 @@
...
+      - Input via command line
+        - `--server-ip-address`
+        - `--transport` (see `runner/versions.ts` for possible variants)
+        - `--upload-bytes` number of bytes to upload per stream.
+        - `--download-bytes` number of bytes to download per stream.
+      - Output
+        - Logging MUST go to stderr.
+        - Measurement output is printed to stdout as JSON in the form of:
+          ```json
+          {"connectionEstablishedSeconds":0.246442851,"uploadSeconds":0.000002077,"downloadSeconds":0.060712241}
```
MarcoPolo, line 51: I think there should just be a single number "seconds" which will be more accurate than trying to figure out exactly when the upload finished. We won't really know when the upload finished, but we will know for sure when the full test finishes. If we want to measure upload we simply don't download anything.

perf/impl/https/v0.1/main.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	defer resp.Body.Close()
+	if uint64(len(respBody)) != downloadBytes {
+		return 0, 0, fmt.Errorf("expected %d bytes in response, but received %d\n", downloadBytes, len(respBody))
+	}
+
+	return uploadDoneTime.Sub(startTime), time.Since(uploadDoneTime), nil
+}
+
+func generateEphemeralCertificate() (tls.Certificate, error) {
```
MarcoPolo, line 88: I'm not reviewing this function. I assume if it works it's good enough.

perf/impl/https/v0.1/main.go
```diff
@@ -0,0 +1,199 @@
...
+				InsecureSkipVerify: true,
+			},
+		},
+	}
+
+	reqBody := make([]byte, 8+uploadBytes)
+	binary.BigEndian.PutUint64(reqBody, uint64(downloadBytes))
+
+	startTime := time.Now()
+	resp, err := client.Post(fmt.Sprintf("https://%s/", serverAddr), "application/octet-stream", bytes.NewReader(reqBody))
```
MarcoPolo, line 70: Check response is OK?

perf/README.md
```diff
@@ -0,0 +1,54 @@
...
+1. `cd runner`
+2. `npm ci`
+3. `npm run start -- --client-public-ip $CLIENT_IP --server-public-ip $SERVER_IP`
+
+## Deprovision infrastructure
+
+1. `cd terraform/configs/local`
+2. `terraform destroy`
+
+## Adding a new implementation
```
MarcoPolo, line 35: Missing building/deploying information.

MarcoPolo (commented):
I didn't review everything. But overall this looks good. Nice work! Just a couple comments of mostly nits.

The biggest concern from me (which we can handle in a future PR) is to only use a single seconds measurement vs trying to separate this out between download and upload.

The problem is that you won't really know when the bytes have been uploaded. Even if they are given to the kernel they still haven't been sent. And even then they may be dropped and need to be retransmitted. The only true measurement we can make is when the test ends.

perf/impl/go-libp2p/v0.27/main.go
mxinden: > We are already printing it anyways.

I am not sure which place you are referring to.

Today there is no way for a perf `impl/` server deployment to pass information (e.g. its peer ID) to its perf `impl/` client deployment counterpart.

We can add a mechanism, e.g. reading the first stdout line of the server or reading the identity from a file.

Given that we are benchmarking performance here and not testing the crypto handshakes, I think this hack is fine, i.e. I don't think it is worth it adding the additional mechanism. In case you think differently @MarcoPolo I suggest doing this as a follow-up.

> I think this should be passed in as part of the multiaddr for the server

Note that the `runner/` is no longer aware of the concept of a multiaddr. It only provides IP and port to the server and client. Otherwise we would need to partse multiaddr in `impl/quic-go` and `impl/https`.

perf/impl/go-libp2p/v0.27/perf.go
marten-seemann: I don't think it matters. It would even be fine allocating this in this function directly. This is a single 64k allocation per run of the perf protocol. Clarity over micro-optimization.

perf/impl/https/v0.1/main.go
marten-seemann: Agree, setting the content length seems reasonable.

perf/impl/https/v0.1/main.go
marten-seemann: That applies to the non-HTTP version as well. The transport may buffer the data before flushing it (quic-go does, albeit just for a single MTU worth of data, so it won't really matter).

.github/workflows/perf.yml
```diff
@@ -0,0 +1,101 @@
+name: libp2p perf test
+
+# How to configure a repository for running this workflow:
+# 1. Configure auth for the AWS provider as per https://registry.terraform.io/providers/hashicorp/aws/latest/docs#authentication-and-configuration
+# 2. Run 'terraform init' and 'terraform apply' in 'perf/terraform/configs/remote' to create the resources needed for this workflow
+# 3. Go to https://console.aws.amazon.com/iamv2/home?#/users/details/perf?section=security_credentials
+# 4. Click 'Create access key' to get the access key ID and secret access key
+# 5. Go to https://github.com/libp2p/test-plans/settings/secrets/actions
+# 6. Click 'New repository secret', set the name to 'PERF_AWS_SECRET_ACCESS_KEY', and paste the secret access key from step 5
+# 7. Go to https://github.com/libp2p/test-plans/settings/variables/actions
```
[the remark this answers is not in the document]
mxinden: For now this workflow will only be run from the `libp2p/test-plans` repository. For every release you would open a pull request against `libp2p/test-plans` and run the workflow from there. In the future we can automate this.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,101 @@
...
+# 8. Click 'New repository variable', set the name to 'PERF_AWS_ACCESS_KEY_ID', and paste the access key ID from step 5
+
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
mxinden: Yes, ideally that would be the case. Though I would like to defer this work to future iterations. Let's get this in a useful state first, actually use it, and then iterate on improvements.

mxinden:
> > > > > We should probably run `iperf` on TCP and on UDP. It's far from guaranteed that we're getting the same result here.
> > > > 
> > > > 
> > > > Good point. I don't think we need to block this pull request on it. Added to the list of follow-up items. Let me know what you think.
> > > 
> > > 
> > > Where's that list? I think this should be resolved rather sooner than later (let's say within the next 2 weeks). Don't care if it's in this PR, as long as it makes it in soon.
> > 
> > 
> > See pull request description @marten-seemann.
> 
> Missed that. Would probably make sense to move to a new issue, so it doesn't get lost when this PR is merged.

I will consolidate the outstanding follow-ups once this is merged on https://github.com/libp2p/test-plans/issues/63.

perf/impl/rust-libp2p/v0.52/Makefile
```diff
@@ -0,0 +1,21 @@
...
+	unzip -o rust-libp2p-${commitSha}.zip
+
+rust-libp2p-${commitSha}.zip:
+	wget -O $@ "https://github.com/libp2p/rust-libp2p/archive/${commitSha}.zip"
+
+clean:
+	rm rust-libp2p-*.zip
+	rm -rf rust-libp2p-*
+
+.PHONY: all clean run
```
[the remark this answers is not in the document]
mxinden: :+1: done with 3b9fafa.

perf/runner/src/index.ts
```diff
@@ -0,0 +1,238 @@
...
+        })
+        .filter((time): time is number => time !== null); // Remove any null values and ensure that array contains only numbers
+
+    return { unit: "s", results: times }
+}
+
+function runIPerf(clientPublicIP: string, serverPublicIP: string): IperfResults {
+    console.error(`= run 60 iPerf UDP from client to server`);
+
+    let killCMD = `ssh -o StrictHostKeyChecking=no ec2-user@${serverPublicIP} 'kill $(cat pidfile); rm pidfile; rm server.log || true'`;
```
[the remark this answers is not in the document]
mxinden: :+1: d13d054

perf/impl/https/v0.1/main.go
mxinden: Good idea. Thanks @MarcoPolo. Done for the client and server in c0c34f4.

perf/impl/https/v0.1/main.go
mxinden: Good catch! Fixed with c0c34f4. Note that I am using a custom `io.Reader`. Let me know in case there is a more idiomatic way using Go's http stack.

perf/impl/https/v0.1/main.go
mxinden: :+1: 3c1a909

perf/README.md
mxinden: Added in 74fdac6.

mxinden:
> The biggest concern from me (which we can handle in a future PR) is to only use a single seconds measurement vs trying to separate this out between download and upload.

Sounds good to me. Added to the top of follow-ups to this pull request.

---

@MarcoPolo thank you for the review! I addressed all your comments. Would you mind taking another look? Anything else missing?

perf/impl/https/v0.1/main.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+}
+
+var zeroSlice = make([]byte, BlockSize) // Pre-made zero-filled slice
+
+type customReader struct {
+	downloadBytes uint64
+	uploadBytes   uint64
+	position      uint64
+}
```
marten-seemann, line 69: Can we just copy-paste the perf implementation we have in the v0.27 folder? The `Read` implementation here looks quite complicated.

mxinden: The `impl/go-libp2p/v0.27` implementation leverages the `io.Writer` interface on `network.Stream` interface. I don't know how to have the Go std http library provide an object that implements `io.Writer` interface for an HTTP request. Instead all the methods that I am aware of require passing an `io.Reader` for an outgoing HTTP request, i.e. `client.Post` and `http.NewRequest`.

@marten-seemann are you aware of an alternative?

perf/impl/https/v0.1/main.go
```diff
@@ -0,0 +1,252 @@
...
+
+func (c *customReader) Read(p []byte) (int, error) {
+	if c.position < 8 {
+		binary.BigEndian.PutUint64(p, c.downloadBytes)
+		c.position += 8
+		return 8, nil
```
MarcoPolo, line 75: This isn't correct. You're assuming the input slice is always at least 8 bytes.

Use [`io.MultiReader`](https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/io/multi.go;drc=946b4baaf6521d521928500b2b57429c149854e7;l=73) to combine a `bytes.NewReader(<buf-with-u64-download-bytes>)` with the and a new zeroReader.

The zeroReader could look something like:

```
type zeroReader struct {
    n int64
}

func (z *zeroReader) Read(p []byte) (int, error) {
    if z.n <= 0 {
        return 0, io.EOF
    }

    for i := range p {
        if z.n <= 0 {
            return i, nil
        }

        p[i] = 0
        z.n--
    }

    return len(p), nil
}
```

perf/impl/https/v0.1/main.go
MarcoPolo: I left a comment below.

MarcoPolo:
Looks good mod last comment about the reader.

perf/impl/https/v0.1/main.go
mxinden: Wasn't aware of `io.MultiReader`. That simplifies things. Thanks!

Fixed with [2296d8c](https://github.com/libp2p/test-plans/pull/184/commits/2296d8c1c8501bebe7a0f5f56379d6374245ebc6). Note that I am still using `copy` instead of zeroing every byte individually in the (maybe premature) hope that Go can optimize some of it. Let me know in case you prefer individual zeroing for each byte instead @MarcoPolo.

marten-seemann (approved):
LGTM, let's ship it!
There's a few follow-up items to this PR, but what we have here is an excellent basis to start iterating on.

Thank you for this clean setup @mxinden, I'm really happy how this turned out

PR #185, opened 2023-05-26
feat: automate performance benchmarking

galargh:
This is a continuation of https://github.com/mxinden/test-plans/pull/2

Diff since the initial review: https://github.com/galorgh/test-plans/compare/perf-terraform...libp2p:test-plans:perf-automation
- removed `workflow_call` trigger from the `perf` workflow (we can add it when we need it)
- added `git push` to the `perf` workflow
- removed `s3 upload` from the `perf` workflow
- moved provisioning/deprovisioning from the `Makefile` to terraform config 
- made it possible to partially apply terraform (the available groups are `common` - IAM user + scale down lambda, `region` - VPC + launch template, `ephemeral` - EC2 instance)
- changed the instructions for local setup back to the ones requiring only `terraform apply` + `terraform destroy`
- created an override terraform config with remote terraform backend for long-lived infra
- moved the lambda testing script to the directory where all the other lambda related files are
- removed s3 from terraform
- added instructions on how to move to ephemeral key pairs as a comment

I'll mark it as ready for review once I reply to the comments raised in the original PR.


perf/terraform/remote/terraform_override.tf
```diff
@@ -0,0 +1,29 @@
+terraform {
+  backend "s3" {
+    bucket         = "terraform-tfstate"
+    key            = "github.com/libp2p/test-plans/perf/terraform/remote/terraform.tfstate"
+    region         = "us-west-2"
+  }
+}
```
galargh, line 7: > Where would the .tfstate file of the long-lived terraform-provisioned infrastructure live?

For IPDX terraforms, we usually store them on S3. We can create an S3 bucket in libp2p account and use it for tfstate storage. I added an example configuration for that here. Bucket name, key, region can be updated of course. Note that this bucket will have to be manually created.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Archive results
+        uses: actions/upload-artifact@v2
+        with:
+          name: results
+          path: perf/runner/benchmark-results.json
+      - name: Push
+        run: |
+          git add perf/runner/benchmark-results.json
+          git commit -m "perf: update benchmark results"
+          git push
```
galargh, line 81: > I don't think this is necessary. For the sake of simplicity, I suggest only pushing the results to the branch of the pull request. Gives us a single source-of-truth, fully version controlled.

Done! This will push benchmark-results.json to the branch from which the workflow was triggered. It will push it using the name of the person who triggered the workflow.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,85 @@
...
+# 4. Go to https://console.aws.amazon.com/iamv2/home?#/users/details/perf?section=security_credentials
+# 5. Click 'Create access key' to get the access key ID and secret access key
+# 6. Go to https://github.com/libp2p/test-plans/settings/secrets/actions
+# 7. Click 'New repository secret', set the name to 'PERF_AWS_SECRET_ACCESS_KEY', and paste the secret access key from step 5
+# 8. Click 'New repository secret', set the name to 'PERF_SSH_PRIVATE_KEY', and paste the private key from step 1
+# 9. Go to https://github.com/libp2p/test-plans/settings/variables/actions
+# 10. Click 'New repository variable', set the name to 'PERF_AWS_ACCESS_KEY_ID', and paste the access key ID from step 5
+
+on:
+  workflow_dispatch:
```
galargh, line 16: > > You'll be able to trigger the workflow only after it is merged to the default branch.

> It is necessary to trigger the workflow before merging into the default branch (i.e. master). Otherwise how would one test a release candidate. Are there any larger blockers to triggering it before the merge?

Sorry, I didn't make it clear enough. I only meant that before we can trigger the workflow for the **first** time, these changes will have to make it to master. Once it does, we'll be able to trigger the workflow from PR branches too.

perf/README.md
```diff
@@ -10,21 +10,27 @@ Benchmark results can be visualized with https://observablehq.com/@mxinden-works
...
-1. `cd terraform`
-2. Save your public SSH key as the file `./user.pub`.
+### Bootstrap
+
+1. Save your public SSH key as the file `./regions/files/user.pub`; or generate a new key pair with `make ssh-keygen` and add it to your SSH agent with `make ssh-add`.
+2. `cd terraform`
 3. `terraform init`
 4. `terraform apply`
+5. `CLIENT_IP=$(terraform output -raw client_ip)`
+6. `SERVER_IP=$(terraform output -raw server_ip)`
```
galargh, line 20: > It adds an additional step to testing against a personal AWS account. (Spin up both long-lived and short-lived resources.)

I restored the original instructions here. With personal AWS account, `terraform apply` will create all the resources required for running the test here, including nodes.

perf/terraform/terraform.tf
```diff
@@ -2,37 +2,78 @@ terraform {
...
 
-module "client_region" {
-  source = "./region"
-  region = "us-east-1"
-  ami = "ami-06e46074ae430fba6"
+variable "common_enabled" {
+  type        = bool
+  description = "Whether or not to create common resources"
+  default     = false
+}
```
galargh, line 21: > Testing against a personal AWS account doesn't require the AWS Lambda.

Using this flag we can now control whether the lambda is created or not. By default it's disabled. So when running this locally, it won't create any lambdas in a personal account.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,85 @@
...
+          path: perf/runner/benchmark-results.json
+      - name: Push
+        run: |
+          git add perf/runner/benchmark-results.json
+          git commit -m "perf: update benchmark results"
+          git push
+      - name: Destroy terraform
+        if: always()
+        run: terraform destroy -auto-approve
+        working-directory: perf/terraform
```
galargh, line 85: > It speeds things up, as everything but the machines no longer need to be spun up per run. Though I don't think the reduced time-to-run matters much for us. As far as I can tell, spinning up the machines is the dominant time factor anyways.

Speed is a nice benefit but in my opinion the biggest advantage of long/short lived infra split is that the cleanup becomes trivial. Finding and terminating EC2 instances is extremely easy.

perf/terraform/region/main.tf
```diff
@@ -111,33 +111,42 @@ resource "aws_security_group" "restricted_inbound" {
   })
 }
 
+# If one wanted to make the key pair ephemeral, these would be the steps:
+# 1. Move this resource to `perf/terraform/ephemeral/main.tf`
+# 2. Add `make ssh-keygen` run step to `.github/workflows/perf.yml`
+# 3. Add `aws_key_pair` cleanup to `perf/terraform/common/files/scale_down.py`
+# 4. Allow `aws_iam_user.perf` from `perf/terraform/common/main.tf` to create/delete/update key pairs
+# 5. Allow `aws_iam_role.scale_down` from `perf/terraform/common/scale_down.tf` to delete key pairs
```
galargh, line 119: > As far as I can tell, it introduces the need for a long lived SSH keypair through the AWS launch template.

We either have to have long lived SSH key pair or allow long lived AWS creds to create new AWS key pairs. With long lived SSH key pair, there's one fewer thing for us to clean up. I included the instructions here on how we could transition to ephemeral SSH keys.

perf/terraform/remote/terraform_override.tf
galargh: > It requires us to persist the .tfstate file for the long-lived resources somewhere. (Though arguably we have to do so for the Lambda anyways.)

⬆️

galargh:
I think I now covered all the suggestions (including applying parts of the _Alternative approach_) from https://github.com/mxinden/test-plans/pull/2. @mxinden let me know what you think :) Once we agree on the design, I'll test it again against IPDX's account (⚠️ WARN ⚠️  it's not tested yet).

.github/workflows/perf.yml
mxinden: Thanks. That makes sense.

perf/terraform/region/main.tf
mxinden: Thanks for writing this out.

> With long lived SSH key pair, there's one fewer thing for us to clean up.

AWS does not charge for SSH keys, right? Given that the clean-up step is just for times where we make mistakes, do we really need to clean them up in the lambda?

Without long-lived SSH keys we have one piece of state less to worry about.

All that said, I am fine either way.

perf/terraform/terraform.tf
mxinden: Very neat! Thank you.

perf/terraform/terraform.tf
```diff
@@ -2,37 +2,78 @@ terraform {
...
     Project = "perf"
   }
+  root = "."
 }
 
-module "client_region" {
-  source = "./region"
-  region = "us-east-1"
-  ami = "ami-06e46074ae430fba6"
+variable "common_enabled" {
```
mxinden, line 17:
```suggestion
variable "cleanup_automation_enabled" {
```

Nit pick: This would be more descriptive to me.

perf/terraform/terraform.tf
```diff
@@ -2,37 +2,78 @@ terraform {
...
+  type        = bool
+  description = "Whether or not to create common resources"
+  default     = false
+}
 
-  common_tags = {
-    Project = "perf"
-  }
+variable "region_enabled" {
```
mxinden, line 23:
```suggestion
variable "long_lived_enabled" {
```

Would be more descriptive to me.

perf/terraform/terraform.tf
```diff
@@ -2,37 +2,78 @@ terraform {
...
+  type        = bool
+  description = "Whether or not to create regional resources"
+  default     = true
 }
 
-output "server_public_ip" {
-  value       = module.server_region.node_public_ip
-  description = "Public IP address of the server instance"
+variable "ephemeral_enabled" {
```
mxinden, line 29:
```suggestion
variable "short_lived_enabled" {
```

perf/terraform/terraform.tf
```diff
@@ -2,37 +2,78 @@ terraform {
...
+  root = "."
 }
 
-module "client_region" {
-  source = "./region"
-  region = "us-east-1"
-  ami = "ami-06e46074ae430fba6"
+variable "common_enabled" {
+  type        = bool
+  description = "Whether or not to create common resources"
```
mxinden, line 19:
```suggestion
  description = "Whether or not to create cleanup automation (lambda) removing old resources"
```

perf/terraform/terraform.tf
```diff
@@ -2,37 +2,78 @@ terraform {
...
+  default     = false
+}
 
-  common_tags = {
-    Project = "perf"
-  }
+variable "region_enabled" {
+  type        = bool
+  description = "Whether or not to create regional resources"
```
mxinden, line 25:
```suggestion
  description = "Whether or not to create long lived regional resources used across CI runs"
```

perf/terraform/terraform.tf
```diff
@@ -2,37 +2,78 @@ terraform {
...
+  default     = true
 }
 
-output "server_public_ip" {
-  value       = module.server_region.node_public_ip
-  description = "Public IP address of the server instance"
+variable "ephemeral_enabled" {
+  type        = bool
+  description = "Whether or not to create ephemeral resources"
```
mxinden, line 31:
```suggestion
  description = "Whether or not to create short lived resources specific to each CI run"
```

.github/workflows/perf.yml
```diff
@@ -0,0 +1,85 @@
...
+        env:
+          SERVER_IP: ${{ steps.server.outputs.ip }}
+          CLIENT_IP: ${{ steps.client.outputs.ip }}
+        run: npm run start -- --client-public-ip $CLIENT_IP --server-public-ip $SERVER_IP
+        working-directory: perf/runner
+      - name: Archive results
+        uses: actions/upload-artifact@v2
+        with:
+          name: results
+          path: perf/runner/benchmark-results.json
```
mxinden, line 76: Is this step still necessary? I guess it doesn't hurt?

mxinden (commented):
Overall this looks good to me. Thank you for the follow-ups! Feel free to ignore the naming suggestions.

> Once we agree on the design, I'll test it again against IPDX's account (warning WARN warning it's not tested yet).

Agreed on the design. Mind doing a test run?

perf/terraform/region/main.tf
galargh: Yeah, it doesn't cost anything. I moved the SSH key generation and added TODOs for cleaning up the key pairs.

mxinden:
Thanks for the follow-ups. Let me know once you want me to take another look.

galargh:
I renamed the modules according to suggestions, made the SSH keys ephemeral (without automated cleanup), skipped archiving in the workflow unless push is disabled and tested the new setup against IPDX AWS account.

Here's an example workflow run: https://github.com/galorgh/test-plans/actions/runs/5111435405/jobs/9188390217

And here's a commit created by the automation: https://github.com/galorgh/test-plans/commit/d787a351bfacfad8008838266689cc34ae899819

Let me know if there's anything else you'd want me to look at 😄

perf/terraform/configs/remote/terraform_override.tf
```diff
@@ -0,0 +1,20 @@
+terraform {
+  # TODO: Uncomment to start using remote backend
+  # backend "s3" {
+  #   bucket         = "terraform-tfstate"
+  #   key            = "github.com/libp2p/test-plans/perf/terraform/configs/remote/terraform.tfstate"
+  #   region         = "us-west-2"
+  # }
+}
```
mxinden, line 8: Shouldn't this be always uncommented? In other words wouldn't you expect someone to always use an s3 bucket to store the long lived state?

perf/terraform/configs/local/terraform.tf
```diff
@@ -0,0 +1,124 @@
...
+  value = var.short_lived_enabled ? module.short_lived_client[0].public_ip : null
+}
+
+output "server_ip" {
+  value = var.short_lived_enabled ? module.short_lived_server[0].public_ip : null
+}
+
+output "test" {
+  value = "hello world"
+}
```
mxinden, line 124: Still needed?

mxinden (approved):
Two smaller comments. Otherwise this is ready from my side.

Feel free to merge into `perf` branch.

perf/terraform/configs/local/terraform.tf
galargh: Good catch! Thanks! You're right, it's a leftover from debugging output retrieval. 

```suggestion
```

perf/terraform/configs/local/terraform.tf
```diff
@@ -0,0 +1,121 @@
...
+  type        = bool
+  description = "Whether or not to create long lived resources (in CI, used across runs; e.g. VPCs)"
+  default     = true
+}
+
+variable "short_lived_enabled" {
+  type        = bool
+  description = "Whether or not to create short lived resources (in CI, specific to each run; e.g. EC2 instances)"
+  default     = true
+}
```
mxinden, line 49: Playing with this, I am still a bit confused.

If I understand correctly the `.github/workflows/perf` workflow will run the `perf/terraform/local` config, see above. That config has `long_lived_enabled` and `short_lived_enabled` set to `true`. The latter makes sense. Why does it set the former? Shouldn't it use the long-lived resources created outside of the workflow?

Looking at it from a different perspective, don't we really need 3 configs?

1. `configs/local` for local development with `ci_enabled` `false`, `long_lived_enabled` `true`, `short_lived_enabled` `true` and no S3 backend.
2. `configs/ci_long_lived` run once outside of the GitHub workflow with `ci_enabled` `true`, `long_lived_enabled` `true`, `short_lived_enabled` `false` and S3 backend.
3. `configs/ci_short_lived` run once outside of the GitHub workflow with `ci_enabled` `false`, `long_lived_enabled` `false`, `short_lived_enabled` `true` and no S3 backend.

perf/terraform/configs/remote/terraform_override.tf
```diff
@@ -0,0 +1,20 @@
+terraform {
+  # TODO: Uncomment to start using remote backend
+  # backend "s3" {
+  #   bucket         = "terraform-tfstate"
```
mxinden, line 4: I just checked and we already have a bucket with the name `libp2p-terraform-state` in the libp2p AWS account. What is the best practice? A single bucket with many `.tfstate` files differentiated by `key`, or multiple buckets, one per project, e.g. in this case one separate bucket for the libp2p-perf project.

perf/terraform/configs/local/terraform.tf
galargh: That's a very valid question because it's not immediately obvious why we have different configs in the first place. And the answer lies within terraform internals which we definitely shouldn't require users to know.

So the terraform specific thing I'm referring to here is that terraform backend (i.e. config of where the state is stored) cannot be defined dynamically. That's the reason why we need 2 separate configs - one where state is stored locally, and the other where the state is stored remotely (in this case, on S3). On the other hand, what modules are created during terraform apply can be configured dynamically through terraform variables. That's why we only need 2 configs.

So how are we running tf and where?
1. We use `config/local` during local development. `config/local` defaults are set so that only `long_lived` and `short_lived` resources are created. Because of that, the user doesn't have to specify any additional input when doing terraform apply or destroy. The state is stored locally.
2. We use `config/remote` when we want to do the initial set up of the resources required for the CI. By default, this config creates only `ci` and `long_lived` resources. Again, no additional input is required when we use this config. The state is stored remotely, in a S3 bucket.
3. We use `config/local` when we want to start instances for CI runs. We use `config/local` because we want the state to be stored locally but `config/local`, by default, creates `long_lived` and `short_lived` resources. This is too much so we have to override the defaults. This is fine because we don't put the burden of overriding defaults on users. Instead, we just configure it once in the workflow YAML via `TF_VAR_*` env vars, see https://github.com/libp2p/test-plans/blob/020d5097fecb4c6e2c271d8748bb5841fe0d9e1a/.github/workflows/perf.yml#L35-L37

Because it's a bit tricky, I included a README next to the configs: https://github.com/libp2p/test-plans/blob/020d5097fecb4c6e2c271d8748bb5841fe0d9e1a/perf/terraform/configs/README.md

perf/terraform/configs/remote/terraform_override.tf
galargh: Personally, if you already have `libp2p-terraform-state` bucket, I'd use this one. I've done bucket per project in some places at PL and it feels like an unnecessary overhead.

```suggestion
  backend "s3" {
    bucket         = "libp2p-terraform-state"
    key            = "github.com/libp2p/test-plans/perf/terraform/configs/remote/terraform.tfstate"
    region         = "us-west-2"
  }
}
```

Could you check what's the region of the bucket?

mxinden: > Could you check what's the region of the bucket?

`us-west-2` if I recall correctly.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,104 @@
...
+
+# How to configure a repository for running this workflow:
+# 1. Configure auth for the AWS provider as per https://registry.terraform.io/providers/hashicorp/aws/latest/docs#authentication-and-configuration
+# 2. Run 'terraform init' and 'terraform apply' in 'perf/terraform/configs/remote' to create the resources needed for this workflow
+# 3. Go to https://console.aws.amazon.com/iamv2/home?#/users/details/perf?section=security_credentials
+# 4. Click 'Create access key' to get the access key ID and secret access key
+# 5. Go to https://github.com/libp2p/test-plans/settings/secrets/actions
+# 6. Click 'New repository secret', set the name to 'PERF_AWS_SECRET_ACCESS_KEY', and paste the secret access key from step 5
+# 7. Go to https://github.com/libp2p/test-plans/settings/variables/actions
+# 8. Click 'New repository variable', set the name to 'PERF_AWS_ACCESS_KEY_ID', and paste the access key ID from step 5
```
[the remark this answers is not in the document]
galargh: Yes, that's right!

perf/terraform/configs/remote/terraform_override.tf
galargh: Cheers :) I uncommented the bucket name now - https://github.com/libp2p/test-plans/pull/185/commits/4179aa67c5836d880d88b000ec254d60cfdfc38c

.github/workflows/perf.yml
```diff
@@ -0,0 +1,104 @@
...
+        shell: bash
+        working-directory: perf
+    env:
+      AWS_ACCESS_KEY_ID: ${{ vars.PERF_AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.PERF_AWS_SECRET_ACCESS_KEY }}
+      TF_IN_AUTOMATION: 1
+      TF_INPUT: 0
+      TF_VAR_ci_enabled: false
+      TF_VAR_long_lived_enabled: false
+      TF_VAR_short_lived_enabled: true
```
mxinden, line 37:
```suggestion
```

perf/terraform/configs/local/terraform.tf
```diff
@@ -0,0 +1,121 @@
...
+variable "ci_enabled" {
+  type        = bool
+  description = "Whether or not to create resources required to automate the setup in CI (e.g. IAM user, cleanup Lambda)"
+  default     = false
+}
+
+variable "long_lived_enabled" {
+  type        = bool
+  description = "Whether or not to create long lived resources (in CI, used across runs; e.g. VPCs)"
+  default     = true
```
mxinden, line 42:
```suggestion
  default     = false
```

perf/terraform/configs/README.md
```diff
@@ -0,0 +1,13 @@
+# Configs
+
+The terraform configs defined in this directory are used to provision the infrastructure for the libp2p perf tests.
+
+The configs are named after the type of backend they use. The defaults for what parts of infrastructure they provision differ between the two.
+
+## local
+
+Terraform state in this configuration will be stored locally. The defaults are configured for local development i.e. terraform apply will bring up ALL the infrastructure needed to run the perf tests. It will skip the steps required only for the CI environment.
```
mxinden, line 9:
```suggestion
Terraform state in this configuration will be stored locally. The defaults are configured for a single performance benchmark run, i.e. `terraform apply` will bring up short-lived infrastructure only. It will skip long-lived infrastructure like the clean-up Lambda and the instance launch template.
```

perf/terraform/configs/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## remote
+
+Terraform state here will be stored remotely in an S3 bucket. The defaults are configured for the CI environment i.e. terraform apply will only bring up the long-lived infrastructure needed to run the perf tests in CI. It will skip launching EC2 instances because they will be brought up by the CI environment.
```
mxinden, line 13:
```suggestion
Terraform state here will be stored remotely in an S3 bucket. `terraform apply` will only bring up the long-lived infrastructure needed to run the performance benchmarks It will skip short-lived infrastructure like launching EC2 instances.
```

mxinden (commented):
@galargh following up on out-of-band discussion, adjusting the default values to targeting both CI and personal run in the libp2p AWS acccount with long-lived infrastructure pre-provisioned. Looks good to you?

mxinden:
`perf/terraform/config/remote` is now deployed with the `.tfstate` in the corresponding S3 bucket. //CC @galargh

mxinden (approved):

PR #186, opened 2023-05-29
chore(.editorconfig): add insert_final_newline option

mxinden:
Instructs editors to insert final newline and thus prevents Git and GitHub to show "No newline at end of file".

See also https://github.com/libp2p/rust-libp2p/pull/3578

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,1 @@
+insert_final_newline=true


marten-seemann:
This should be in your local editor config. Otherwise we'd have to commit this to every single repository in the org.

mxinden:
:D this is usually my exact same argument when folks try to include their editor XXX config files in a pull request. 

I don't have a strong opinion here, other than definitely not including editor XXX config files. I usually treat `.editorconfig` as generic enough to be included in repositories directly.

Anyways, closing here.

PR #187, opened 2023-06-02
Add js-libp2p v0.45.0

MarcoPolo:
Uses the new in-repo test to test js-libp2p. 

Note some of the build steps are copied here because we actually want a slightly different build setup. We want to support running `make` and it building in a reproducible way on both MacOS and Linux. In the js-libp2p repo, we fetch node_modules in the CI context (not docker) to get a slight speed up. Here we do `npm i` within the docker build process.

This won't slow us down here however because we can cache these builds. That's not the case in the js-libp2p head builds when you're always building a new head.

### DIFF
--- a/multidim-interop/impl/js/v0.45/BrowserDockerfile
+++ b/multidim-interop/impl/js/v0.45/BrowserDockerfile
@@ -1,0 +1,18 @@
+# syntax=docker/dockerfile:1
+
+# Copied since we won't have the repo to use if expanding from cache.
+
+# Workaround: https://github.com/docker/cli/issues/996
+ARG BASE_IMAGE=node-js-libp2p-head
+FROM ${BASE_IMAGE} as js-libp2p-base
+
+FROM mcr.microsoft.com/playwright
+
+
+COPY --from=js-libp2p-base /app/ /app/
+WORKDIR /app/interop
+RUN ./node_modules/.bin/playwright install
+ARG BROWSER=chromium # Options: chromium, firefox, webkit
+ENV BROWSER=$BROWSER
+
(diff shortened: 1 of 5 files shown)


multidim-interop/versions.ts
```diff
@@ -98,6 +108,21 @@ export const versions: Array<Version> = [
         secureChannels: ["noise"],
         muxers: ["mplex", "yamux"],
     },
+    // TODO enable webrtc direct for v0.45
```
mxinden, line 111: What is the blocker here? We do support it for `js-libp2p` `v0.44`.

https://github.com/libp2p/test-plans/blob/4a3553280826eb1386b67e183aa521b27ef7ac83/multidim-interop/versions.ts#L94-L100

multidim-interop/impl/js/v0.45/BrowserDockerfile
```diff
@@ -0,0 +1,18 @@
+# syntax=docker/dockerfile:1
+
+# Copied since we won't have the repo to use if expanding from cache.
+
+# Workaround: https://github.com/docker/cli/issues/996
+ARG BASE_IMAGE=node-js-libp2p-head
+FROM ${BASE_IMAGE} as js-libp2p-base
+
+FROM mcr.microsoft.com/playwright
+
```
mxinden, line 10:
```suggestion
```

mxinden (approved):

multidim-interop/impl/js/v0.45/BrowserDockerfile
MarcoPolo: Will do in a follow up PR

PR #189, opened 2023-06-02
Reintroduce js-libp2p v0.45.0 and fix build from cache

MarcoPolo:
#187 with an extra commit that fixes the builds from cache. The issue is that the browser images which are based on the node image for speed relied on the image being tagged with a certain name. When loading from cache we don't load the tag name again, just the image and reference it via the digest.

If we could reference base images by their digest that would have also fixed this issue.

### DIFF
--- a/multidim-interop/impl/js/v0.45/Makefile
+++ b/multidim-interop/impl/js/v0.45/Makefile
@@ -1,0 +1,30 @@
+image_name := js-v0.45
+commitSha := 41641f1a7656aa654234f6f849b1749786867121
+
+# TODO Enable webkit once https://github.com/libp2p/js-libp2p/pull/1627 is in
+all: image.json chromium-image.json firefox-image.json
+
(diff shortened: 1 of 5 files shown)


multidim-interop/impl/js/v0.45/Makefile
```diff
@@ -4,12 +4,16 @@ commitSha := 41641f1a7656aa654234f6f849b1749786867121
 # TODO Enable webkit once https://github.com/libp2p/js-libp2p/pull/1627 is in
 all: image.json chromium-image.json firefox-image.json
 
-chromium-image.json: image.json BrowserDockerfile
```
mxinden: The dependency of `BrowserDockerfile` is still valid, right? In other words, changing the `BrowserDockerfile` should lead to a rebuild of the `chromium-image.json` target.

mxinden (approved):

PR #190, opened 2023-06-02
Don't include muxer and secure channel for webrtc

MarcoPolo:
As with quic, webtransport, and webrtc-direct.


mxinden (approved):
Looks good to me. 

In addition `webrtc` should be filtered out from `queryResults`, right?

``` diff
@@ -61,6 +61,7 @@ export async function buildTestSpecs(versions: Array<Version>): Promise<Array<Co
                      AND ma.muxer == mb.muxer
                      -- quic only uses its own muxer/securechannel
                      AND a.transport != "webtransport"
+                     AND a.transport != "webrtc"
                      AND a.transport != "webrtc-direct"
                      AND a.transport != "quic"
                      AND a.transport != "quic-v1";`);
```

mxinden (approved):
:pray:

PR #192, opened 2023-06-13
chore: add go-libp2p 0.28.0

p-shahi:
closes https://github.com/libp2p/test-plans/issues/191

### DIFF
--- a/multidim-interop/impl/go/v0.28/Makefile
+++ b/multidim-interop/impl/go/v0.28/Makefile
@@ -1,0 +1,21 @@
+image_name := go-v0.28
+commitSha := 703c3a4377d1ca2cf3fc93c92c883be1644f3ba6
+
+all: image.json
+
+image.json: go-libp2p-${commitSha}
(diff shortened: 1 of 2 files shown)


marten-seemann (approved):
Haven’t checked the commit hash, but other than that, lgtm.

As an aside: it would be nice if the setup didn’t require us to insert a hash here, but accepted everything that git itself accepts as a reference (branches, tags, etc.).

MarcoPolo:
@marten-seemann please create an issue. We can do something like this, but we should pin some hash. Branch names and tags are mutable.

marten-seemann:
Since we don't retract / change tags, they're not really mutable.

And even if we did, that would be a feature, not a bug. We're not creating a system here that's resistant against supply chain attacks (that's the reason we use hashes in Unified CI), but a system to test the releases that we create ourselves. Of course if we say that we want to test vX, this should test vX.
Addressing them by their tag name makes the configuration a lot more readable than a hash.

MarcoPolo:
Let's continue discussion there.

PR #199, opened 2023-06-16
Rename `build-without-cache` to `build-without-secrets`

MarcoPolo:
This workflow uses the public cache. It just can't update it. The goal of this workflow is to see how public forks behave. The name is a vestige from when we didn't have a public cache. But the original goal was always to see how public forks and external repos would behave.


thomaseizinger (approved):

thomaseizinger:
Thanks!

mxinden (approved):
:+1:

PR #197, opened 2023-06-16
add Java to interop tests

kevodwyer:
all java impl tests pass in both directions

PR #200, opened 2023-06-20
fix(java-libp2p): update test client with resource fix

kevodwyer:
Remove incorrectly calling node.stop() twice
commit: https://github.com/Peergos/nabu/commit/6fad470c92a5c4ab368bc615e64dbf40b5ff9ccb


thomaseizinger:
Also see https://github.com/libp2p/test-plans/pull/197#issuecomment-1598179085.

thomaseizinger (approved):
Thanks!

PR #201, opened 2023-06-21
perf: simplify go-libp2p and https implementation

marten-seemann:
Lots of nits. The biggest change here is that I refactored `sendBytes` and `drainStream` into pure functions, and copy-pasted them into the https implementation. Using an `io.MultiReader` allows us to simplify the reading of the `http.Request.Body`.

@mxinden Feel free to merge (this is targeting #184). Does this PR trigger a CI run? Would be good in order to see if this actually works.

### DIFF
--- a/perf/impl/https/v0.1/main.go
+++ b/perf/impl/https/v0.1/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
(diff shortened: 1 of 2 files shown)


perf/impl/https/v0.1/main.go
```diff
@@ -21,88 +22,71 @@ import (
...
+	if uint64(len(b)) > remaining {
+		l = remaining
 	}
-	return b
+	r.read += l
+	if r.read == r.N {
+		return int(l), io.EOF
+	}
+	return int(l), nil
 }
```
mxinden, line 70: I am assuming that it is safe in Golang to return "uninitialized" bytes, i.e. the original bytes in `b` to a remote machine. E.g. there is no risk of `b` containing deallocated secret key material.

Just for others, returning non-zero bytes if just fine according to the perf specification:

> The bytes themselves should be a predetermined arbitrary set of bytes. Zero is fine, but so is random bytes (as long as it's not a different set of random bytes, because then you may be limited by how fast you can generate random bytes).

https://github.com/libp2p/specs/blob/master/perf/perf.md

mxinden (approved):
Great. Thank you @marten-seemann!

PR #203, opened 2023-06-22
fix(perf/https): request downloadBytes instead of uploadBytes

mxinden:
Small follow-up to https://github.com/libp2p/test-plans/pull/184.


mxinden:
Merging as is here. `perf/` is not tested by our multidimensional interop CI tests.

PR #205, opened 2023-06-23
fix(perf/go-libp2p): close write after sending in client

mxinden:
Follow up to https://github.com/libp2p/test-plans/pull/184.


mxinden:
Merging here since none of our interop CI tests test these code paths.

PR #206, opened 2023-06-23
feat(interop): add rust-libp2p v0.52

mxinden:
Adds rust-libp2p v0.52 to interop tests.

See https://github.com/libp2p/rust-libp2p/releases/tag/libp2p-v0.52.0

---

@MarcoPolo to reduce the time to run the whole interop test suite, should I remove [rust-libp2p v0.48](https://github.com/libp2p/rust-libp2p/releases/tag/v0.48.0) released in Sep 7, 2022 within the same commit. I am fine either way, I don't think we need to continue testing v0.48, at the same time, I don't mind the longer test time.


thomaseizinger (approved):
FWIW, I am in favour of removing `0.48`

multidim-interop/versions.ts
```diff
@@ -61,6 +61,13 @@ export const versions: Array<Version> = [
         secureChannels: ["tls", "noise"],
         muxers: ["mplex", "yamux"],
     },
+    {
+        id: "rust-v0.52.0",
+        containerImageID: rustv051.imageID,
```
MarcoPolo, line 66:
```suggestion
        containerImageID: rustv052.imageID,
```

MarcoPolo: We could make this derived from the id which would prevent this kind of issue.

MarcoPolo:
In the future, could you please create a PR from a branch in this repo rather than a fork so we can have CI cache this?

multidim-interop/versions.ts
```diff
@@ -33,13 +34,6 @@ export type Version = {
 }
 
 export const versions: Array<Version> = [
-    {
-        id: "rust-v0.48.0",
```
marten-seemann, line 37: Do you also need to remove the import in L8?

marten-seemann (approved):

multidim-interop/versions.ts
mxinden: Good catch. Fixed with https://github.com/libp2p/test-plans/pull/206/commits/dc85d3a93ab1e8b38421e863c00458e386f9ecff.

PR #207, opened 2023-06-23
refactor(perf): expose single latency measurement 

mxinden:
Instead of exposing the time to establish a connection, the time to upload the bytes and the time to download the bytes, expose a single time including all three.

The rational here is, that differentiation of the three is flawed. E.g. when does one stop the upload timer and start the download timer? When the last byte is sent? When the last byte is flushed? When the first byte is received?

See https://github.com/libp2p/test-plans/pull/184#pullrequestreview-1482600521 for past discussion.

I will update the dashboard once this is merged.

Blocked on https://github.com/libp2p/rust-libp2p/pull/4105 and https://github.com/quic-go/perf/pull/12.

### DIFF
--- a/perf/impl/quic-go/v0.34/Makefile
+++ b/perf/impl/quic-go/v0.34/Makefile
@@ -1,4 +1,4 @@
-commitSha := 82a80f5174ba4f84f99cdbbf119b0bfe1e1f971b
+commitSha := a5cd126c97b6d8d8328141bfa84cc57e74ebc57c
 
 all: perf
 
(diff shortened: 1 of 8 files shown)


perf/impl/quic-go/v0.34/Makefile
```diff
@@ -10,7 +10,7 @@ perf-${commitSha}: perf-${commitSha}.zip
 	unzip -o perf-${commitSha}.zip
 
 perf-${commitSha}.zip:
-	wget -O $@ "https://github.com/quic-go/perf/archive/${commitSha}.zip"
+	wget -O $@ "https://github.com/mxinden/perf/archive/${commitSha}.zip"
```
marten-seemann, line 13: I just merged https://github.com/quic-go/perf/pull/12, let's update the PR here.

marten-seemann (approved):
LGTM, module the updated quic-go/perf commit hash. Feel free to merge once you've made that change.

PR #208, opened 2023-06-23
Less error-prone way of defining image IDs and other quality of life improvements

MarcoPolo:
Changes how imageIDs are loaded for a given version.

* imageID can now be a function that returns an imageID rather than directly a string.
* Defines a new canonical lookup function which matches how all current implementations are defined. A new implementation is not required to lay things out like this, but if it does it can reuse this function.
* imageIDs are now lazily loaded _after_ running the name-filter. This means, for example, if you only want to test WebTransport you don't have to build every implementation first.

This should make errors like [this](https://github.com/libp2p/test-plans/pull/206#discussion_r1240153895) harder to make. (I've made this error several times as well).

### DIFF
--- a/multidim-interop/versions.ts
+++ b/multidim-interop/versions.ts
@@ -1,30 +1,10 @@
-import gov028 from "./impl/go/v0.28/image.json"
-import gov027 from "./impl/go/v0.27/image.json"
-import gov026 from "./impl/go/v0.26/image.json"
-import gov025 from "./impl/go/v0.25/image.json"
-import gov024 from "./impl/go/v0.24/image.json"
-import gov023 from "./impl/go/v0.23/image.json"
-import gov022 from "./impl/go/v0.22/image.json"
-import rustv049 from "./impl/rust/v0.49/image.json"
-import rustv050 from "./impl/rust/v0.50/image.json"
-import rustv051 from "./impl/rust/v0.51/image.json"
-import rustv052 from "./impl/rust/v0.52/image.json"
-import jsV041 from "./impl/js/v0.41/node-image.json"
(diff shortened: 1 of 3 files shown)


multidim-interop/versions.ts
```diff
@@ -1,195 +1,193 @@
...
+    const imageIDJSON = fs.readFileSync(`./impl/${impl}/${versionFolder}/image.json`, "utf8")
+    const imageID = JSON.parse(imageIDJSON).imageID
+    return imageID
+}
+
 export const versions: Array<Version> = [
     {
         id: "rust-v0.48.0",
-        containerImageID: rustv048.imageID,
+        containerImageID: canonicalImageIDLookup,
```
marten-seemann, line 36: Can this be made the default, i.e. use `canonicalImageIDLookup` if unset, so you don't have to define it here 50 times?

mxinden: Good idea!

mxinden (approved):
Thank you @MarcoPolo!

I am fine merging as is, though preference to include Marten's suggestion.

multidim-interop/versions.ts
MarcoPolo: I thought about making it default, but I felt like it obfuscated what was happening a bit more than being explicit. Think of someone looking at this file for the first time. The containerImageID can be a string or it can be a function that returns a string. If it's a function you can go directly to that function at look at what it does. Straightforward, and nothing hidden. If we make this default, there's a missing piece between where the version is defined and how the containerImageID is defined.

If I'm in the minority opinion here, I'm fine to make this change. Just want to outline my thoughts.

PR #211, opened 2023-06-29
multidim-interop: Remove often wrong patch version from id

MarcoPolo:
These patch versions are often wrong since we rarely update them. It's not necessary to include the patch version in the id, so we may as well omit rather than be wrong.


marten-seemann (approved):

thomaseizinger (approved):

thomaseizinger:
What if we were to remove `versions.ts` completely in favor of `version.json` files within the various directories and infer the version from the directory name?

MarcoPolo:
> What if we were to remove `versions.ts` completely in favor of `version.json` files within the various directories and infer the version from the directory name?

A couple of benefits of preferring TypeScript:
* The data is typed.
* Comments are supported.
* We can use functions to load the image id as done in https://github.com/libp2p/test-plans/pull/208.

Having separate `version.ts` files in each impl folder would work, but seems at least a little tricky. We'd have to import them in the main `version.ts`.

If we used a JSON file we'd lose the above benefits, but could at run time check for `version.json` in all the folders. There's also a slight friction in figuring out the image sha to use for image id to put in the JSON file. You could have the makefile generate the version.json, but then things would silently break if this didn't exist. Or you could use a tag name, but then you may be testing the wrong exact version without noticing (e.g. testing v0.51.0 instead of v0.51.1 if the tag name is rust-v0.51).

Maybe there's something else you can do here that would improve this, I haven't spent much thought. But would it be _that_ much better? If it's only marginally better I wouldn't want to spend time doing it.

thomaseizinger:
> > What if we were to remove `versions.ts` completely in favor of `version.json` files within the various directories and infer the version from the directory name?
> 
> A couple of benefits of preferring TypeScript:
> * The data is typed.

We could achieve something similar using JSON schema.

> * Comments are supported.

There is JSONC (JSON with comments).

> * We can use functions to load the image id as done in https://github.com/libp2p/test-plans/pull/208.

We can still do that!

> could at run time check for `version.json` in all the folders.

That is what I mean. At the moment, there is a duplication here where each directory needs an entry in this array.

It doesn't really matter whether we specify the supported transports here in one place or in each directory, is there?

> may be testing the wrong exact version without noticing (e.g. testing v0.51.0 instead of v0.51.1 if the tag name is rust-v0.51).

Isn't that already a problem? What you reference by hash needs to match what you specify in this array here. Thinking about it, because we have the JSON file in the repository anyway (at least for rust-libp2p, but I think for go too?), wouldn't it make even more sense to just directly refer to that?

> Maybe there's something else you can do here that would improve this, I haven't spent much thought. But would it be _that_ much better? If it's only marginally better I wouldn't want to spend time doing it.

I agree that it is probably not worth investing much time. I might look into it as part of the relay tests.

thomaseizinger:
> Thinking about it, because we have the JSON file in the repository anyway (at least for rust-libp2p, but I think for go too?), wouldn't it make even more sense to just directly refer to that?

The logic could be: The runner looks for a `version.jsonc` file in each impl directory.

For implementations like `rust-libp2p`, we would just copy the one from the source code zip into the top-level directory.

Other impls can also statically include it there in case the tests are not part of the repository (yet).

The container hash would be looked up in an `image.json` file and merged with `version.jsonc` by the runner.

PR #213, opened 2023-06-30
fix(perf): increase iterations

mxinden:
- Increase upload/download benchmark iterations. Still conservative value of `10` due to slow rust-libp2p tcp. See https://github.com/libp2p/rust-yamux/issues/162. Note however that this is 10x an upload of 100MB.
- Increase latency benchmark iterations.

### DIFF
--- a/perf/runner/src/index.ts
+++ b/perf/runner/src/index.ts
@@ -19,7 +19,7 @@ async function main(clientPublicIP: string, serverPublicIP: string) {
                  uploadBytes: 100 << 20,
                  downloadBytes: 0,
                  unit: "bit/s",
-                 iterations: 5,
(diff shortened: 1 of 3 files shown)


perf/runner/src/index.ts
```diff
@@ -37,7 +37,7 @@ async function main(clientPublicIP: string, serverPublicIP: string) {
                  uploadBytes: 1,
                  downloadBytes: 1,
                  unit: "s",
-                 iterations: 100,
+                 iterations: 1000,
```
mxinden, line 40:
```suggestion
                 iterations: 100,
```

mxinden:
Merging here without review given the time that passed and the fact that it is a small change only.

PR #215, opened 2023-07-03
fix(perf): increase MAX_AGE_MINUTES to 120

mxinden:
- When experimenting with increasing iterations, instances are taken down prematurely. See e.g. https://github.com/libp2p/test-plans/actions/runs/5418669631/jobs/9850992882

- When benchmarking from a local developer machine, experimenting for longer is useful.

- The AWS Lambda is only a safety net machanism. In the rare case where we miss to clean up, running the machines for <1h or 2h doesn't matter cost wise.


perf/terraform/modules/ci/cleanup.tf
```diff
@@ -18,7 +18,7 @@ resource "aws_lambda_function" "cleanup" {
     variables = {
       REGIONS         = jsonencode(var.regions)
       TAGS            = jsonencode(var.tags)
-      MAX_AGE_MINUTES = 50
+      MAX_AGE_MINUTES = 120
```
mxinden, line 21: First

perf/terraform/modules/ci/files/cleanup.yml
```diff
@@ -13,7 +13,7 @@ Resources:
         Variables:
           REGIONS: '["us-west-2", "us-east-1"]'
           TAGS: '{"Project":"perf", "Name":"node"}'
-          MAX_AGE_MINUTES: '30'
+          MAX_AGE_MINUTES: '120'
```
mxinden, line 16: Second

mxinden (commented):
@galargh I don't understand why we need two assignments to `MAX_AGE_MINUTES`. Can you comment on this?

perf/terraform/modules/ci/files/cleanup.yml
galargh: This YAML is only used for local testing. I just proposed moving the test files to a separate directory to make the two sets more distinct - https://github.com/libp2p/test-plans/pull/216

galargh (approved):

mxinden:
Merging here. None of these changes affect `multidim-interop/` and thus none of the CI tests apply.

mxinden:
I will apply the changes via terraform. Will report once done.

mxinden:
> I will apply the changes via terraform. Will report once done.

Applied via `perf/config/remote`. State is stored in the S3 bucket.

Issue #218, opened 2023-07-03
interop: stop testing `webtransport` with sec and mux


mxinden:
Based on the logs we seem to be testing the WebTransport Transport across the security and multiplexer dimension:

```
Running test spec: go-v0.24 x go-v0.28 (webtransport, tls, yamux)
Finished: go-v0.24 x go-v0.24 (wss, noise, mplex) { handshakePlusOneRTTMillis: 11.253, pingRTTMilllis: 1.863 }
Running test spec: go-v0.24 x go-v0.28 (webtransport, noise, yamux)
Finished: go-v0.24 x go-v0.28 (webtransport, tls, yamux) { handshakePlusOneRTTMillis: 103.277, pingRTTMilllis: 2.847 }
```

https://github.com/libp2p/test-plans/actions/runs/5441933224/jobs/9896488441?pr=217

Given that WebTransport Transport leverages QUIC, there is no additional security and multiplexer upgrade required. Something is off here.

Am I missing something @MarcoPolo?

PR #217, opened 2023-07-03
feat(interop): add rust-libp2p chromium WASM WebTransport

mxinden:
Introduced in rust-libp2p with https://github.com/libp2p/rust-libp2p/pull/4015.

PR #219, opened 2023-07-05
fix(interop): use standaloneTransports as string instead of param

mxinden:
When generating the test specification (docker-compose file) we enumerate all transport permutations. While some of our transports need to be upgraded (e.g. TCP and WS) some don't (e.g. QUIC or WebRTC). To differentiate the two we execute two SQL queries (`queryResults` and `standaloneTransportsQueryResults`), each with either a `AND _ NOT IN` or `AND _ IN`.

The previous implementation would use `node-sqlite3` query parameters (see https://github.com/TryGhost/node-sqlite3/wiki/API#runsql--param---callback) to provide the list of transport protocols that don't need upgrading as an array.

The problem is that `node-sqlite3` does not support arrays as query parameters, see https://github.com/TryGhost/node-sqlite3/issues/762.

This commit uses string interpolation instead to inject the list of standalone transports into the SQL queries.

Follow-up to https://github.com/libp2p/test-plans/pull/210/.
Fixes https://github.com/libp2p/test-plans/issues/218
Unblocks https://github.com/libp2p/test-plans/pull/217


thomaseizinger:
Should we have an output from the action, counting the number of run tests so we can assert it after?

MarcoPolo (approved):
Unfortunate that we **have** to use string interopolation here :(

MarcoPolo:
I'm going to merge this despite CI not being happy because the current state of master is wrong. CI is being flaky, but should be fixed by https://github.com/libp2p/test-plans/pull/224

PR #224, opened 2023-07-05
More resilient stdout parsing of dialer stdout

MarcoPolo:
First off, I don't really understand why compose split the line [here](https://github.com/libp2p/test-plans/actions/runs/5468772660/jobs/9956843684?pr=219#step:3:1199). It may be a bug in zig-libp2p, but there are no other callers of stdout (no race). And there's nothing fancy going on, it just writes to the stdout file the bytes directly (no buffering).

Whatever the reason, we should probably be a bit more resilient with parsing of the output compared to the initial regex. This adds a helper to help with parsing compose's output as well as a test case to make sure that we don't silently break this.

To avoid adding a whole new dependency on some test framework that opens the door to the nightmare of JS bundling, the test is a simple function that is run if you run the file directly (as opposed to importing it). And it's run as part of `npm test` when you run the interop tests (so it'll happen in CI as well).

### DIFF
--- a/multidim-interop/package.json
+++ b/multidim-interop/package.json
@@ -4,7 +4,7 @@
     "description": "Tests libp2p implementations against each other across various dimensions.",
     "main": "testplans.ts",
     "scripts": {
-        "test": "ts-node testplans.ts",
+        "test": "ts-node src/compose-stdout-helper.ts && ts-node testplans.ts",
         "renderResults": "ts-node renderResults.ts",
         "cache": "ts-node helpers/cache.ts"
     },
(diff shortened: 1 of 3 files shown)


thomaseizinger:
> To avoid adding a whole new dependency on some test framework that opens the door to the nightmare of JS bundling

I still think deno could be a good fit for the problem we have here but I agree with your particular solution :)

MarcoPolo:
Could I get a review from someone? friendly ping to @thomaseizinger or @mxinden

MarcoPolo:
Merging this for now since it's been causing a lot of false negatives. I'm still interested in a review please :)

mxinden (commented):
:+1:

thomaseizinger (commented):
Sorry for the late review, one suggestion but thank you for the fix!

PR #225, opened 2023-07-06
multidim-interop: Add name ignore option

### LINKED ISSUE #221: Skip testing certain implementations
Nim has a [bug around identify](https://github.com/status-im/nim-libp2p/issues/924) which is breaking js-libp2p CI.

Since the multidim tests are now on the critical path for merging PRs this is something of a showstopper.

Is there a way to temporarily skip testing certain implementations?

MarcoPolo:
Closes #221

### DIFF
--- a/multidim-interop/testplans.ts
+++ b/multidim-interop/testplans.ts
@@ -12,11 +12,11 @@ import path from "path";
     const argv = await yargs(process.argv.slice(2))
         .options({
             'name-filter': {
-                description: 'Only run named test',
+                description: 'Only run tests including this name',
                 default: "",
             },
             'name-ignore': {
-                description: 'Ignore named test',
+                description: 'Do not run any tests including this name ',
                 default: "",
             },
             'emit-only': {
--- a/multidim-interop/testplans.ts
+++ b/multidim-interop/testplans.ts
@@ -15,6 +15,10 @@ import path from "path";
                 description: 'Only run named test',
                 default: "",
             },
+            'name-ignore': {
+                description: 'Ignore named test',
+                default: "",
+            },
             'emit-only': {
                 alias: 'e',
                 description: 'Only print the compose.yaml file',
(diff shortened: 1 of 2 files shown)


multidim-interop/testplans.ts
```diff
@@ -15,6 +15,10 @@ import path from "path";
                 description: 'Only run named test',
                 default: "",
             },
+            'name-ignore': {
+                description: 'Ignore named test',
```
thomaseizinger, line 19:
```suggestion
                description: 'Only run tests including this name',
                default: "",
            },
            'name-ignore': {
                description: 'Do not run any tests including this name ',
```

thomaseizinger (approved):

achingbrain:
Could an example be added to the docs/readme of how to use this option?

MarcoPolo:
> Could an example be added to the docs/readme of how to use this option?

Is the `--help` enough?

```
❯ npm run test -- --help                                   

> Multidimensional libp2p interop test@0.0.1 test
> ts-node testplans.ts --help

Options:
      --name-filter         Only run tests including this name     [default: ""]
      --name-ignore         Do not run any tests including this name
                                                                   [default: ""]
  -e, --emit-only           Only print the compose.yaml file
                                                      [boolean] [default: false]
      --extra-versions-dir  Look for extra versions in this directory. Version
                            files must be in json format  [string] [default: ""]
      --extra-version       Paths to JSON files for additional versions to
                            include in the test matrix     [array] [default: []]
  -h, --help                Show help                                  [boolean]
  ```

MarcoPolo:
Build is failing here because we are running out of space. Maybe time to upgrade to self hosted runners.

achingbrain:
> Is the --help enough?

Forgive me but the help doesn't have an example either?

I don't want to block on this but since this is a hand-rolled test runner docs are very important.

What I mean is "Do not run any tests including this name" is fine if you know what the test name is - is it "ping"?, is it ["nim"](https://github.com/libp2p/test-plans/issues/221)?, is it "rust-v0.51 x node-js-libp2p-head (ws, noise, yamux)"? does it include wildcards? regex?

A short section in the README would answer this question and help out people who aren't familiar with this test suite.

PR #231, opened 2023-07-12
Revert "Install playwright in the base image so it is cached (#226)"

MarcoPolo:
Reverts libp2p/test-plans#226

Looking at this run has us downloading browsers on each interop test: https://github.com/libp2p/test-plans/actions/runs/5535076758/jobs/10100889429. I think what's happening is that playwright will helpfully (though not helpful in this case) download the latest browser. If we have an older browser because we cached it earlier then it will download a new one at run time.

By moving this back to the the post-cache build phase we download the browsers on every build yes, but hopefully avoid downloading them on every run.

I'll look at the CI logs to make sure that's the case.


MarcoPolo:
Yep, looks like this fixes it.

PR #232, opened 2023-07-12
multidim-interop: Update rust-libp2p .gitignore with chromium-image.json

MarcoPolo:
The cache works by mapping a set of inputs to a set out outputs. It fingerprints the inputs by looking at everything in the folder that isn't gitignored. The idea being that if it's ignored it's a build artifact and if it isn't than it is a build input. Since we didn't gitignore chromium-image.json it got treated as an input when we generated the fingerprint. This meant that the cache-key was different before the build and after the build.

I've added a check that will fail CI early if we detect a dirty Git tree after building. That should catch this class of errors.

This should fix the cache misses we've been seeing with rust-libp2p v0.52

### DIFF
--- a/.github/actions/run-interop-ping-test/action.yml
+++ b/.github/actions/run-interop-ping-test/action.yml
@@ -77,6 +77,16 @@ runs:
       run: npm run cache -- load
       shell: bash
 
+    - name: Assert Git tree is clean.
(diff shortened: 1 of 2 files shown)


mxinden (approved):
My bad. Thank you @MarcoPolo for both the fix and the additional CI step.

Issue #229, opened 2023-07-12
Getting close to the 14 GB Actions limit


MarcoPolo:
I think we're getting dangerously close to GitHub Action's 14 GB storage limit on action runners. We should print out stats for the sizes of each image so we can get a feel for the total usage and where it makes sense to trim.

achingbrain:
I see `ENOSPC: no space left on device` errors in when running multidim interop  tests in CI 😢 

Eg: https://github.com/libp2p/js-libp2p/actions/runs/5643526691/job/15285746637

MarcoPolo:
Js will hit this issue more often. @galargh how do we get js-libp2p to run on self hosted runners?

galargh:
We did get JS on self-hosted in https://github.com/libp2p/js-libp2p/pull/1891.

PR #233, opened 2023-07-14
multidim-interop: Prune older versions

MarcoPolo:
We need to clean up older versions because we're getting close to the GHA limit: https://github.com/libp2p/test-plans/issues/229. Zig-libp2p, for example, already fails to build these tests because we run out of space.

js-libp2p is especially heavy since it includes the browsers. The older go-libp2p versions didn't create minimal images, so those are also heavier than they need to be.

I think it's safe to remove these since:
1. They've been interop tested for a while now, unlikely to find anything new.
2. While it's possible new versions slowly break compatibility with older versions, I feel like it's unlikely without also breaking a less old version (i.e. transitively testing). It's not guaranteed, but I think it's an okay tradeoff.


MarcoPolo:
Builds are already failing without this, so I'm going to merge this for now to unblock this repo and other repose as well.

mxinden (commented):
:+1:

Issue #235, opened 2023-07-14
js-libp2p webrtc tests flaky


MarcoPolo:
Related: https://github.com/libp2p/js-libp2p/issues/1846

There's a lot of stuff in the stack, so this needs investigation as to what is being flaky. Is it:
- js-libp2p
- js-libp2p-webrtc
- js-libp2p circuitV2
- js-libp2p p2p-circuit transport
- Browser itself
- Something else?

My hope is that it's not the browser as everything else is easier to fix. Maybe a very simple non-js-libp2p browser only interop test would be a good way to test if the browser is at fault.

MarcoPolo:
cc @maschad

MarcoPolo:
I would use https://playwright.dev directly, avoiding Aegir and playwright-test.

I'll probably do this at some point this week, but if you have bandwidth and it seems interesting, let me know and you can take.

maschad:
Good point, Thanks for clarifying. I will take a look into it this week.

MarcoPolo:
Sounds good. Let's sync up when you get a chance.

maschad:
>  Maybe a very simple non-js-libp2p browser only interop test would be a good way to test if the browser is at fault.

So after syncing with @MarcoPolo , we decided to setup a very simple test case that communicates via `RTCDataChannels` , inspired by [this example](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Simple_RTCDataChannel_sample). I pushed the tests to [this repo](https://github.com/maschad/simple-playwright-webrtc-test)  and I can see there is definitely a discrepancy when running on ubuntu vs macOS for webkit, in this case the fill [element handle](https://playwright.dev/docs/api/class-elementhandle#element-handle-fill) doesn't work [on webkit running on ubuntu](https://github.com/maschad/simple-playwright-webrtc-test/actions/runs/5593484361/jobs/10227176721#step:6:47) but [it works on webkit on mac os](https://github.com/maschad/simple-playwright-webrtc-test/actions/runs/5593620301/jobs/10227473291)

I think these types of discrepancies may be related to some errors in tests but it doesn't explain why a firefox or chrome test may produce a false negative, so doubt it's a browser issue.

sukunrt:
Is there any update here? We are adding support for webrtc in go-libp2p and it'd be nice to test interop with js-libp2p. I have pretty much copied the js-libp2p test implementation: https://github.com/libp2p/go-libp2p/pull/2584

[...]

PR #240, opened 2023-07-19
DO NOT MERGE: testing webrtc using multidim stack

maschad:
Related #235


MarcoPolo:
I don't understand why we're copying so much of the multidim-interop test runner code. Shouldn't this be able to live inside `test-plans/multidim-interop/impl/no-js-libp2p/...`?

simple-webrtc-test/tests/simple-rtc-datachannel.spec.js
```diff
@@ -0,0 +1,31 @@
...
+	await disconnectButton.click();
+});
+
+
+test.describe('Simple RTC DataChannel', () => {
+	test('should send a message from pc1 to pc2', async ({ page }) => {
+	    	const connectButton = page.getByTestId('connectButton');
+		const messageInputBox = page.getByTestId('message');
+		const sendButton = page.getByTestId('sendButton');
+		const receivedBox = page.getByTestId('receivebox');
```
MarcoPolo, line 22: I don't understand. Are these referencing the same page? Are you not testing communication between two browsers?

maschad: If the intention is to set up an RTCDataChannel between two browsers, wouldn't that essentially be a simplistic replication of [the test here](https://github.com/libp2p/js-libp2p/blob/master/packages/transport-webrtc/test/basics.spec.ts#L100)?

MarcoPolo: Yes, that's the goal. Except without any js-libp2p. From the [original issue](https://github.com/libp2p/test-plans/issues/235#issue-1805319171):

> Maybe a very simple non-js-libp2p browser only interop test would be a good way to test if the browser is at fault.

maschad: I understand, but upon reconsideration I wondering what insights that may yield given this test passes consistently. Do you think it would be fruitful to run that test in the multidim interop setup?

MarcoPolo: Does that test run with multiple browsers? Looking at it, it seems like it's run within a single context.

maschad: Gotcha, you are correct it does not run in multiple browsers.

PR #241, opened 2023-07-25
fix(perf): run iperf on TCP and set TCP send/receive window to bdp

mxinden:
This commit makes iperf use TCP instead of UDP. Using UDP allows the sender to send as much data as they can, not as much data as the receiver can handle. Using TCP measures the actual bandwidth of the connection.

This commit sets the TCP send and receive window to the bandwidth-delay-product. In order to exhaust the whole bandwidth between the two nodes, TCP needs to be able to buffer all bytes on the wire in the case of packet loss.

### DIFF
--- a/perf/terraform/modules/long_lived/files/user-data.sh
+++ b/perf/terraform/modules/long_lived/files/user-data.sh
@@ -7,6 +7,14 @@ sudo yum -y install iperf3
 # Bump UDP receive buffer size. See https://github.com/quic-go/quic-go/wiki/UDP-Receive-Buffer-Size.
 sudo sysctl -w net.core.rmem_max=2500000
 
+# Set maximum TCP send and receive window to bandwidth-delay-product.
+#
+# With a bandwidth of 25 Gbit/s per machine and a ping of 60 ms between the two
+# machines, the bandwidth-delay-product is ~178.81 MiB. Set send and receive
+# window to 200 MiB.
+sudo sysctl -w net.ipv4.tcp_rmem='4096 131072 200000000'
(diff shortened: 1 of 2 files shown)


mxinden:
Before (low https with median 301 Mbit/s)

newplot(3)

after (high https with median 781 Mbit/s)

newplot(2)

mxinden:
@MarcoPolo @marten-seemann can one of you give this a review?

perf/terraform/modules/long_lived/files/user-data.sh
```diff
@@ -7,6 +7,14 @@ sudo yum -y install iperf3
 # Bump UDP receive buffer size. See https://github.com/quic-go/quic-go/wiki/UDP-Receive-Buffer-Size.
 sudo sysctl -w net.core.rmem_max=2500000
 
+# Set maximum TCP send and receive window to bandwidth-delay-product.
+#
+# With a bandwidth of 25 Gbit/s per machine and a ping of 60 ms between the two
+# machines, the bandwidth-delay-product is ~178.81 MiB. Set send and receive
+# window to 200 MiB.
+sudo sysctl -w net.ipv4.tcp_rmem='4096 131072 200000000'
+sudo sysctl -w net.ipv4.tcp_wmem='4096 20480 200000000'
```
marten-seemann, line 16: Do I understand correctly that this is set for every run, not just for iperf?

marten-seemann (approved):

perf/terraform/modules/long_lived/files/user-data.sh
mxinden: Correct. This set on startup on both machines and is thus set for all benchmarks, all implementations and all versions.

PR #243, opened 2023-07-26
docs(perf): document how to add a new version of existing impl

mxinden:


mxinden:
Trivial documentation change in `perf/` only. Will go ahead and merge. Happy to address any follow-up reviews.

PR #244, opened 2023-07-26
feat: add js-libp2p perf tests

maschad:
Related https://github.com/libp2p/js-libp2p/pull/1604


mxinden:
Great to see @maschad. Thank you for the work. Let me know once this is ready for a review. Also, in case you don't have permissions to trigger the `libp2p perf` GitHub Action, I am happy to do it for you.

perf/README.md
```diff
@@ -51,7 +51,7 @@ _WARNING_: Running the perf tests might take a while.
           - `--run-server`
       - Running as a libp2p-perf client
           - Input via command line
-            - `--server-ip-address`
+            - `--server-address`
```
mxinden, line 54: Good catch. Thank you!

perf/impl/js-libp2p/v0.45/Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+extract: js-libp2p-${commitSha}.zip
+	unzip -o js-libp2p-${commitSha}.zip
+
+js-libp2p-${commitSha}.zip:
+	wget -O $@ "https://github.com/libp2p/js-libp2p/archive/${commitSha}.zip"
+
+clean:
+	rm -rf js-libp2p-*.zip
+	rm -rf js-libp2p-*
```
mxinden, line 20: How about removing the `perf` binary as well?

perf/impl/js-libp2p/v0.45/Makefile
```diff
@@ -0,0 +1,22 @@
...
+all: perf
+
+perf: build-project
+	cp -r ./js-libp2p-${commitSha}/packages/perf perf-tests
+
+build-project: extract
+	docker run --rm -v "$(shell pwd)/js-libp2p-${commitSha}":/usr/src/myapp -w /usr/src/myapp node:18.17.0 npm install
+	docker run --rm -v "$(shell pwd)/js-libp2p-${commitSha}":/usr/src/myapp -w /usr/src/myapp node:18.17.0 npm run build
+
+extract: js-libp2p-${commitSha}.zip
```
mxinden, line 12: Why use a `PHONY` target even though this target actually creates files, namely the extracted `js-libp2p-${commitSha}` (note the missing `.zip`) folder?

mxinden (commented):
Thank you for the work here. Once https://github.com/libp2p/test-plans/pull/249 is merged I can trigger the CI action.

perf/runner/terraform.tfstate
mxinden, line 1: I am surprised why this file is needed. Shouldn't you have `.tfstate` files in the `configs/` subfolders only? Also, how come this is added even though it is listed in `.gitignore`?

perf/terraform.tfstate
mxinden, line 1: same here

perf/terraform/configs/local/terraform.tf
```diff
@@ -39,7 +39,7 @@ variable "ci_enabled" {
 variable "long_lived_enabled" {
   type        = bool
   description = "Whether or not to create long lived resources (in CI, used across runs; e.g. VPCs)"
-  default     = false
+  default     = true
```
mxinden, line 42: Let's revert this before merge.

perf/impl/js-libp2p/v0.45/Makefile
maschad: The perf bash script is needed since it runs the JS code via node, given that JS is a scripting language the approach is slightly different than general purpose languages which produce a binary executable.

perf/impl/js-libp2p/v0.45/Makefile
maschad: Good point

perf/runner/terraform.tfstate
mxinden: Why did you run `terraform apply` in the `perf/runner` directory? `terraform` is only relevant in the `perf/terraform` directory.

perf/impl/js-libp2p/v0.45/Makefile
mxinden: Makes sense. Thank you.

mxinden:
@maschad please ping me once this is ready for another review.

mxinden:
I triggered the `libp2p perf` GitHub Action for the `libp2p/test-plans` `perf/js-libp2p` branch:

https://github.com/libp2p/test-plans/actions/runs/5818528858/job/15775124342

@maschad do I understand correctly that the branch on `libp2p/test-plans` was created by you? The origin of this pull request is `maschad/test-plans`. How about using `libp2p/test-plans` only by closing this pull request and replacing it with a new one off of `libp2p/test-plans`? That would avoid confusion.

mxinden:
Test run (see link above) failed with:

```
=== Starting client js-libp2p/v0.45/tcp
bash: line 1: ./impl/js-libp2p/v0.45/perf: Permission denied
```

The `perf` file seems to not be marked as executable:

```
ls -la

total 16
drwxrwxr-x 2 mxinden mxinden 4096 Aug 10 10:43 .
drwxrwxr-x 3 mxinden mxinden 4096 Aug 10 10:43 ..
-rw-rw-r-- 1 mxinden mxinden  699 Aug 10 10:43 Makefile
-rw-rw-r-- 1 mxinden mxinden  876 Aug 10 10:43 perf
```

maschad:
> @maschad do I understand correctly that the branch on libp2p/test-plans was created by you? The origin of this pull request is maschad/test-plans. How about using libp2p/test-plans only by closing this pull request and replacing it with a new one off of libp2p/test-plans? That would avoid confusion.

Thanks for running the CI @mxinden I don't have write access to the repo so this is the only branch I have been using i.e. `maschad/test-plans:perf/js-libp2p`

> The perf file seems to not be marked as executable:

Thanks for helping me resolve this, was a file permissions issue, resolved via `chmod +x`

mxinden:
> > The perf file seems to not be marked as executable:
> 
> Thanks for helping me resolve this, was a file permissions issue, resolved via `chmod +x`

But you didn't push the change, right @maschad?

maschad:
> But you didn't push the change, right @maschad?

Correct, I'm debugging another issue and then will push the changes.

perf/terraform/configs/local/terraform.tf
mxinden: This change will make CI fail. It will try to spin up the long-lived resources which already exist in the libp2p AWS account. Mind reverting @maschad?

perf/terraform/modules/short_lived/main.tf
```diff
@@ -19,7 +19,7 @@ resource "aws_instance" "perf" {
 
   launch_template {
     name = "perf-node"
-    version = "3"
+    version = "1"
```
mxinden, line 22: Same here. In the libp2p AWS account the launch template to use is `version` `3`. Mind reverting this change? Otherwise I can not trigger CI.

maschad:
@mxinden I've run this locally and generated these [benchmark results](https://github.com/libp2p/test-plans/pull/244/files#diff-c5b04dce18f8d7b44769534ed718164efaf58396d8f9031af22169b5421f646b) when you have a chance could you please trigger a CI run? Thanks.

maschad:
Once the CI run is successful and [js-libp2p protocol perf 1.1.0 is merged and published](https://github.com/libp2p/js-libp2p/pull/1960) I can follow up with PR that will make builds faster by running the perf package solely (as opposed to installing and building the entire monorepo).

p-shahi:
Can you add js-libp2p v0.46.x here as well please @maschad ?

mxinden:
> @mxinden I've run this locally and generated these [benchmark results](https://github.com/libp2p/test-plans/pull/244/files#diff-c5b04dce18f8d7b44769534ed718164efaf58396d8f9031af22169b5421f646b) when you have a chance could you please trigger a CI run? Thanks.

CI run: https://github.com/libp2p/test-plans/actions/runs/5867985535/job/15909853329

perf/runner/src/index.ts
```diff
@@ -56,7 +56,7 @@ async function main(clientPublicIP: string, serverPublicIP: string) {
 function runPing(clientPublicIP: string, serverPublicIP: string): PingResults {
     console.error(`= run 100 pings from client to server`);
 
-    const cmd = `ssh -o StrictHostKeyChecking=no ec2-user@${clientPublicIP} 'ping -c 100 ${serverPublicIP}'`;
+    const cmd = `ssh -o StrictHostKeyChecking=no ec2-user@${clientPublicIP} 'ping -c 1 ${serverPublicIP}'`;
```
mxinden, line 59:
```suggestion
    const cmd = `ssh -o StrictHostKeyChecking=no ec2-user@${clientPublicIP} 'ping -c 100 ${serverPublicIP}'`;
```

Let's revert this change. Gives installation of `docker` and `iperf` a bit more time.

perf/impl/js-libp2p/v0.45/perf
```diff
@@ -0,0 +1,44 @@
...
+# Check if Node.js is installed
+if [ -z "$node_path" ]; then
+  echo "Node.js is not installed or not in the PATH."
+  # Update package list and install Node.js
+  sudo yum update -y
+  sudo yum install -y gcc-c++ make
+  curl -fsSL https://deb.nodesource.com/setup_18.x | sudo -E bash -
+  sudo yum install -y nodejs
+
+fi
```
mxinden, line 15: Seems like installing NodeJS is failing:

https://github.com/libp2p/test-plans/actions/runs/5868079634/job/15910134917

@maschad are you seeing the same issue when running against your AWS instances?

maschad: This [error](https://github.com/libp2p/test-plans/actions/runs/5868079634/job/15910134917#step:12:4296) is actually when parsing the output of the tests, not related to installing node on the AWS.

I don't get this error when running with my AWS infrastructure so this is strange as I was able to generate the [benchmark results](https://github.com/libp2p/test-plans/pull/244/files#diff-c5b04dce18f8d7b44769534ed718164efaf58396d8f9031af22169b5421f646b)

maschad:
@mxinden I've updated the `Makefile` to use the [npm published package](https://www.npmjs.com/package/@libp2p/perf) so that we can have faster builds, could you trigger a re-run so that we can test it. Also how would I be able to view the server logs that run on these AWS instances should we encounter an error?

perf/impl/js-libp2p/v0.45/perf
maschad: So I was able to reproduce this issue on my instance, it seems as if the runner script is executed before the Node js installation is complete. I wonder if this is the best place to install node or if we should add it to the [terraform template](https://github.com/libp2p/test-plans/blob/perf/js-libp2p/perf/terraform/configs/local/terraform.tf#L16-L30) like:

```
user_data = <<-EOF
              #!/bin/bash
              sudo yum update -y
              sudo yum install -y git nodejs npm
              EOF
```

mxinden: Both the server and the client need to install NodeJS before they can execute the server or client software respectively. Maybe the client is done installing NodeJS faster than the server, then tries to connect to the server, but given that the server is still installing NodeJS, it is not yet ready to receive the request. Could that be the case @maschad?

We could install `nodejs` inside of [`user-data.sh`](https://github.com/libp2p/test-plans/blob/master/perf/terraform/modules/long_lived/files/user-data.sh).

That said, `user-data.sh` is part of the long-lived infrastructure. Thus, when one e.g. wants to change the NodeJS version, it is not as simple as editing `user-data.sh`. Instead one would need to run `terraform apply` in `config/remote` before the change takes effect.

How about installing NodeJS in the `impl/js-libp2p/v0.45` `Makefile` as part of the installation (i.e. `make`) instead. Maybe a hack, but gets the job done. At least for now.

maschad: Good idea @mxinden I've made that change to the `Makefile` in 0a220bfabc2cd13e93aa298d521c31106df02128

maschad:
> libp2p/test-plans/actions/runs/5881769686/job/15950930996

Looks like it passed 🚀 

I've restored the perf defaults and also untracked the [local terraform lock file](https://github.com/libp2p/test-plans/pull/244/files#diff-76346a1e208d9bd0079837dc9e2a604df635662c90c1196fbd641d0ccc6fec49) and [ignored it](https://github.com/libp2p/test-plans/pull/244/files#diff-4dae626f2c899270b1cf664f0b7ca4efe46b107ed375463d6aeeab2862342bb6R3). 

Would you be able to review again please?

perf/terraform/.gitignore
```diff
@@ -1,5 +1,6 @@
 # Local .terraform directories
 **/.terraform/*
+configs/local/.terraform.lock.hcl
```
mxinden, line 3: The lock file's purpose is to ensure that we have consistent `terraform` executions across machines. To ensure that, one needs to check it into version control. Is that not the best practice for terraform lock files?

perf/impl/js-libp2p/v0.46/Makefile
```diff
@@ -0,0 +1,25 @@
...
+perf: install_node
+	mkdir -p $(SOURCE_DIR)
+	$(DOCKER_RUN) npm install $(NPM_PACKAGE_NAME)
+
+install_node:
+	sudo yum update -y
+	curl -fsSL https://rpm.nodesource.com/setup_lts.x | sudo bash -
+	sudo yum install -y nodejs
+	node -v
+	npm -v
```
mxinden, line 21: The `install_node` target alters state beyond this repository folder, i.e. it alters global state, the systems NodeJS installation. In addition it assumes that it is running on a RPM based Linux distribution. Neither is ideal. As a user of the libp2p perf tooling when I run `cd perf/ && make` I expect all changes to be limited to this folder (ignoring caches). As a user of the libp2p perf tooling, I expect to be able to run `cd perf/ && make` ideally on a Mac, though at least on any Linux distribution, not just the RPM ones.

Can we do better? Ideas that come to my mind:

- We could run the server and client in Docker. Though that might have a performance impact which is the reason we run all the other implementations outside of Docker.
- Install NodeJS globally on the machine (`user-data.sh`). Downside is that different js-libp2p versions can not use different NodeJS versions.
- Use tooling like [pkg](https://github.com/vercel/pkg).

I have little to no experience with packaging NodeJS applications. @maschad @achingbrain @MarcoPolo do you have better ideas?

maschad: Well the assumption is that this `Makefile` is intended to be run on the Terraform infrastructure which at the moment is AWS, in this specific case, EC2 instances. This is also the same premise for the `perf` bash script which calls [upon node to start the server/client](https://github.com/libp2p/test-plans/pull/244/files#diff-d634d23030fcd1a8ce7d7daa1cc2431802c5ff307bd84652786e9059925022c7R33).

These `Makefile`s across the implementations aren't really useful outside of the use case of [running the benchmark results](https://github.com/libp2p/test-plans/blob/perf/js-libp2p/perf/runner/src/index.ts#L219), which would rely on the aforementioned Terraform infrastructure, and so I don't see a rationale for a user running `cd perf/impl/ && make` locally.

Thus tailoring the `Makefile` to an RPM Linux distro and altering global state should be fine imo considering these instances will be destroyed once the infrastructure is tore down after a test run.

perf/terraform/.gitignore
maschad: Terraform automatically creates or updates the dependency lock file each time [the terraform init command](https://developer.hashicorp.com/terraform/cli/commands/init) is run and whilst it is recommended to version it, it is for the use case of discussing potential changes to external dependencies - which would be dictated by the [remote terraform configuration](https://github.com/libp2p/test-plans/blob/master/perf/terraform/configs/remote/.terraform.lock.hcl) as opposed one's changes on their local machine, so versioning the local configuration lock file just makes the PR more noisy imo.

That being said this isn't a strong opinion if you feel there is a benefit in reviewing these changes.

perf/impl/js-libp2p/v0.46/Makefile
mxinden, line 21:
```suggestion
perf:
	mkdir -p $(SOURCE_DIR)
	$(DOCKER_RUN) npm install $(NPM_PACKAGE_NAME)
```

mxinden: Installed with https://github.com/libp2p/test-plans/pull/266 now.

mxinden:
Perf run using pre-installed NodeJS installation:

https://github.com/libp2p/test-plans/actions/runs/5902023806/job/16009253523

maschad:
> Perf run with pre-installed NodeJS was successful. Only thing missing is addressing the outstanding conversations.

Thanks for running this @mxinden , I've addressed the other comments.

perf/impl/js-libp2p/v0.46/Makefile
```diff
@@ -0,0 +1,18 @@
...
+DOCKER_IMAGE := node:18.17.1
+DOCKER_RUN := docker run --rm -v "$(shell pwd)/$(SOURCE_DIR)":/usr/src/myapp -w /usr/src/myapp $(DOCKER_IMAGE)
+
+all: perf
+
+perf:
+	mkdir -p $(SOURCE_DIR)
+	$(DOCKER_RUN) npm install $(NPM_PACKAGE_NAME)
+
+clean: rm -rf js-libp2p-*
```
mxinden, line 16:
```suggestion
clean:
	-rm -rf js-libp2p-*
```

`rm` is not another target but a command.

perf/impl/js-libp2p/v0.46/Makefile
```diff
@@ -0,0 +1,18 @@
...
+
+.PHONY: all clean perf install_node
```
mxinden, line 18:
```suggestion
.PHONY: all clean perf
```

mxinden (commented):
Two more comments. Otherwise this looks good to me.

mxinden:
Another workflow run, now that all other implementations are re-enabled.

https://github.com/libp2p/test-plans/actions/runs/5906112245/job/16021494862

maschad:
> Another workflow run, now that all other implementations are re-enabled.
> 
> [libp2p/test-plans/actions/runs/5906112245/job/16021494862](https://github.com/libp2p/test-plans/actions/runs/5906112245/job/16021494862)

This workflow failed because [node was not installed](https://github.com/libp2p/test-plans/actions/runs/5906112245/job/16021494862#step:12:7393) 

looking through it seems like the [aforementioned successful run](https://github.com/libp2p/test-plans/pull/244#issuecomment-1683767701) was still relying on the [Makefile's NodeJS installation](https://github.com/libp2p/test-plans/commit/0a220bfabc2cd13e93aa298d521c31106df02128#diff-3ad47321b6b446cfa388772547c901727bab82867fba727eee828999d979a9a4R12-R21)  given that the successful run's i.e. 2cb6950 parent commit is 0a220bf **instead of** f57408d

@mxinden I am still comfortable using the Makefile to install NodeJS for reasons outlined [here](https://github.com/libp2p/test-plans/pull/244#discussion_r1296391261). Should we need to isolate Node environments for perf runs I think that could be a in a follow up PR. What do you think?

p-shahi:
> > Another workflow run, now that all other implementations are re-enabled.
> > [libp2p/test-plans/actions/runs/5906112245/job/16021494862](https://github.com/libp2p/test-plans/actions/runs/5906112245/job/16021494862)
> 
> This workflow failed because [node was not installed](https://github.com/libp2p/test-plans/actions/runs/5906112245/job/16021494862#step:12:7393)
> 
> looking through it seems like the [aforementioned successful run](https://github.com/libp2p/test-plans/pull/244#issuecomment-1683767701) was still relying on the [Makefile's NodeJS installation](https://github.com/libp2p/test-plans/commit/0a220bfabc2cd13e93aa298d521c31106df02128#diff-3ad47321b6b446cfa388772547c901727bab82867fba727eee828999d979a9a4R12-R21) given that the successful run's i.e. [2cb6950](https://github.com/libp2p/test-plans/commit/2cb69509409eb4fbb8084a46a3ef473e7b7df569) parent commit is [0a220bf](https://github.com/libp2p/test-plans/commit/0a220bfabc2cd13e93aa298d521c31106df02128) **instead of** [f57408d](https://github.com/libp2p/test-plans/commit/f57408de37ff809d03a0c60607e7685e7156f00f)
> 
> @mxinden I am still comfortable using the Makefile to install NodeJS for reasons outlined [here](https://github.com/libp2p/test-plans/pull/244#discussion_r1296391261). Should we need to isolate Node environments for perf runs I think that could be a in a follow up PR. What do you think?

@maschad I agree w/ @mxinden  that altering global state on individual test runs is not ideal, it seems like an antipattern. Installing node as a part of perf/terraform/modules/long_lived/files/user-data.sh seems right, can you investigate why it's not installed as a part of user-data.sh

maschad:
@p-shahi @mxinden  I was able to successfully run the tests on my machine using [yum to install nodejs](https://github.com/libp2p/test-plans/pull/244/commits/e7aee86ed07a52e68cb7a59f70b1c0b429310d97#diff-b83de3058c52917c778f6327b2f2165e044bacee5e9b90167fd57f760c5e5ea2R27) as opposed to `nvm` , would you be able to do a re-run?

p-shahi:
Triggered run: https://github.com/libp2p/test-plans/actions/runs/5907686611/job/16026013565

maschad:
> retriggered due to unrelated failure: [libp2p/test-plans/actions/runs/5907686611/job/16026286429](https://github.com/libp2p/test-plans/actions/runs/5907686611/job/16026286429)

Looks like only the [short-lived infra was created](https://github.com/libp2p/test-plans/actions/runs/5907686611/job/16026286429#step:8:148) but the [user-data.sh changes](https://github.com/libp2p/test-plans/pull/244/files#diff-b83de3058c52917c778f6327b2f2165e044bacee5e9b90167fd57f760c5e5ea2R26-R27) in e7aee86ed07a52e68cb7a59f70b1c0b429310d97 are to _long lived_ infrastructure @mxinden how do you normally ensure that the changes to long lived infrastructure are applied in CI runs?

perf/terraform/modules/short_lived/main.tf
```diff
@@ -22,6 +22,9 @@ resource "aws_instance" "perf" {
   }
 
   key_name = aws_key_pair.perf.key_name
+
+  user_data_replace_on_change = true
+
```
mxinden, line 27:
```suggestion
```

p-shahi:
@mxinden I saw you pushed some changes, did you execute a perf run as well and is it approved for merge now?

mxinden:
The NodeJS installation issue is resolved with https://github.com/libp2p/test-plans/pull/267. My bad for the broken https://github.com/libp2p/test-plans/pull/266.

> I saw you pushed some changes

I reverted your changes to the NodeJS installation instructions. They are no longer needed.

NodeJS installation is now properly in place. See js-libp2p perf test being executed in https://github.com/libp2p/test-plans/actions/runs/5923420241/job/16060336976.

---

Unfortunately perf runs are failing consistently, namely on `rust-libp2p/v0.52/tcp`. See e.g. https://github.com/libp2p/test-plans/actions/runs/5923420241/job/16060336976.

The rust-libp2p server is unable to bind to the `4001` TCP port:

```
$ cat server.log 
Error: 

Caused by:
    0: Address already in use (os error 98)
    1: Address already in use (os error 98)
```

and thus the Rust client fails to connect:

```
[2023-08-21T09:17:05.031Z INFO  perf] start benchmark: custom
Error: Outgoing connection error to None: Transport([("/ip4/54.212.203.99/tcp/4001", Other(Custom { kind: Other, error: Transport(Right(Left(Right(Apply(ClientUpgrade(Custom { kind: Other, error: Failed })))))) }))])
```

The problem is, that the js-libp2p NodeJS process is still running and binded to `4001` TCP:

```
$ sudo netstat -tulnep | grep 4001
tcp        0      0 0.0.0.0:4001            0.0.0.0:*               LISTEN      1000       96570      72416/node          
```

@maschad could it be that the js-libp2p server is not reacting to `kill $PID`?

maschad:
> @maschad could it be that the js-libp2p server is not reacting to kill $PID?
 
 That's correct. The `node` process has a separate `PID` from the bash executable, and it was the bash script's `PID` being written to  _pidfile_ and subsequently being killed. 
 
 I have modified the [runner to kill the process running on **port 4001**](https://github.com/libp2p/test-plans/pull/244/files#diff-0dcf0f7d1009b89017414bcafde3e81fad7b8ef0055b13e6fe9351c37ec0dbe5R128) as an alternative, rather than writing the **PID** to a _pidfile_, then reading that PID and then killing it. I have ran this a few times on my machine successfully.

p-shahi:
Results of run: https://github.com/libp2p/test-plans/actions/runs/5932647404
benchmark results committed to perf/js-libp2p branch: https://github.com/libp2p/test-plans/commit/84295bca741d93f008bf937ad823e581dc617ac3

p-shahi:
Is the ready for merge?

> > @maschad could it be that the js-libp2p server is not reacting to kill $PID?
> 
> That's correct. The `node` process has a separate `PID` from the bash executable, and it was the bash script's `PID` being written to _pidfile_ and subsequently being killed.
> 
> I have modified the [runner to kill the process running on **port 4001**](https://github.com/libp2p/test-plans/pull/244/files#diff-0dcf0f7d1009b89017414bcafde3e81fad7b8ef0055b13e6fe9351c37ec0dbe5R128) as an alternative, rather than writing the **PID** to a _pidfile_, then reading that PID and then killing it. I have ran this a few times on my machine successfully.

drive by comment here: why not get the PID of node process within the perf shell script itself? seems better than assuming that we have to kill the process listening on port 4001 - that could change in the future?

mxinden:
I suggest containing the cleanup of the NodeJS process within the `perf` bash script. I implemented this in https://github.com/libp2p/test-plans/commit/6d4c137530670ca595112744a97f64c5fb0eef4d. It is currently running on https://github.com/libp2p/test-plans/actions/runs/5939464620/job/16105895040. Unless @maschad you object I will push that commit here on green.

maschad:
> Is the ready for merge?
> 
> > >

pending an approval from @mxinden I would say yes

> drive by comment here: why not get the PID of node process within the perf shell script itself? seems better than assuming that we have to kill the process listening on port 4001 - that could change in the future?

That was my original course of action but it's not an assumption since the port is set by the [nohup command that runs the server](https://github.com/libp2p/test-plans/pull/244/files#diff-0dcf0f7d1009b89017414bcafde3e81fad7b8ef0055b13e6fe9351c37ec0dbe5R132) and it's unlikely that would change.

It's more efficient to kill the process running at the port since that was pre-determined to be the server and not another process, as opposed to writing it to a file and then reading that file, then deleting it.

maschad:
> I suggest containing the cleanup of the NodeJS process within the `perf` bash script. I implemented this in [6d4c137](https://github.com/libp2p/test-plans/commit/6d4c137530670ca595112744a97f64c5fb0eef4d). It is currently running on [libp2p/test-plans/actions/runs/5939464620/job/16105895040](https://github.com/libp2p/test-plans/actions/runs/5939464620/job/16105895040). Unless @maschad you object I will push that commit here on green.

I spoke to @p-shahi on the maintainers call on 22-08-23 and he explained why that's his preference, I don't object to it so please feel free to go ahead and push that commit in order that this can be merged.

maschad:
> Latest perf run: [libp2p/test-plans/actions/runs/5949260069/job/16134709657](https://github.com/libp2p/test-plans/actions/runs/5949260069/job/16134709657)

Thanks @mxinden looks good. This is ready for merge from my end.

mxinden:
To view the test results on the dashboard:

https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=perf%2Fjs-libp2p#branch

mxinden:
> To view the test results on the dashboard:
> 
> https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=perf%2Fjs-libp2p#branch

Latency median of 1s seems off. Is this expected @maschad? Maybe a unit conversion error?

maschad:
> Latency median of 1s seems off. Is this expected @maschad? Maybe a unit conversion error?

This has been what I have observed in my tests runs, on average I have seen latencies of `1.03` seconds when running the [test plans perf benchmark](https://github.com/libp2p/test-plans/blob/master/perf/runner/src/index.ts#L33-L41) on my machine and I think the issue [may be related to known performance issues with our yamux implementation](https://github.com/ChainSafe/js-libp2p-yamux/issues/42)

I don't think it's a conversion error given we the [conversion from milliseconds to seconds](https://github.com/libp2p/js-libp2p/blob/master/packages/protocol-perf/src/main.ts#L95) is correct.

I've ran the [perf package](https://github.com/libp2p/js-libp2p/tree/master/packages/protocol-perf) locally in two terminals over 100 iterations and averaged `0.02` seconds but that of course excludes connection establishment latency.

mxinden:
Adding @achingbrain and @MarcoPolo here for visibility.

**Summary: we are measuring a median of 1s when establishing a js-libp2p TCP+Plaintext+Yamux connection and sending an empty request and receive an empty response.** For comparison, the median latency between the machines is 60ms and an https (TCP+TLS) request takes median 180ms (3 round-trips).

See dashboard below for higher resolution.

https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=perf%2Fjs-libp2p#branch

---

Moving forward here anyways. In my eyes measuring in itself already provides value.

Any related fixes can go into follow-up pull requests.

PR #253, opened 2023-08-10
chore: rm rust-libp2p v0.49 and v0.50, update v0.52

p-shahi:
I'm not sure if we should still test v0.49 and v0.50 given that they are quite old (from 2022)
0.51 and 0.52 have been out for some time now.
Wdyt @mxinden @thomaseizinger? If there are notable/many rust-libp2p users still relying on 0.50 or 0.49, we should keep them


mxinden (approved):
Looks good to me. Thank you for porting our `Makefile` to the new `version.lock` mechanism @p-shahi!

[Substrate is on `v0.52.1`](https://github.com/paritytech/substrate/blob/master/client/network/Cargo.toml#L28). [Lighthouse is upgrading to `v0.52`](https://github.com/sigp/lighthouse/pull/4087). [Most downloaded version is `v0.52.1`](https://crates.io/crates/libp2p).

p-shahi:
@mxinden I pushed some more fixes to the makefile. it seems like it's failing now as it doesn't detect webrtc-direct as a transport? I haven't looked further but tagging you in case you've seen something like this before.

multidim-interop/impl/rust/v0.51/Makefile
mxinden, line 1: I investigated a bit.

The `v0.51.3` tag points (as it should) to the commit we cut the `libp2p` `v0.51.3` release off of. Since then we have cut many patch releases of sub-crates. But those patch releases of sub-crates don't require an additional patch release of the meta crate (i.e. `libp2p`). Downstream users can upgrade their sub-crate version without having to upgrade the meta crate version.

The previously specified `commitSha` pointed to the latest commit compatible with the `libp2p` `v0.51.3` release. I.e. it contained all sub-crate patch releases. Thus CI was green.

The `0.51.3` git tag only points to the much older `v0.51.3` release commit. It does not contain the patches of the sub-crates. Thus CI fails.

You can see the difference here.

https://github.com/libp2p/rust-libp2p/compare/3c5940aeadb9ed8527b6f7aa158797359085293d...8c26c617b94d992c5fbd91efc73049c838740df1

We could cut `v0.51.4`. But I am hesitant to do so for CI only.

Is the move from `commitSha` to `version` relevant? If not, I suggest we don't do it for `v0.51.3`. In other words I suggest leaving this file untouched.

multidim-interop/impl/rust/v0.52/Makefile
```diff
@@ -1,23 +1,35 @@
 image_name := rust-v0.52
-commitSha := 68e6bf9c3cd0d3317415a5ba31a62e91cba0a5d2
+version := 0.52.1
```
mxinden, line 2:
```suggestion
version := 0.52.2
```

I released `libp2p` `v0.52.2` today. Want to include it in this pull request?

https://github.com/libp2p/rust-libp2p/pull/4312

No worries in case you don't want to mix these efforts in the same PR. I am happy to do a follow-up pull request.

mxinden (commented):
Referencing a single version instead of a commit sha is a bit tricky with rust-libp2p's ability to release patch releases of sub-crates. I am sorry for not catching this earlier @p-shahi. Suggestions below.

multidim-interop/impl/rust/v0.52/Makefile
p-shahi: I'll include 0.52.2 in here 👍

multidim-interop/impl/rust/v0.51/Makefile
p-shahi: > Is the move from commitSha to version relevant? If not, I suggest we don't do it for v0.51.3. In other words I suggest leaving this file untouched.

I am ok with that

mxinden (approved):
Thanks @p-shahi! Feel free to merge once CI is green.

PR #254, opened 2023-08-11
fix(perf): increase maximum udp send buffer size

mxinden:
Previously only the maximum udp **receive** buffer size was increased to `2500000` bytes. With this commit both the send and the receive buffer size is increase.

### DIFF
--- a/perf/runner/benchmark-results.json
+++ b/perf/runner/benchmark-results.json
@@ -7,34 +7,34 @@
         {
           "result": [
             {
-              "latency": 1.112042404
+              "latency": 1.010132731
             },
             {
-              "latency": 1.040904815
+              "latency": 1.015080006
             },
             {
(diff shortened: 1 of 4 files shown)


mxinden:
Change has no significant impact on throughput.

**without this pull request**

newplot(8)

https://observablehq.com/@libp2p-workspace/performance-dashboard

**with this pull request**

newplot(9)

https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=perf-udp-send-buffer

Anyways. I am fine moving forward here, i.e. following the suggestion by quic-go increasing the maximum udp send buffer size.

@marten-seemann want to take a look as well?

marten-seemann:
This makes sense, assuming that we're running with CAP_NET_ADMIN: https://github.com/quic-go/quic-go/issues/3795.

PR #257, opened 2023-08-14
fix(perf): use `$Latest` launch template

mxinden:
Our perf terraform setup differentiates in long lived and short lived resources.

On CI, our long lived resources are spun up once and our short lived resources are spun up on each CI run. From time to time we have to adjust the long lived resources. End result is a new launch template that needs to be referenced in the short lived resources by version.

Next to our CI, the perf terraform setup can as well be used on personal AWS accounts. Their long lived launch template version likely doesn't match the configured launch template version of the short lived aws_instance.

Instead of specifying a specific version, instruct terraform to use the latest, thus supporting both our CI and personal AWS account use-case.

See also terraform `aws_instance` specification.

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#launch-template-specification

### DIFF
--- a/perf/runner/benchmark-results.json
+++ b/perf/runner/benchmark-results.json
@@ -7,34 +7,34 @@
         {
           "result": [
             {
-              "latency": 1.071285577
+              "latency": 1.028831684
             },
             {
-              "latency": 1.069395315
+              "latency": 1.086669124
             },
             {
-              "latency": 1.042424524
+              "latency": 1.071025996
             },
             {
-              "latency": 1.056611303
+              "latency": 1.095323838
             },
             {
-              "latency": 1.071165158
+              "latency": 1.006762935
             },
             {
-              "latency": 1.04385611
+              "latency": 1.079348034
             },
             {
-              "latency": 1.063371162
+              "latency": 1.082299325
             },
             {
-              "latency": 1.083406849
+              "latency": 1.099030779
             },
             {
-              "latency": 1.099260456
+              "latency": 1.03714172
             },
             {
-              "latency": 1.015545856
+              "latency": 1.076451765
             }
           ],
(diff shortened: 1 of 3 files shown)


mxinden:
//CC @marten-seemann this would make your life a bit easier, i.e. not require the diff below:

``` diff
modified   perf/terraform/modules/short_lived/main.tf
@@ -19,7 +19,7 @@ resource "aws_instance" "perf" {
 
   launch_template {
     name = "perf-node"
-    version = "4"
+    version = "1"
   }
 
   key_name = aws_key_pair.perf.key_name
```

mxinden:
@galargh do you have alternative suggestions?

mxinden:
Github Action _perf_ run:

https://github.com/libp2p/test-plans/actions/runs/5853132413/job/15866553674

galargh:
I see it was added here https://github.com/libp2p/test-plans/pull/241 and before that we were defaulting to the `$Default` version (as per https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#launch-template-specification).

The other option would be to set [update_default_version](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#update_default_version) to true on [aws_launch_template.perf](https://github.com/libp2p/test-plans/blob/142557b9ebbe5353047e3a7ea42e770148953ca8/perf/terraform/modules/long_lived/main.tf#L94), and then use the `$Default` here again. 

It'd be pretty much the same thing though so it's just a matter of preference.

galargh (approved):

mxinden:
Today I learned about `update_default_version`. Changed to it. Thank you @galargh.

Triggered run: https://github.com/libp2p/test-plans/actions/runs/5889989158/job/15974217657

Issue #222, opened 2023-07-05
Benchmarking feedback/notes


MarcoPolo:
Spent a bit of time looking at some parts of the benchmarking setup, and had a couple of notes and comments:

* I think you're using iperf wrong. You're using the data from the sender, you should be looking at the receiver. Notice how the Bitrate is very different for sender vs receiver in this example:
```
$ iperf -c 127.0.0.1 -u -b 10g
Connecting to host 127.0.0.1, port 5201
[  5] local 127.0.0.1 port 50191 connected to 127.0.0.1 port 5201
[ ID] Interval           Transfer     Bitrate         Total Datagrams
[  5]   0.00-1.00   sec  1.16 GBytes  10.0 Gbits/sec  38132  
[  5]   1.00-2.00   sec  1.16 GBytes  10.0 Gbits/sec  38161  
[  5]   2.00-3.00   sec  1.16 GBytes  9.99 Gbits/sec  38106  
[  5]   3.00-4.00   sec  1.17 GBytes  10.0 Gbits/sec  38184  
[  5]   4.00-5.00   sec  1.16 GBytes  10.0 Gbits/sec  38151  
[  5]   5.00-6.00   sec  1.16 GBytes  10.0 Gbits/sec  38143  
[  5]   6.00-7.00   sec  1.16 GBytes  9.99 Gbits/sec  38114  
[  5]   7.00-8.00   sec  1.16 GBytes  10.0 Gbits/sec  38165  
[  5]   8.00-9.00   sec  1.16 GBytes  9.99 Gbits/sec  38140  
[  5]   9.00-10.00  sec  1.16 GBytes  10.0 Gbits/sec  38169  
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[  5]   0.00-10.00  sec  11.6 GBytes  10.0 Gbits/sec  0.000 ms  0/381465 (0%)  sender
[  5]   0.00-10.00  sec  8.43 GBytes  7.24 Gbits/sec  0.023 ms  105024/381411 (28%)  receiver
```

You need to use the bitrate on the receiver side. The sender can push as much data as you want, but for these measurements we care about the data that was actually received. Look at the difference here: https://github.com/libp2p/test-plans/actions/runs/5466146370/jobs/9950640038#step:12:29
[...]

PR #258, opened 2023-08-14
feat(perf): Added new CLI flags for iterations + updated docs

maschad:
While integrating [the js impl of perf](https://github.com/libp2p/js-libp2p/pull/1604) I encountered some obstacles in running this infrastructure as well as adding and testing the _perf_ implementation. These are some improvements I thought would help new users of _perf_

Related https://github.com/libp2p/test-plans/issues/222

### DIFF
--- a/perf/README.md
+++ b/perf/README.md
@@ -15,24 +15,49 @@ Benchmark results can be visualized with https://observablehq.com/@libp2p-worksp
 3. Wait for action run to finish and to push a commit to your branch.
 4. Visualize results on https://observablehq.com/@libp2p-workspace/performance-dashboard.
 
-## Running via local machine
+## Running with Terraform on AWS manually
 
+### Pre-requisites
+
+- Terraform 1.5.5 or later
+- Node.js 18 or later
+- [an AWS IAM user](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users.html)
+##
 ### Provision infrastructure
 
-1. Save your public SSH key as the file `./terraform/modules/short_lived/files/perf.pub`; or generate a new key pair with `make ssh-keygen` and add it to your SSH agent with `make ssh-add`.
-2. `cd terraform/configs/local`
-3. `terraform init`
-4. `terraform apply`
-5. `CLIENT_IP=$(terraform output -raw client_ip)`
-6. `SERVER_IP=$(terraform output -raw server_ip)`
+1. Ensure you ssh agent is running, you can check that via
+```bash
+   eval "$(ssh-agent -s)"
+```
+2. Save your public SSH key as the file `./terraform/modules/short_lived/files/perf.pub`; or generate a new key pair with `make ssh-keygen` and add it to your SSH agent with `make ssh-add`.
+3. Change to the terraform local directory ```cd terraform/configs/local``` and initalize terraform ```terraform init```
+4. `terraform apply` A terraform plan will be generated, type `yes` to continue. -To preview the plan, you can run `terraform plan`
+5. Once that is finished, you can get the public IP addresses of the client and server with `terraform output client_ip` and `terraform output server_ip`. Set these to the environment variables via:
+   1. `CLIENT_IP=$(terraform output -raw client_ip)`
+   2. `SERVER_IP=$(terraform output -raw server_ip)`
 
-### Build and run implementations
+**Notes**
+- You may need to reset the infrastructure if you encounter any errors, you can do that by running `terraform destroy` and then `terraform apply`.
+- While running terraform you may encounter the following error:
+  ```bash
+    Error: collecting instance settings: reading EC2 Launch Template versions: couldn't find resource
+    │
+    │   with module.short_lived_server[0].aws_instance.perf,
+    │   on ../../modules/short_lived/main.tf line 15, in resource "aws_instance" "perf":
+    │   15: resource "aws_instance" "perf" {
+  ```
+  If you set *TF_VAR* [`long_lived_enabled`](./terraform/configs/local/terraform.tf#L42) env variable to default to **true** terraform should spin up the long-lived resources that are required for the short-lived resources to be created.
 
-_WARNING_: Running the perf tests might take a while.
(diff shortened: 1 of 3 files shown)


perf/README.md
```diff
@@ -15,24 +15,49 @@ Benchmark results can be visualized with https://observablehq.com/@libp2p-worksp
...
-1. Save your public SSH key as the file `./terraform/modules/short_lived/files/perf.pub`; or generate a new key pair with `make ssh-keygen` and add it to your SSH agent with `make ssh-add`.
-2. `cd terraform/configs/local`
-3. `terraform init`
-4. `terraform apply`
-5. `CLIENT_IP=$(terraform output -raw client_ip)`
-6. `SERVER_IP=$(terraform output -raw server_ip)`
+1. Ensure you ssh agent is running, you can check that via
+```bash
+   eval "$(ssh-agent -s)"
+```
```
marten-seemann, line 31: I don't think SSH requires any background tasks.

p-shahi (commented):
drive by review

perf/README.md
maschad: If I try to run [make ssh-add](https://github.com/libp2p/test-plans/blob/master/perf/Makefile#L4) without running an ssh-agent I will get the error

```
Could not add identity "./terraform/modules/short_lived/files/perf": agent refused operation
make: *** [ssh-add] Error 1
```

marten-seemann: It’s not a requirement for using SSH.
I don’t think we need to point out utterly trivial things.

perf/README.md
```diff
@@ -26,13 +33,28 @@ Benchmark results can be visualized with https://observablehq.com/@libp2p-worksp
...
+
+### Build and run libp2p implementations
 
-_WARNING_: Running the perf tests might take a while.
+Given you have provisioned your infrastructure, you can now build and run the libp2p implementations on the AWS instances.
 
 1. `cd runner`
 2. `npm ci`
-3. `npm run start -- --client-public-ip $CLIENT_IP --server-public-ip $SERVER_IP`
+3.  `npm run start -- --client-public-ip $CLIENT_IP --server-public-ip $SERVER_IP --iterations 50`
```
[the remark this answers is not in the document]
maschad: The intention was to highlight the usage of the `iterations` flag, but given that's [now addressed in a note](https://github.com/libp2p/test-plans/pull/258/files#diff-f67fd5d0f863534b4c355fab5e18cb17656ef46a4bb892f1fd885a36ce01e3e5R57) this probably isn't that useful anymore.

p-shahi (approved):

perf/runner/src/index.ts
```diff
@@ -37,7 +37,7 @@ async function main(clientPublicIP: string, serverPublicIP: string) {
                  uploadBytes: 1,
                  downloadBytes: 1,
                  unit: "s",
-                 iterations: 100,
+                 iterations: iterations * 10,
```
mxinden, line 40: I don't think this is intuitive. When I set `--iterations 10` I don't expect `100` iterations to run instead.

perf/README.md
```diff
@@ -15,7 +15,14 @@ Benchmark results can be visualized with https://observablehq.com/@libp2p-worksp
 3. Wait for action run to finish and to push a commit to your branch.
 4. Visualize results on https://observablehq.com/@libp2p-workspace/performance-dashboard.
 
-## Running via local machine
+## Running with Terraform on AWS manually
+
+### Prerequisites
+
+- Terraform 1.5.5 or later
```
mxinden, line 22: Note that I am using `terraform` `v1.5.4` without issues thus far.

perf/README.md
```diff
@@ -26,13 +33,28 @@ Benchmark results can be visualized with https://observablehq.com/@libp2p-worksp
 5. `CLIENT_IP=$(terraform output -raw client_ip)`
 6. `SERVER_IP=$(terraform output -raw server_ip)`
 
-### Build and run implementations
+**Notes**
+- You may need to reset the infrastructure if you encounter any errors, you can do that by running `terraform destroy` and then `terraform apply`.
```
mxinden, line 37: I don't think this is a good general hint. In the most cases terraform should be able to reconciliate any changes without a previous `terraform destroy`. On a high level, I am hesitant to give general terraform advice here. I think that is much better placed in the official terraform documentation.

perf/README.md
```diff
@@ -41,27 +63,28 @@ _WARNING_: Running the perf tests might take a while.
...
-            - `--transport` (see `runner/versions.ts` for possible variants)
+            - `--server-address`
+            - `--transport` (see [`runner/versions.ts`](./runner/src/versions.ts#L7-L43) for possible variants)
             - `--upload-bytes` number of bytes to upload per stream.
             - `--download-bytes` number of bytes to download per stream.
           - Output
-            - Logging MUST go to stderr.
-            - Measurement output is printed to stdout as JSON in the form of:
+            - Logging MUST go to `stderr`.
+            - Measurement output is printed to **stdout** as JSON in the form of:
```
mxinden, line 83: Nit pick. Either use monospace or bold.

perf/README.md
```diff
@@ -26,13 +33,28 @@ Benchmark results can be visualized with https://observablehq.com/@libp2p-worksp
...
+**Notes**
+- You may need to reset the infrastructure if you encounter any errors, you can do that by running `terraform destroy` and then `terraform apply`.
+- While running terraform you may encounter the following error:
+  ```bash
+    Error: collecting instance settings: reading EC2 Launch Template versions: couldn't find resource
+    │
+    │   with module.short_lived_server[0].aws_instance.perf,
+    │   on ../../modules/short_lived/main.tf line 15, in resource "aws_instance" "perf":
+    │   15: resource "aws_instance" "perf" {
+  ```
```
mxinden, line 45: Let's add some context here along the lines of the below.

> This implies that you haven't deployed the long-living infrastructure (e.g. launch template) on your AWS account. To do so along with each short-lived deployment, you can set the `long_lived_enabled` ...

mxinden (commented):
I don't think this pull request was ready for merge. This is partially on me, due to my delayed review.

I don't think the `--iterations` flag is intuitive. See comment below. Also it is likely going away again with https://github.com/libp2p/test-plans/issues/261, as upload and download bandwidth tests will likely only run as a single iteration.

mxinden:
How about replacing the `--iterations` flag with a `--testing` flag, which would reduce all benchmarks (ping, iperf, ...) to run a single time only. That would remove the counterintuitive 100 iterations on `--iterations 10` and is future proof for https://github.com/libp2p/test-plans/issues/261.

PR #267, opened 2023-08-21
fix(perf): install nvm for ec2-user instead of root

mxinden:
On root the nvm install script can not find a shell config.

```
=> Profile not found. Tried ~/.bashrc, ~/.bash_profile, ~/.zprofile, ~/.zshrc, and ~/.profile.
```


mxinden:
`perf/` related only. Works on https://github.com/libp2p/test-plans/pull/244/. Merging here.

PR #270, opened 2023-08-22
ci: only run interop tests on changes to multidim-interop dir

p-shahi:
Only run interop tests on changes to the multidim-interop/ directory (no use running them on changes to perf directory)


galargh (approved):

mxinden:
Cool. Thank you @p-shahi.

PR #276, opened 2023-08-24
feat(perf): continuosly measure on single conn (iperf-style)

### LINKED ISSUE #261: perf: throughput test (TCP, QUIC, libp2p, but not iperf) never exits slow start
Moving a conversation from Slack here: https://filecoinproject.slack.com/archives/C03K82MU486/p1692012437887379

I spent some time playing around with the perf setup. The fact that iPerf is so much faster than everything else really bothered me. iPerf is a sanity check for our setup, and if the value measured for iPerf differs significantly from the HTTPS value we measure, something is wrong in the test setup.

It turns out that what we’re seeing here is not a performance measurement at all, but purely [Reno / Cubic slow start](https://en.wikipedia.org/wiki/TCP_congestion_control#Slow_start). Reno starts with a congestion window of 10 packets, and increases the cwnd by 1 packet for every ACK received, effectively leading to a doubling of the cwnd every RTT. Slow start is only exited once the first packet loss (or ECN marking) occurs.
I created a spreadsheet with a simple back-of-the-envelope calculation: https://docs.google.com/spreadsheets/d/1LYVOh3kkefhD5t-rR0JEgPJ-YqKpqaXmUCYu_Z-t0xg/edit?usp=sharing

To send 100 MB (the amount of data transferred in our tests) in slow start, it takes between 12 and 13 RTTs. At an RTT of 61ms, this corresponds to an average “bandwidth” of somewhere around 1 Gbit/s, which pretty much matches our measurement result.

--- 

## What do we actually want to show in this benchmark?

This might be controversial, but I'd argue that we **don't** want to measure:
* Specifics of the congestion controller in use (unless it's completely and horribly broken)
* How well the underlying stack makes use of hardware offloads (QUIC crypto offload for example)

What we actually want to show is that using libp2p doesn't impose a (significant) performance penalty when compared to vanilla HTTP(S), on typical machines.

This suggests that using a machine that has a 25 Gbit/s link speed might not be the most useful setup. Achieving even 5 Gbit/s throughput on a single connection (usually handled by a single CPU core) is challenging when using QUIC, unless various hardware offload are taken advantage of.

Using a high bandwidth also means that flow control windows need to large enough to accommodate a large BDP, which requires a high memory commitment, which a p2p node might not be willing to do due to DoS concerns.

## Measuring Throughput

We currently calculate throughput as `data transferred / time taken`. This is correct in the `t -> inf` limit, but less useful when slow start takes a dozen roundtrips and transfers O(500 MB) data. To average out the slow start period, we'd need to transfer an order of magnitude more data at least. This comes with two problems:
* Some implementations are currently very slow, for these implementations, transferring 10 GB of data would take a very long time.
* We're hardcoding assumptions about slow start and about the bandwidth of the connection into the amount of data we're requesting.

A better solution would be to have the client request an infinite amount of data, and calculate the _current_ bandwidth of the transfer every couple of seconds (i.e. only taking into account the amount of data transferred in the last interval). It would then be immediately obvious when these values have converged and we can stop the measurement. As a simplification, we can also just run the test for a fixed time (30s?).

mxinden:
Our current throughput tests open a connection, open a stream, up- or download 100MB and close the connection. 100 MB is not enough on the given path (60ms, ~5gbit/s) to exit congestion controller's slow-start. See https://github.com/libp2p/test-plans/issues/261 for details.

Instead of downloading 100MB multiple times, each on a new connection, establish a single connection and continuously measure the throughput for a fixed duration (20s).

Closes #261


marten-seemann (commented):
Should we do 3 iterations at 20 seconds? Slow start won't take more than 1-2s, so this should give us plenty of time to converge, and it would show situations where the congestion controllers runs into a state that it takes long to recover from (sudden cross-traffic).

marten-seemann:
> Preliminary results show fruitful for `https` but not `rust-libp2p/quic`.

Are these results available anywhere?

mxinden:
> > Preliminary results show fruitful for `https` but not `rust-libp2p/quic`.
> 
> Are these results available anywhere?

Not yet. Still in a very work-in-progress state.

mxinden:
Reaching 4.5 gbit/s with https and >5 gbit/s with rust-libp2p. Still testing. Though this is looking promising.

mxinden:
iperf throughput mismatch was due to Nagle's algorithm. Disabled now via `-N`. See previous commit. (Default for the https implementation already https://github.com/golang/go/issues/57530.)

newplot(10)

https://observablehq.com/d/682dcea9fe2505c4?branch=perf-exit-slow-start#branch

I still need to investigate more for the other measurements (*-libp2p and quic-go). Will try with a fixed MTU of 1500 next.

//CC @marten-seemann

marten-seemann:
> iperf throughput mismatch was due to Nagle's algorithm

Interesting! It's great to see iperf and HTTPS achieving roughly similar results (at least in the limit). This means that our setup is getting more trustworthy!

Looking at the graphs, why are some measurements drawn as boxes and some as points? Why do some have error bars and others don't? The spread seems pretty high, do we need more iterations?

I wouldn't be surprised if quic-go maxed out somewhere around 2 Gbps. At some point, your transfer becomes CPU-limited, depending on the number of kernel offloads that your QUIC stack uses (and that's not the thing we want to benchmark here). That said, I just updated quic-go/perf to quic-go v0.38.1 (https://github.com/quic-go/perf/pull/16, I'll merge the PR once GHA is not broken anymore...), which uses GSO by default. Might be worth rebasing your branch to see if this changes anything.

In go-libp2p, Yamux uses a [16 MB receive window](https://github.com/libp2p/go-libp2p/blob/dce942454b6b85c08fc58867f6e4d5506718e56f/p2p/muxer/yamux/transport.go#L24), which should limit us to roughly 2 Gbps (minus some muxer overhead). It's interesting to see that we're achieving roughly half of that. Could be a coincidence, or point to a bug in our flow control autotuning. I'd be happy to debug this using the current setup (assuming I can still run it manually as I could with the version on master), please let me know.

QUIC uses a [10 MB window](https://github.com/libp2p/go-libp2p/blob/dce942454b6b85c08fc58867f6e4d5506718e56f/p2p/transport/quicreuse/config.go#L13), which limits the bandwidth to 1.25 Gbps. That means we're not quite at the optimum, but pretty close. Would it be helpful for you if we prioritize resolving https://github.com/libp2p/go-libp2p/issues/2290? Alternatively, we could also just have a go-libp2p branch that bumps that value, so we can see if that's actually the root of the problem.

marten-seemann:
Here are some interesting result from running the HTTPS test and analyzing the tcpdump. The congestion controller used for this test is Cubic. First interesting result: importing and processing an 8 GB pcap into Wireshark takes a pretty long time O(30min) ;)

Here's the RTT distribution:
Screenshot 2023-09-07 at 22 59 31
There's definitely some queues building up in the network, but it's not too terrible.

Here's the sequence plot (ignore the wrapping of the packet number, obviously), showing the time when packet loss occurred:
Screenshot 2023-09-07 at 15 51 10

And here's the throughput:
Screenshot 2023-09-07 at 15 01 25

Obviously, we're very far from reaching a steady state. 

Here's some  back-of-the-envelope math to calculate the recovery time (i.e. the time it takes to ramp up the congestion window to its original size after a loss), and assuming a BDP (at 5 gbps and 65ms RTT) of roughly 40 MB:
* On Reno, a packet loss halves the cwnd, and every round-trip without a loss event increases it by one MSS. Thus the recovery time is (20 MB / 1400 bytes) ~14000 RTTs, which is 15 minutes (!).
* On Cubic, it's harder to estimate. The [L4S Prague paper](https://www.bobbriscoe.net/projects/latency/tcp-prague-netdev0x13.pdf) claims that at 100 mbps the recovery time is 250 RTTs, and doubles for every 8x increase in bandwidth. That's roughly 1000 RTTs, which is 1 minute.

In the sequence plot above, we see packet loss happening 10x as frequently as this calculation suggests. This might be due to the more shallow buffer, but I don't know precisely how the recovery time scales with the buffer size.

---

What does this mean for our perf setup? At the BDP that we chose for our test, we're running into limitations imposed by the congestion controllers:
* With a recovery time of 15 mintues, Reno has no chance against Cubic whatsoever, yet this is what RFC 9002 recommends for QUIC implementations and what major players have deployed in their QUIC stacks.
* Even Cubic's recovery time is pretty long. At a sampling frequency of 1s, we will pick up the saw-tooth pattern inherent to the congestion controller, and we will inevitably see a wide spread of measurement results.

mxinden:
> > Will try with a fixed MTU of 1500 next.
> 
> Does AWS allow larger MTUs on their backbone? That would indeed give TCP an unfair advantage over at least quic-go. Have you verified that using tcpdump / Wireshark?

Turns out, it does not:

```
[ec2-user@ip-]$ ping -M do -s 1472 -c 4 x.x.x.x
PING  () 1472(1500) bytes of data.
1480 bytes from : icmp_seq=1 ttl=109 time=63.4 ms
1480 bytes from : icmp_seq=2 ttl=109 time=63.4 ms
1480 bytes from : icmp_seq=3 ttl=109 time=63.4 ms
1480 bytes from : icmp_seq=4 ttl=109 time=63.4 ms

---  ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 3004ms
rtt min/avg/max/mdev = 63.384/63.408/63.441/0.024 ms
```
```
[ec2-user@ip-]$ ping -M do -s 1500 -c 4 x.x.x.x
PING  () 1500(1528) bytes of data.
From  icmp_seq=1 Frag needed and DF set (mtu = 1500)
ping: local error: message too long, mtu=1500
ping: local error: message too long, mtu=1500
ping: local error: message too long, mtu=1500

---  ping statistics ---
4 packets transmitted, 0 received, +4 errors, 100% packet loss, time 3086ms

File: perf/README.md [hunk 1]
```diff
@@ -79,11 +79,33 @@ Given you have provisioned your infrastructure, you can now build and run the li
...
+               ```
+            - Before terminating the client must print a final summary. See example below. Note the `type: "final"`. Also note that the measurement includes the time to (1) establish the connection, (2) upload the bytes and (3) download the bytes.
+               ``` json
+               {
+                 "type": "final",
+                 "timeSeconds": 60.127230659,
+                 "uploadBytes": 4382392320,
+                 "downloadBytes": 0
+               }
+               ```
```
mxinden, line 109: This pull request requires the following changes to a perf binaries output format.

This pull request already updates the `go-libp2p` and `https` implementation. https://github.com/quic-go/perf/pull/17 updates the `quic-go` implementation. https://github.com/libp2p/rust-libp2p/pull/4382 updates the rust-libp2p implementation.

The motivation for the larger effort is documented in https://github.com/libp2p/test-plans/issues/261.

@maschad do you have time and interest to do the corresponding changes on the js-libp2p side? If so, as a first step, you could update the existing `stdout` output to the above detailed `type: "final"` output. As a second step the `type: "intermediate"` can follow.

mxinden:
> Looking at the graphs, why are some measurements drawn as boxes and some as points?

The Box visualizes Q1 to Q3 with Q2 (median) denoted with a line within the box. The lines are the whiskers, representing Q0 (minimum) and Q4 (maximum). The dots represent outliers.

https://en.wikipedia.org/wiki/Box_plot has a good explanation for each of these.

> Why do some have error bars and others don't?

I am not sure what you refer to with "error bars" @marten-seemann.

> The spread seems pretty high, do we need more iterations?

I decreased each measurement duration to 20 seconds and increased the iterations per implementation and transport to 10. I will trigger a new CI run soon so that we get more reliable data.

mxinden:
> I wouldn't be surprised if quic-go maxed out somewhere around 2 Gbps. At some point, your transfer becomes CPU-limited, depending on the number of kernel offloads that your QUIC stack uses (and that's not the thing we want to benchmark here). That said, I just updated quic-go/perf to quic-go v0.38.1 ([quic-go/perf#16](https://github.com/quic-go/perf/pull/16), I'll merge the PR once GHA is not broken anymore...), which uses GSO by default. Might be worth rebasing your branch to see if this changes anything.

:+1: Note that I merged current `master` into https://github.com/quic-go/perf/pull/17 and updated the reference here. Thus with the next update to `benchmark-results.json` we will see the impact of GSO.

mxinden:
> In go-libp2p, Yamux uses a [16 MB receive window](https://github.com/libp2p/go-libp2p/blob/dce942454b6b85c08fc58867f6e4d5506718e56f/p2p/muxer/yamux/transport.go#L24), which should limit us to roughly 2 Gbps (minus some muxer overhead). It's interesting to see that we're achieving roughly half of that. Could be a coincidence, or point to a bug in our flow control autotuning. I'd be happy to debug this using the current setup (assuming I can still run it manually as I could with the version on master), please let me know.

Indeed surprising. You can still run it manually. Please go ahead. Thank you @marten-seemann.

perf/README.md
maschad: Thanks for the work on this PR @mxinden and the clear write up. I created an issue https://github.com/libp2p/js-libp2p/issues/2064 to track this and will tackle it once I have completed my current in-flight work. Once that is merged and released I will open a PR to https://github.com/libp2p/test-plans

mxinden:
I have updated the forked dashboard to the latest data format:

https://observablehq.com/d/682dcea9fe2505c4?branch=27d07a6f47c2bc1a9c9d9a9f6626b536248284f5

mxinden (commented):
@marten-seemann @sukunrt can either of you give the `perf/impl/https` and `perf/impl/go-libp2p` changes a review?

perf/impl/go-libp2p/v0.29/main.go
```diff
@@ -99,7 +102,10 @@ func main() {
 }
 
 type Result struct {
-	Latency float64 `json:"latency"`
+	Type          string  `json:"type"`
+	TimeSeconds   float64 `json:"timeSeconds"`
+	UploadBytes   uint    `json:"uploadBytes"`
```
marten-seemann, line 107: This would better be a uint64, otherwise you'll quickly run into limits on 32 bit machines.

perf/impl/go-libp2p/v0.29/perf.go
```diff
@@ -89,7 +91,26 @@ func sendBytes(s io.Writer, bytesToSend uint64) error {
 	buf := pool.Get(blockSize)
 	defer pool.Put(buf)
 
+	lastReportTime := time.Now()
+	lastReportWrite := 0
+
 	for bytesToSend > 0 {
+		now := time.Now()
+		if now.Sub(lastReportTime) > time.Second {
```
marten-seemann, line 99:
```suggestion
		if now.Sub(lastReportTime) >= time.Second {
```

perf/impl/go-libp2p/v0.29/perf.go
```diff
@@ -100,15 +121,49 @@ func sendBytes(s io.Writer, bytesToSend uint64) error {
...
 		bytesToSend -= uint64(n)
+		lastReportWrite += n
 	}
 	return nil
 }
 
 func drainStream(s io.Reader) (uint64, error) {
 	var recvd int64
-	recvd, err := io.Copy(io.Discard, s)
+	recvd, err := io.Copy(io.Discard, & reportingReader { orig: s, LastReportTime: time.Now() })
```
marten-seemann, line 131: Does it even work with the spaces? Can you run `gofmt ./...` on the code?
```suggestion
	recvd, err := io.Copy(io.Discard, &reportingReader{ orig: s, LastReportTime: time.Now() })
```

perf/impl/go-libp2p/v0.29/perf.go
```diff
@@ -100,15 +121,49 @@ func sendBytes(s io.Writer, bytesToSend uint64) error {
...
+}
+
+var _ io.Reader = &reportingReader{}
+
+func (r *reportingReader) Read(b []byte) (int, error) {
+	n, err := r.orig.Read(b)
+	r.lastReportRead += uint64(n)
+
+	now := time.Now()
+	if now.Sub(r.LastReportTime) > time.Second {
```
marten-seemann, line 151:
```suggestion
	if now.Sub(r.LastReportTime) >= time.Second {
```

marten-seemann (commented):
I reviewed the go-libp2p implementation.

mxinden:
The last commit removes js-libp2p. Once https://github.com/libp2p/js-libp2p/pull/2067 is merged, we can re-introduce it here.

Issue #279, opened 2023-08-25
interop: building js-libp2p v0.46 image fails


marten-seemann:
Building locally (M1 MBP):
```
 => ERROR [4/6] RUN npm i && npm run build                                                                                                                                                                         407.5s
------
 > [4/6] RUN npm i && npm run build:
308.3 npm WARN deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
308.6 npm WARN deprecated rollup-plugin-terser@7.0.2: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-terser
308.6 npm WARN deprecated workbox-cacheable-response@6.6.0: workbox-background-sync@6.6.0
310.3 npm WARN deprecated stable@0.1.8: Modern JS already guarantees Array#sort() is a stable sort, so this library is deprecated. See the compatibility table on MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#browser_compatibility
312.3 npm WARN deprecated multibase@4.0.6: This module has been superseded by the multiformats module
312.6 npm WARN deprecated ipfs-core-utils@0.18.1: js-IPFS has been deprecated in favour of Helia - please see https://github.com/ipfs/js-ipfs/issues/4336 for details
312.7 npm WARN deprecated w3c-hr-time@1.0.2: Use your platform's native performance.now() and performance.timeOrigin.
314.6 npm WARN deprecated svgo@1.3.2: This SVGO version is no longer supported. Upgrade to v2.x.x.
315.3 npm WARN deprecated ipfs-core-types@0.14.1: js-IPFS has been deprecated in favour of Helia - please see https://github.com/ipfs/js-ipfs/issues/4336 for details
407.1 npm notice
407.1 npm notice New minor version of npm available! 9.6.7 -> 9.8.1
407.1 npm notice Changelog: <https://github.com/npm/cli/releases/tag/v9.8.1>
407.1 npm notice Run `npm install -g npm@9.8.1` to update!
407.1 npm notice
407.1 npm ERR! code 2
407.1 npm ERR! path /app/node_modules/node-datachannel
407.1 npm ERR! command failed
407.1 npm ERR! command sh -c prebuild-install || (npm install --ignore-scripts --production=false && npm run _prebuild)
407.1 npm ERR! added 600 packages, removed 1 package, and audited 658 packages in 37s
407.1 npm ERR!
[...]

PR #286, opened 2023-08-28
multidim: Reproducible js-libp2p

MarcoPolo:
Only v0.46, with the thought that we'll soon release v0.47 and drop v0.45.

This solves the problem of the js-libp2p build here not being reproducible. This caused some considerable frustration from @marten-seemann . One problem is that js-libp2p does not commit it's [package-lock.json file](https://github.com/libp2p/js-libp2p/issues/1724), and the interop image doesn't commit the lock file either. This means that a build that worked one day, may fail the next because of a subtle dependency change.

This solves the problem by committing the package-lock.json file into this repo. This file is copied into the build and serves to pin the exact versions of every dependency.

Moving forward, we should continue to include the package-lock.json in here until/unless we include one in js-libp2p.

PR #288, opened 2023-08-30
fix(multidim): make js-libp2p builds reproducible

maschad:
Uses the `@libp2p/multidim-interop` for interop tests.

~~Depends on https://github.com/libp2p/js-libp2p/pull/2023~~

Closes https://github.com/libp2p/test-plans/issues/279


p-shahi:
I just re-triggered failing CI jobs (which ran 4 days ago and failed because they couldn't find @libp2p/multidim-interop v1.0.0; resolved per https://github.com/libp2p/js-libp2p/pull/2008#issuecomment-1703031949)

Now failing due to: 

> #5 [2/3] WORKDIR /app
> #5 DONE 0.0s
> 
> #6 [3/3] RUN npm run build
> #6 0.599 npm ERR! code ENOENT
> #6 0.601 npm ERR! syscall open
> #6 0.601 npm ERR! path /app/package.json
> #6 0.602 npm ERR! errno -2
> #6 0.603 npm ERR! enoent ENOENT: no such file or directory, open '/app/package.json'




Assuming this PR supplants https://github.com/libp2p/test-plans/pull/286 - want to close that (once CI is green and we have reproducible builds here)

maschad:
@p-shahi the build required [multidim interop to be published with aegir as a dependency](https://www.npmjs.com/package/@libp2p/multidim-interop/v/1.0.0-c4eff4c5?activeTab=dependencies) once that was resolved I had intended to test it locally then push my changes, which I have done now.

multidim-interop/impl/js/v0.46/Makefile
```diff
@@ -1,5 +1,4 @@
 image_name := js-v0.46
-commitSha := b7e608998cc88860d9ec8a3ed7c03fdfb3eccb3b
```
p-shahi, line 2: Now that the commit hash is removed, what version of js-libp2p is being tested? The image_name states v0.46 but what patch version is used?

p-shahi (commented):
Why aren't you including the package-lock.json that Marco added here: https://github.com/libp2p/test-plans/pull/286/files#diff-91f1832c796ab4c1041210aeb8f5d0d179090661044a631ac2919098e1338b07

Everyone was in agreement that it's important to add here since js-libp2p monorepo doesn't have a lock file. Do we have a reason for not including it?

multidim-interop/impl/js/v0.46/Dockerfile
```diff
@@ -1,11 +1,9 @@
 # Here because we want to fetch the node_modules within docker so that it's
 # installed on the same platform the test is run. Otherwise tools like `esbuild` will fail to run
-FROM node:18
+FROM node:18.17.1
 WORKDIR /app
-COPY . .
-RUN npm i && npm run build
+RUN npm install @libp2p/multidim-interop@1.0.0-c4eff4c5
```
p-shahi, line 5:
```suggestion
RUN apt-get update && apt-get install -y cmake
RUN npm install @libp2p/multidim-interop@1.0.0-c4eff4c5
```

This comment needs to be addressed: https://github.com/libp2p/js-libp2p/pull/2008#issuecomment-1703067528
When I try to build this locally without that patch this fails at the cmake step

maschad: Good catch thanks!

multidim-interop/impl/js/v0.46/Makefile
maschad: the commit hash is in [npm package version](https://github.com/libp2p/test-plans/pull/288/files#diff-1d6f516e3ef877a2f1b1b2f21c642b29c3c0236271a8cc7bc4bcbbb6c3d1a634R3). As of now the latest patch libp2p i.e. `0.46.9` is being tested.

If we want to have the patch version fixed we can do that in the monorepo.

maschad:
My ori

> Why aren't you including the package-lock.json that Marco added here: [#286 (files)](https://github.com/libp2p/test-plans/pull/286/files#diff-91f1832c796ab4c1041210aeb8f5d0d179090661044a631ac2919098e1338b07)
> 
> Everyone was in agreement that it's important to add here since js-libp2p monorepo doesn't have a lock file. Do we have a reason for not including it?

My original thought was that we would fix the dependencies in the `@libp2p/multidim-interop` package given we control the releases but upon reconsideration it may be more maintainable to commit the lockfile.

maschad:
Unrelated [failure](https://github.com/libp2p/test-plans/actions/runs/6099624404/job/16552212643?pr=288#step:3:3897) between two rust `v0.52` nodes on webRTCDirect. @mxinden have you seen this failure before?

multidim-interop/impl/js/v0.46/Makefile
p-shahi: So the version of js-libp2p tested here is fetched transitively by the multidim-interop version used i.e. `"@libp2p/multidim-interop": "^1.0.0-c4eff4c5"` ?

Please add the js-libp2p version in package.json too - it is a dependency of this test suite and we should be explicit which version is being tested (and not have to infer it from the package-lock created by the multidim-interop package).

I'd also ask that the image be updated here to js-v0.46.9

> If we want to have the patch version fixed we can do that in the monorepo.

What does this mean?

p-shahi:
> My original thought was that we would fix the dependencies in the `@libp2p/multidim-interop` package given js-libp2p controls the releases but upon reconsideration it may be more maintainable to commit the lockfile.

Sorry I don't understand how "fix the dependencies in the `@libp2p/multidim-interop` package" would solve the build reproducibility issue described [here](https://github.com/libp2p/test-plans/pull/286#issue-1870617580)
"One problem is that js-libp2p does not commit it's https://github.com/libp2p/js-libp2p/issues/1724, and the interop image doesn't commit the lock file either. This means that a build that worked one day, may fail the next because of a subtle dependency change."
I might be missing something on how JS specific package management works though🤔 ?

maschad:
> Sorry I don't understand how "fix the dependencies in the @libp2p/multidim-interop package" would solve the build reproducibility issue described https://github.com/libp2p/test-plans/pull/286#issue-1870617580

Perhaps the use of the word "fix" in this context was a bit misleading, if [the dependencies](https://github.com/libp2p/js-libp2p/blob/master/interop/package.json#L52-L63) of `@libp2p/multdim-interop` are set to a static version e.g. `"0.46.9"`  as opposed to a [version range](https://github.com/npm/node-semver#ranges) then npm will download specifically that version, which is the equivalent behaviour of a lockfile and thus reproducible.

Given we maintain the `@libp2p/multdim-interop` repo and it's dependencies, and it is only consumed by this repository, the versions can be set for this use case as opposed to other packages which have multiple consumers. 

The downside is that this is requires manually updating each dependency in that package (thus more maintenance work), the upside is that we have more granular control over which dependencies change in the next package (as opposed to a new `@libp2p/multidim-interop` package which may have multiple dependency updates).

multidim-interop/impl/js/v0.46/Makefile
maschad: > I'd also ask that the image be updated here to js-v0.46.9

I will update it.

> What does this mean?

See [my comment](https://github.com/libp2p/test-plans/pull/288#issuecomment-1708802958)

multidim-interop/impl/js/v0.46/package-lock.json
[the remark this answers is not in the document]
maschad: it's actually the "js-libp2p-multidim-interop-test" , I've updated it.

multidim-interop/impl/js/v0.46/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@libp2p/multidim-interop": "^1.0.0-c4eff4c5"
+  }
```
p-shahi, line 4:
```suggestion
  },
  "overrides": {
    "@libp2p/mplex": "9.0.5",
    "@libp2p/tcp": "8.0.5",
    "@libp2p/webrtc": "3.1.9",
    "@libp2p/websockets": "7.0.5",
    "@libp2p/webtransport": "3.0.9",
    "libp2p": "0.46.5"
  }
```

You should add this to pin the dependencies in the monorepo to the commit of the actual release.
Otherwise, you'll install a dirty `-c4eff4c5` for each package which is [evident in the lock file](https://github.com/libp2p/test-plans/blob/dc63bc5be8442dcf833a5c52c273bdde88d344f3/multidim-interop/impl/js/v0.46/package-lock.json).
```
        "@libp2p/mplex": "9.0.5-c4eff4c5",
        "@libp2p/tcp": "8.0.5-c4eff4c5",
        "@libp2p/webrtc": "3.1.9-c4eff4c5",
        "@libp2p/websockets": "7.0.5-c4eff4c5",
        "@libp2p/webtransport": "3.0.9-c4eff4c5",
        "@multiformats/mafmt": "^12.1.2",
        "@multiformats/multiaddr": "^12.1.5",
        "aegir": "^40.0.12",
        "libp2p": "0.46.9-c4eff4c5",
```
this may not seem like a problem but I don't think it's good practice to test against unreleased commits.

adding these overrides at least allows us to explicitly specify which version of libp2p packages are being tested without having to search in the lock file

p-shahi (approved):

p-shahi:
@marten-seemann can you verify this branch builds js-libp2p for you; otherwise it shouldn't be merged

PR #269, opened 2023-08-22
chore: add go-libp2p v0.30 and remove v0.27 and v0.28

p-shahi:

PR #289, opened 2023-08-31
chore(multidim-interop): add go v0.31

p-shahi:


marten-seemann:
Resolved by #295.

p-shahi:
@marten-seemann #295 adds this to perf/ not multidim-interop/

p-shahi:
Failure is a flaky test: chromium-js-v0.46 x rust-v0.52 (webrtc-direct),failure

marten-seemann:
> @marten-seemann #295 adds this to perf/ not multidim-interop/

Sorry, I must have missed that!

marten-seemann (changes requested):
Why are we removing v0.29? It's the version that's used on the latest Kubo release? What's the policy you're following here to drop v0.27 and v0.28?

We should consider dropping v0.30 though, because it was never deployed.

p-shahi:
> What's the policy you're following here to drop v0.27 and v0.28?

v0.27 and v0.28 were already removed in https://github.com/libp2p/test-plans/pull/269 - I'm doing some cleanup in this PR

No strict policy but the rationale for removing older versions was discussed here: https://filecoinproject.slack.com/archives/C03K82MU486/p1691624748108769

i.e. we can use the PRs as a place to decide whether it's ok to remove an old version.

If you think we should keep v0.29 for a bit longer I'm ok to add it back.
Dropping v0.30 because it wasn't widely deployed doesn't seem like a compelling enough reason to me - shouldn't the versions tested here be contiguous?

marten-seemann:
> shouldn't the versions tested here be contiguous?

I don't see why. We should be testing what's in use. v0.30 was primarily a release to get Go 1.21 compatibility in.

p-shahi:
> > shouldn't the versions tested here be contiguous?
> 
> I don't see why. We should be testing what's in use. v0.30 was primarily a release to get Go 1.21 compatibility in.

Let's keep it and drop it when 0.32 is released? Unless you feel super strongly this is my preference

marten-seemann:
As long as we keep testing v0.29, this is fine. I see you re-added it, can you update the title of this PR as well?

p-shahi:
Done

p-shahi:
image

marten-seemann (approved):

PR #294, opened 2023-09-04
fix: how new go-libp2p versions are added to perf

galargh:
Tested in https://github.com/libp2p/test-plans/actions/runs/6070021302


.github/workflows/add-new-impl-versions.yml
```diff
@@ -27,7 +27,8 @@ jobs:
         uses: actions/checkout@v3
       - name: Configure git
         run: |
-          git fetch $BRANCH && git checkout $BRANCH && git rebase $GITHUB_REF -X theirs || git checkout -b $BRANCH
+          git fetch origin $BRANCH && git checkout $BRANCH || git checkout -b $BRANCH
+          git rebase $GITHUB_REF -X theirs || git rebase --abort
```
galargh, line 31: This fixes checkout which was previously always failing, and rebase so that we can get out of a situation where we cannot rebase cleanly.

perf/impl/go-libp2p/v0.29/Makefile
```diff
@@ -3,10 +3,9 @@ GO_FILES := $(wildcard *.go)
 all: perf
 
 perf: $(GO_FILES)
-	docker run --rm --user "$(shell id -u):$(shell id -g)" -v "$(shell pwd)":/usr/src/myapp -w /usr/src/myapp -e GOCACHE=/usr/src/myapp/.cache golang:1.20 go build -o perf .
+	docker run --rm --user "$(shell id -u):$(shell id -g)" -v "$(shell pwd)":/usr/src/myapp -w /usr/src/myapp -e GOCACHE=/usr/src/myapp/.cache golang:$(shell awk '/^go [0-9]+\.[0-9]+$$/ {print $$2}' go.mod) go build -o perf .
```
galargh, line 6: We should be using the version from go.mod here.

PR #296, opened 2023-09-04
feat(.github/add-new-impl): reference metrics dashboard

mxinden:
Reference the perf dashboard, linking to the branch of the pull request, i.e. perf: https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=perf%2Fgo-libp2p#branch

### DIFF
--- a/.github/workflows/add-new-impl-versions.yml
+++ b/.github/workflows/add-new-impl-versions.yml
@@ -65,7 +65,7 @@ jobs:
           git push origin $BRANCH --force
           # create a PR if an open one doesn't exist yet
           if [[ $(gh pr list --state open --base $GITHUB_REF --head $BRANCH | wc -l) -eq 0 ]]; then
-            gh pr create --title "chore: add go-libp2p@$REMOTE_VERSION to $DIR" --body "This PR adds go-libp2p@$REMOTE_VERSION to $DIR" --head $BRANCH --base $GITHUB_REF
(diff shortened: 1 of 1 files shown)


.github/workflows/add-new-impl-versions.yml
```diff
@@ -65,7 +65,7 @@ jobs:
           git push origin $BRANCH --force
           # create a PR if an open one doesn't exist yet
           if [[ $(gh pr list --state open --base $GITHUB_REF --head $BRANCH | wc -l) -eq 0 ]]; then
-            gh pr create --title "chore: add go-libp2p@$REMOTE_VERSION to $DIR" --body "This PR adds go-libp2p@$REMOTE_VERSION to $DIR" --head $BRANCH --base $GITHUB_REF
+            gh pr create --title "chore: add go-libp2p@$REMOTE_VERSION to $DIR" --body "This PR adds go-libp2p@$REMOTE_VERSION to $DIR\nSee new metrics at https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=perf%2Fgo-libp2p#branch once \"perf: update benchmark results\" commit is pushed." --head $BRANCH --base $GITHUB_REF
```
marten-seemann, line 68: This always references the `branch` branch, or am I misunderstanding the link?

galargh: How about we create a comment on a PR (if it exists) when we push the commit (in the perf job itself)? That way we could reference the exact SHA of the commit and the link would keep working after the branch is merged & deleted. I can put a proposal for that in a sec.

galargh: That's what I had in mind - https://github.com/libp2p/test-plans/pull/297/files

PR #298, opened 2023-09-04
Add .NET version of tcp/noise/yamux

flcl42:
Adds .NET implementation of tcp/noise/yamux stack of libp2p to the interoperability testing suite


multidim-interop/impl/dotnet/v1.0/.gitignore
```diff
@@ -0,0 +1,5 @@
+bin/
+obj/
+.vs/
+.vscode/
```
mxinden, line 4: Please no editor specific configurations in the project `.gitignore`. These should go into your home `~/.gitignore`.

multidim-interop/impl/dotnet/v1.0/TestPlansApp.csproj
mxinden, line 1: I am not familiar with .net's package management. We try to make the interop tests as reproducible as possible. Does .net have a lock file that one can check in to guarantee that the dependencies don't change?

flcl42: It's differently possible using `packages.lock.json` afaik. Added

flcl42:
Let's add quic also, converting to draft

mxinden:
Approved and ran CI. 

> Let's add quic also, converting to draft

I am happy to merge here without QUIC already. Your call.

transport-interop/impl/dotnet/v1.0/Program.cs
```diff
@@ -0,0 +1,152 @@
...
+
+    public TestPlansPeerFactoryBuilder(string transport, string? muxer, string? security)
+        : base(new ServiceCollection()
+            //   .AddLogging(builder =>
+            //       builder.SetMinimumLevel(LogLevel.Trace)
+            //           .AddSimpleConsole(l =>
+            //           {
+            //               l.SingleLine = true;
+            //               l.TimestampFormat = "[HH:mm:ss.FFF]";
+            //           }))
```
mxinden, line 103: Needed?

flcl42: Not really

mxinden:
@flcl42 seems like dotnet is failing with all other implementations. Mind taking a look?

transport-interop/impl/dotnet/v1.0/Dockerfile
mxinden, line 1: @thomaseizinger given your familiarity with the transport-interop caching system, would you mind taking a look at this Dockerfile?

flcl42:
> @flcl42 seems like dotnet is failing with all other implementations. Mind taking a look?

The current implementation:
- is not compatible with zig and rust via quic-v1 
- has flaky tests for js x dotnet, let me try to fix it again. 
Other pairs work.

If it's better to have it 100% passing, let me disable quic for now.

transport-interop/impl/dotnet/v1.0/Dockerfile
thomaseizinger: Looks good to me. Final layer should be as small as possible.

transport-interop/impl/dotnet/v1.0/Program.cs
thomaseizinger, line 1: For all other implementations, we keep the actual test within the repository so it can be atomically updated with the library code.

Unless there is a strong reason to not do that for `.net`, I'd suggest to do the same here.

transport-interop/impl/zig/v0.0.1/Makefile
thomaseizinger, line 1: I am assuming this needs to be reverted?

transport-interop/src/compose-runner.ts
```diff
@@ -9,7 +9,7 @@ import { stringify } from 'yaml';
 import { dialerStdout, dialerTimings } from './compose-stdout-helper';
 
 const exec = util.promisify(execStd);
-const timeoutSecs = 3 * 60
+const timeoutSecs = 42
```
thomaseizinger, line 12: Is this intentional?

transport-interop/src/compose-runner.ts
```diff
@@ -50,7 +50,9 @@ export async function run(namespace: string, compose: ComposeSpecification, opts
         const controller = new AbortController();
         const timeoutId = setTimeout(() => controller.abort(), 1000 * timeoutSecs)
         const { signal } = controller;
-        const { stdout, stderr } = await exec(`docker compose -f ${path.join(dir, "compose.yaml")} up ${upFlags.join(" ")}`, { signal })
+        const dcPath = `docker compose -f ${path.join(dir, "compose.yaml")} up ${upFlags.join(" ")}`
+        console.warn(dcPath)
```
thomaseizinger, line 54: I'd prefer making this a more descriptive log line if we intend to keep it.

flcl42:
@thomaseizinger Thanks for the remarks! Right now we are refactoring/fixing the library to fit the tests. I'll also address the suggestions in new update.

dhuseby:
@flcl42 @seetadev let's see if we can get this updated and landed now that we have people looking for tasks to do.

seetadev:
@dhuseby : Definitely, Dave. Thank you so much for sharing. Appreciate it.

We did have a discussion on it in the last community call. Also, shared the similar PRs undertaken in Py-libp2p.

Will follow up with @flcl42 .

Issue #301, opened 2023-09-13
Rust's WASM container is not cached properly


thomaseizinger:
Not sure what is going on here but we are spending significant time in the test suite to re-build the `v0.52` rust container for the wasm build. See https://github.com/libp2p/test-plans/actions/runs/6105882311/job/16570044582#step:3:1899.

For some reason, the caching is not working for this one?

thomaseizinger:
This might be because the cache optimisation assumes a single image per `impl` directory.

PR #302, opened 2023-09-13
fix: make `rust-chromium` a separate implementation

thomaseizinger:
The current caching implementation expects only a single docker container per implementation directory. This was not true for Rust `v0.52` which produced two containers: A native one and one with chromium and a WASM build.

Thus, the latter never got cached and we were rebuilding it on every run. By making it a separate implementation, the cache should be picked up again.

I refrained from extending the cache implementation to handle this because I think it would get too complicated.

Resolves: #301.

### DIFF
--- a/multidim-interop/impl/rust-chromium/.gitignore
+++ b/multidim-interop/impl/rust-chromium/.gitignore
@@ -1,0 +1,5 @@
+rust-libp2p-*.zip
+rust-libp2p-*
+rust-libp2p-*/*
+image.json
+chromium-image.json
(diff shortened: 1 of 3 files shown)


thomaseizinger:
Marking this as draft until I can confirm that it actually works.

thomaseizinger:
It should work now:

- Here is where the cache for the chromium build would be loaded: https://github.com/libp2p/test-plans/actions/runs/6170406724/job/16746635420?pr=302#step:3:2485
- Hereis where the cache for the native build would be loaded: https://github.com/libp2p/test-plans/actions/runs/6170406724/job/16746635420?pr=302#step:3:734

We can see that the two keys are different.

mxinden (approved):
Sounds good to me! Thank you for debugging.

Issue #303, opened 2023-09-13
Re-consider container caching strategy to use registries


thomaseizinger:
Whilst working on the interop-tests and taking a closer look at our container caching in general, I had the following thoughts:

The current caching mechanism introduces a fair bit of complexity:
- ~ 20 lines of Makefiles per version: https://github.com/libp2p/test-plans/blob/master/multidim-interop/impl/rust/v0.51/Makefile
- Custom logic of computing the cache key and up/downloading the cache to S3: https://github.com/libp2p/test-plans/blob/master/multidim-interop/helpers/cache.ts
- The above script needs to be manually called: https://github.com/libp2p/test-plans/blob/b8235c9eae35670a9aa4edf62b16119ea506cb2e/.github/actions/run-interop-ping-test/action.yml#L79-L104
- We need to manage AWS credentials: https://github.com/libp2p/test-plans/blob/b8235c9eae35670a9aa4edf62b16119ea506cb2e/.github/actions/run-interop-ping-test/action.yml#L39-L43
- We need to set custom caching options for each docker build: https://github.com/libp2p/test-plans/blob/master/multidim-interop/dockerBuildWrapper.sh
- I am about to duplicate all of the above for the hole-punching tests :sob: 

The benefit we are getting from this is that we can pretty much plug any commit of a repository into the Makefile, hit make and we end up with a container. The cache is thus purely a performance optimisation.

Is this worth the complexity? Yesterday, I discovered a subtle bug in our setup that made us not cache a particular Rust image, see https://github.com/libp2p/test-plans/issues/301.

The test runner is already designed to work in phases:

1. Generate the permutations of test cases
2. Generate a `docker-compose.yml` file
3. Run a particular `docker-compose.yml` file

If we would use container registries instead, we could delete all of the above code by just referencing image IDs in the `versions.ts` file.

When debugging code, we could always generate the offending `docker-compose.yml` file first and swap the container reference out to a point at a `Dockerfile` instead which would build a local container instead.https://github.com/libp2p/test-plans/issues/282 already hints at this too.

Currently, the "contract" between `libp2p/test-plans` and the repositories is that the need to provide a `Makefile` which builds a container. The new contract would be that they need to provide a `Dockerfile` that builds a functioning version. This would likely be more useful because not every developer has all build-environments of other languages set up.

thomaseizinger:
cc @mxinden @achingbrain @marten-seemann

mxinden:
Sorry for the late reply.

[...]

PR #304, opened 2023-09-14
feat(hole-punch): add hole-punch interoperability test suite

thomaseizinger:
## How to review this

1. Read the [README](https://github.com/thomaseizinger/test-plans/blob/feat/hole-punch-tests/hole-punch-interop/README.md)
2. Look at `src/generator.ts`
3. Look at `testplans.ts`

Most of the other files like `helpers/cache.ts` are a 1-to-1 copy from `multidim-interop`.

## Follow-ups

- [ ] Investigate why Rust test has double the RTT for the hole-punched connection on TCP in **one** direction
- [ ] Assert RTT once above is fixed


thomaseizinger:
Don't be fooled by green CI, it is actually failing :D (But was working locally)

thomaseizinger:
> Don't be fooled by green CI, it is actually failing :D (But was working locally)

Now we actually have passing tests!

thomaseizinger:
One interesting thing I just learned is that delay added with `tc` only affects _outgoing_ packets. That is why the RTT is only half of what you'd expect.

thomaseizinger:
Another interesting learning, setting `TCP_NODELAY` is _really_ important when measuring RTTs :upside_down_face:

thomaseizinger:
So I've got the RTT check working. It is all fine for QUIC but for TCP, I am getting an RTT of 200ms for the listener (which would be what I'd expect from setting a 100ms delay on each router) but a 400ms RTT for the dialer. Not sure why the RTT would be different from one direction vs the other :man_shrugging:

thomaseizinger:
> So I've got the RTT check working. It is all fine for QUIC but for TCP, I am getting an RTT of 200ms for the listener (which you'd expect from setting a 100ms delay on each router) but a 400ms RTT for the dialer. Not sure why the RTT would be different from one direction vs the other when all we open is a stream in each direction?

Note to myself: Try swapping noise for TLS and yamux for mplex to narrow down problem.

thomaseizinger:
> > So I've got the RTT check working. It is all fine for QUIC but for TCP, I am getting an RTT of 200ms for the listener (which you'd expect from setting a 100ms delay on each router) but a 400ms RTT for the dialer. Not sure why the RTT would be different from one direction vs the other when all we open is a stream in each direction?
> 
> Note to myself: Try swapping noise for TLS and yamux for mplex to narrow down problem.

Swapped both of them, still the same issue :(

thomaseizinger:
I've been analyzing some tcpdumps and weirdly, the IPs are only correct for the _listener_. In the _dialer_ log, it appears as if it would only talk to the relay? Will continue analyzing tomorrow ..

thomaseizinger:
Well, I am glad I found https://github.com/docker/compose/issues/10805 which seems to finally fix the flakiness in CI!

thomaseizinger:
@marten-seemann This is ready to have a go-peer added! I am curious to learn whether the go-peer also observed a weird RTT via the TCP connection. That would tell us whether this is an issue with our network setup or a bug in `rust-libp2p` somewhere.

hole-punch-interop/README.md
```diff
@@ -0,0 +1,84 @@
+# Hole punch tests
+
+## How to run locally
+
+1. `npm run install`
```
mxinden, line 5:
```suggestion
1. `npm install`
```

mxinden (commented):
:tada:

```
➜  hole-punch-interop git:(feat/hole-punch-tests) npm run test

> libp2p hole-punch test@0.0.1 test
> ts-node src/*.test.ts && ts-node testplans.ts

Running 2 tests
Running test spec: rust-v0.52 x rust-v0.52 (quic)
Running test spec: rust-v0.52 x rust-v0.52 (tcp)
Expected RTT of direct connection to be ~200ms but was 401ms
0 failures:
Run complete
```

The amount of code duplication to the existing interop tests is unfortunate. That said, the duplication might just be easier here than a bad abstraction.

thomaseizinger:
> The amount of code duplication to the existing interop tests is unfortunate. That said, the duplication might just be easier here than a bad abstraction.

I have two ideas to deal with this:

1. Get rid of most of the code which just deals with caching: https://github.com/libp2p/test-plans/issues/303
	- Makefiles
	- Cache helpers
	- Build shell-script
	- Version resolution 
3. Rewrite the rest to use `deno`:
	- Deno is much more a "convention over configuration" approach. It directly supports typescript for example.
	- Sharing utility functions across the various test runners should be easy as a result

thomaseizinger:
@Menduist Would you have time to write a `nim-libp2p client for this test suite so we can get a rough idea of whether it is working across implementations at this point?

thomaseizinger:
Hey everyone! This has been ready for additions of other languages for a while. Instead of having it sit here, my suggestion would be to merge it as is and activate it in `rust-libp2p` for now. That way, we get some insight on how stable the tests are once they are run on each PR.

Other languages are then free to add their implementation at their own pace.

Any objections @mxinden @marten-seemann @MarcoPolo @achingbrain @maschad @Menduist (not sure who else to tag?)?

mxinden (approved):
I am in favor of moving forward here. What do others think?

Menduist:
Sorry, just seen the message, didn't realize this was already ready for integration
@diegomrsantos will add a nim implementation, but feel free to merge in the meantime :)

thomaseizinger:
> I am in favor of moving forward here. What do others think?

I am planning to merge this in ~24h unless there are any concerns raised!

PR #305, opened 2023-09-15
chore(transport-interop): add js-libp2p v0.46.21

maschad:
Also upgrade the dependencies


maschad:
@p-shahi would you remind giving this a review?

p-shahi (approved):

multidim-interop/impl/js/v0.46/package.json
```diff
@@ -1,19 +1,19 @@
 {
   "dependencies": {
-    "@libp2p/multidim-interop": "1.0.0-c4eff4c5"
+    "@libp2p/multidim-interop": "1.0.1"
```
achingbrain, line 3: Maybe use a range here?

```suggestion
    "@libp2p/multidim-interop": "^1.0.1"
```

multidim-interop/impl/js/v0.46/package.json
```diff
@@ -1,19 +1,19 @@
...
   },
   "overrides": {
     "@chainsafe/libp2p-noise": "13.0.1",
     "@chainsafe/libp2p-yamux": "5.0.0",
     "@libp2p/mplex": "9.0.5",
-    "@libp2p/tcp": "8.0.5",
```
achingbrain, line 9: Are all these overrides still necessary? Even with a lock file?

achingbrain (changes requested):
LGTM, a couple of questions about dep versions inline

multidim-interop/impl/js/v0.46/package.json
maschad: It's fixed specifically for purposes of strict reproducibility, see https://github.com/libp2p/test-plans/pull/288 for more details.

achingbrain: Is the lockfile not enough for this?

multidim-interop/impl/js/v0.46/package.json
maschad: They aren't necessary for functionality but rather they are there to explicitly state the versions of the dependencies being used, without requiring one to search the lockfile.

achingbrain: That sounds like they should be `"dependencies"` and not `"overrides"`?

multidim-interop/impl/js/v0.46/package.json
maschad: What would be the benefit of adding the range?

multidim-interop/impl/js/v0.46/package.json
maschad: They are dependencies of the dependency being installed i.e. `@libp2p/multidim-interop`  and according to [npm docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) since we are making sure that the same version of a package is used everywhere,  `overrides` is recommended.

maschad:
@achingbrain I've updated this PR to `v0.46.14` so that we can test out the webRTC fixes in https://github.com/libp2p/test-plans/pull/310

maschad:
friendly ping @achingbrain 

if there are no objections here @mxinden  I think we can proceed with merging so that we can enable the webRTC tests as mentioned in #310

achingbrain:
`libp2p@0.46.17` has shipped, it'd be good to test with that

multidim-interop/impl/js/v0.46/package.json
achingbrain: So you don't have to manually update patch release versions all the time.  The lockfile gives you repeatability in CI, the range gives you less to do in dev.

multidim-interop/impl/js/v0.46/package.json
achingbrain: Generally letting your dependencies manage the versions of the modules they depend on is recommended, otherwise it becomes very hard to reason about what is going on when things go wrong.

Issue #306, opened 2023-09-26
multidim-interop: rename to "transport"?


thomaseizinger:
With the upcoming additional interoperability tests, I was wondering whether we shouldn't rename the existing `multidim-interop` tests to `transport-interop`?

I am also open to other suggestions but I find "multidim" by itself to be too implementation specific. It tells us _how_ we are testing things but not what we are testing. Plus, in the upcoming hole-punching tests, I am using the same "multidim" technique.

Thoughts?

mxinden:
`transport-interop` is fine by me.

marten-seemann:
I agree, in fact I posted pretty much exactly this in my review the blog post: https://pl-strflt.notion.site/Multidimensional-Transport-Interoperability-6d2cc542309e413a930b4ec233b8cc5e.

thomaseizinger:
> I agree, in fact I posted pretty much exactly this in my review the blog post: [pl-strflt.notion.site/Multidimensional-Transport-Interoperability-6d2cc542309e413a930b4ec233b8cc5e](https://pl-strflt.notion.site/Multidimensional-Transport-Interoperability-6d2cc542309e413a930b4ec233b8cc5e).

Great minds think alike :blush: 

Any preferences of `transport-interop` vs `transport-multidim-interop`? Personally, I'd go for the former because it is shorter.

At present state, the hole punching tests only have one dimension (transports). One could imagine that we add a 2nd dimension for a different NAT type perhaps. Hence, I'd argue that all our interop tests have a kind of "multidim" aspect to them so it is redundant to restate that. Lastly, the fact that we are testing different versions against each other might also be seen as a dimension.

marten-seemann:
`transport-interop` sgtm.

PR #308, opened 2023-09-29
Rename `multidim-interop` to `transport-interop`

thomaseizinger:
For now, I've left the old action in place to allow us updating all repositories separately. Once they are updated we should remove the old `run-interop-ping-test` action.

Resolves: #306.


thomaseizinger:
cc @Menduist for `nim-libp2p`
cc @kevodwyer for `java-libp2p`

thomaseizinger:
@maschad @achingbrain I saw that js-libp2p depends on a dedicated package that also still refers to `multidim`. No pressure to rename that too, I just wanted to point it out! :)

maschad:
@thomaseizinger we can go through the process of [deprecating the npm package](https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions#deprecating-an-entire-package) and renaming it, as well as [renaming the script ](https://github.com/libp2p/js-libp2p/blob/master/interop/package.json#L49) which is [current test causing the failure](https://github.com/libp2p/test-plans/actions/runs/6345734371/job/17243601217?pr=308#step:3:1251) but if you don't want to be blocked by those tasks you should perhaps revert the [js dockerfile changes](https://github.com/libp2p/test-plans/pull/308/files#diff-4c6bdf2e70acc38c82364f9ac4d6e7f686774a4e7e1f7fb117fd42421a9dab6aR13) in the interim and leave a TODO.

thomaseizinger:
> @thomaseizinger we can go through the process of [deprecating the npm package](https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions#deprecating-an-entire-package) and renaming it, as well as [renaming the script ](https://github.com/libp2p/js-libp2p/blob/master/interop/package.json?rgh-link-date=2023-10-04T02%3A00%3A37Z#L49) which is [current test causing the failure](https://github.com/libp2p/test-plans/actions/runs/6345734371/job/17243601217?pr=308#step:3:1251) but if you don't want to be blocked by those tasks you should perhaps revert the [js dockerfile changes](https://github.com/libp2p/test-plans/pull/308/files#diff-4c6bdf2e70acc38c82364f9ac4d6e7f686774a4e7e1f7fb117fd42421a9dab6aR13) in the interim and leave a TODO.

Thanks for pointing that out! I've reverted that change and leave it up to you guys whether or not you want to go through the procedure of renaming that or leaving it as is.

mxinden (approved):
Works for me. Thank you!

achingbrain:
js-libp2p has an `@libp2p/multidim-interop` package that contains one interop test - `ping.spec.ts`.

The intention there was that it would one day include more than one test.

Is the expectation now that there would be multiple interop test modules each with one specific test?  This seems like quite a lot of scaffolding if so.

If we keep it as one module you can filter which tests are run with something like:

```console
$ aegir test -t node --grep 'ping'
```

or

```console
$ aegir test -t node --grep 'holepunch'
```

or whatever, which should cover our requirements here?

maschad:
That should work @achingbrain I agree it's not worth the extra scaffolding.

thomaseizinger:
@achingbrain At the moment there is a lot of duplication between the test _runners_ in here, correct. I think the best way to fix this is to see how they evolve. The upcoming holepunch tests already use a slightly different structure in how things are emitted and it was easy because I just forked the multidim implementation.

I have some thoughts on rewriting all this to deno to remove some of the nodejs boilerplate and hopefully be able to more easily share scripts and modules across the various test runners. I don't expect us to add more "tests" to the multidim (now transport) interop. The relay tests will likely be yet another test runner.

PR #314, opened 2023-10-24
fix(hole-punch): add local delay

thomaseizinger:
It seems that the tests are sometimes flaky in CI. Some debugging showed that the timing between the packets was a bit off so I am adding another delay from the "local machine" to the "router".

### DIFF
--- a/hole-punch-interop/src/generator.ts
+++ b/hole-punch-interop/src/generator.ts
@@ -5,7 +5,7 @@ import {ComposeSpecification} from "../compose-spec/compose-spec";
 import {sanitizeComposeName} from "./lib";
 import path from "path";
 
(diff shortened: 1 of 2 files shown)


hole-punch-interop/impl/rust/v0.52/Makefile
```diff
@@ -1,5 +1,5 @@
 image_name := rust-v0.52
-commitSha := d99fa2f87406cbd1a207b29f8cbcec32b7b180f7
+commitSha := 1ead41ed04ce89e50b11a078586e47d89e17d3c4
```
thomaseizinger, line 2: All the bugs around flakiness were in `rust-libp2p`. See the PR that this currently points to: https://github.com/libp2p/rust-libp2p/pull/4549

thomaseizinger:
Merging here to move https://github.com/libp2p/rust-libp2p/pull/4549 forward.

thomaseizinger:
I ran the tests in a matrix of 25 entries several times and could no longer observed any failures.

PR #315, opened 2023-10-25
ci: add dependabot

thomaseizinger:
Ensures we keep things up to date like the `rust-relay`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/hole-punch-interop/rust-relay"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "deps"
+    rebase-strategy: "disabled"


marten-seemann:
Please don't. This bot is super annoying, and we've disabled it on purpose in all go-libp2p repos.

thomaseizinger:
> Please don't. This bot is super annoying, and we've disabled it on purpose in all go-libp2p repos.

That is a matter of taste I guess. We've had good experiences with in in `rust-libp2p`. I can remove it for `github-actions` if you want but I'd like to have it for `rust-relay` to ensure we are using a recent / the latest version.

marten-seemann:
How frequently do you release rust-relay?

marten-seemann:
That’s a huge amount of noise, for a component that’s not even the focus of this test. I don’t think that’s justified.

marten-seemann:
Maybe we should switch to a more stable libp2p relay implementation? https://github.com/libp2p/go-libp2p-relay-daemon mostly just works and gets updated only occasionally.

thomaseizinger:
> That’s a huge amount of noise, for a component that’s not even the focus of this test. I don’t think that’s justified.

We can configure it to bundle all updates into one PR and it is already set to "weekly", so it would only be one PR per week.

> Maybe we should switch to a more stable libp2p relay implementation? [libp2p/go-libp2p-relay-daemon](https://github.com/libp2p/go-libp2p-relay-daemon) mostly just works and gets updated only occasionally.

We don't _have_ to update it but why let it bit-rot when we can also just keep it up to date? I don't mind at all which implementation we are using. The argument is the same for any implementation.

thomaseizinger:
I really didn't expect this to create so much friction so I'll just close this and update it manually occasionally.

mxinden:
> > That’s a huge amount of noise, for a component that’s not even the focus of this test. I don’t think that’s justified.
> 
> We can configure it to bundle all updates into one PR and it is already set to "weekly", so it would only be one PR per week.

In my eyes, that is a reasonable compromise. I don't think one pull request per week adds too much noise.

marten-seemann:
If you describe it as a compromise, there must be something on the other side of the equation. However, there's literally 0 we gain from updating the relay implementation if it's not broken, so any cost seems to high to pay for that. Again, we can use the go-libp2p-relay-daemon, which is well tested and just does its job without requiring weekly updates.

I will for sure unsubscribe from all notifications in this repo if this dependabot spam is enabled.

mxinden:
> If you describe it as a compromise, there must be something on the other side of the equation.

To me the benefits of regularly updating the relay code (i.e. "the other side of the equation") is:

- Reduced technical dept. Updating frequently and incrementally is in my eyes easier.
- Catching regressions early. Continuously testing with the newest version ensures we test what our users use in production.
- Potentially benefiting from future improvements, automatically having them trickle through the various dependents, instead of having to manually find and update them.

> Again, we can use the go-libp2p-relay-daemon, which is well tested and just does its job without requiring weekly updates.

I don't think the above benefits are specific to rust-libp2p's relay server implementation. I think they apply to any implementation, e.g. as well to the Golang implementation. I don't have an opinion on which one to use. Ideally we would eventually test with both.

> I will for sure unsubscribe from all notifications in this repo if this dependabot spam is enabled.

While I don't consider a weekly notification a lot of noise, your mileage might vary. Depending on your notification workflow, you can just either add a filter (GitHub notifications) or set-up some email client automation (e-mail client).

PR #316, opened 2023-10-25
feat(perf/rust): experiment with large yamux receive window

mxinden:
Adds experimental rust-libp2p using 16MB Yamux receive window.


mxinden:
This led to https://github.com/libp2p/rust-yamux/pull/176. Closing here since Rust's Yamux implementation now does auto-tuning :tada:

PR #317, opened 2023-10-26
feat(hole-punch): update rust `v0.52` to `master`

thomaseizinger:
The `v0.52` version string was really just a placeholder. Initially, I thought that we might merge https://github.com/libp2p/rust-libp2p/pull/4549 before the release cut-off for `v0.53` but that didn't happen.

To communicate this more clearly, we rename the current image to `rust-master`. We are also updating it to actually point _to_ a commit on master. Previously, it was pointing to a commit in https://github.com/libp2p/rust-libp2p/pull/4549. This new commit includes an important bug-fix (https://github.com/libp2p/rust-libp2p/pull/4747) that caused the QUIC hole-punch test to occasionally fail.

Once we cut the v0.53 release, this version should be promoted to that.

As a side-benefit, the new commit points to a version that includes the updated Dockerfile and allows the tests to be run on ARM64.

### DIFF
--- a/hole-punch-interop/impl/rust/master/Makefile
+++ b/hole-punch-interop/impl/rust/master/Makefile
@@ -1,5 +1,5 @@
-image_name := rust-v0.52
-commitSha := 1ead41ed04ce89e50b11a078586e47d89e17d3c4
+image_name := rust-master
+commitSha := 7517934a0f629309f83b69dc00eef44c44ddbc72
 
 all: image.json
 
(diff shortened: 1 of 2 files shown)


thomaseizinger:
cc @diegomrsantos Once this is merged, you should be good to go to run the entire test suite on ARM64 without any other patches.

hole-punch-interop/impl/rust/master/Makefile
```diff
@@ -1,5 +1,5 @@
-image_name := rust-v0.52
-commitSha := 1ead41ed04ce89e50b11a078586e47d89e17d3c4
+image_name := rust-master
+commitSha := 7517934a0f629309f83b69dc00eef44c44ddbc72
```
thomaseizinger, line 2:
```suggestion
commitSha := 461209ab1ba6b8e2a653ca354a5aad38802a35f9
```

thomaseizinger:
Merging here because it includes a bugfix that caused flakiness.

PR #319, opened 2023-10-27
wip: debug flaky QUIC hole-punch

thomaseizinger:


thomaseizinger:
OMG I found the bug. It is one we hit already previously: Closing a QUIC stream that has been closed be the remote already can result in an error ...

PR #320, opened 2023-10-27
feat(perf/rust-libp2p): test mtu path discovery

mxinden:
See https://github.com/libp2p/rust-libp2p/pull/4750.


marten-seemann:
To DPLPMTUD or DPLMTUPD, that's the question.

PR #324, opened 2023-11-07
docs: include required chunk size in perf readme

achingbrain:
I almost changed the default to something bigger because it improves the benchmark results..

It's probably worth stating explicitly what the chunk size is so every implementation uses the same value for consistency.


mxinden:
> because it improves the benchmark results..

Great. Then I suggest increasing it in js-libp2p. That said, worth investigating why js-libp2p is CPU bottlenecked on the chunk size?

achingbrain:
> Then I suggest increasing it in js-libp2p

The point of this PR is so that all implementations use the same chunk size so there is a better basis for comparison.  Is this not something we want?

PR #327, opened 2023-11-16
deps(rust-libp2p): point hole-punch tests to released version

thomaseizinger:
Now that we have a released version of `rust-libp2p` that includes the hole-punching tests, point the test suite to that version instead of a commit on `master`.

### DIFF
--- a/hole-punch-interop/versions.ts
+++ b/hole-punch-interop/versions.ts
@@ -10,9 +10,8 @@ export type Version = {
 
 export const versions: Array<Version> = [
     {
-        id: "rust-master",
-        transports: ["tcp", "quic"],
-        containerImageID: readImageId("./impl/rust/master/image.json"),
+        id: "rust-v0.53",
+        transports: ["tcp", "quic"]
     } as Version,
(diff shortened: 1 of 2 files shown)


hole-punch-interop/versions.ts
```diff
@@ -10,9 +10,8 @@ export type Version = {
 
 export const versions: Array<Version> = [
     {
-        id: "rust-master",
-        transports: ["tcp", "quic"],
-        containerImageID: readImageId("./impl/rust/master/image.json"),
```
mxinden, line 15: I am assuming that this is a left-over @thomaseizinger?

mxinden (approved):

hole-punch-interop/versions.ts
thomaseizinger: What do you mean with left-over?

mxinden: I am surprised that the field is removed (not altered) and thus assume that is not actually needed in the first place. Is that not the case @thomaseizinger?

thomaseizinger: The field was needed because the naming of the version ("master") did not conform with the default naming pattern of semver and thus the default resolution function of the JSON file would not have worked.

PR #328, opened 2023-11-21
experiment: try different js chunk sizes

achingbrain:


achingbrain:
Conclusion: no significant difference.

Uses 64k/128k/256k/512k/1024k chunks:

image

PR #342, opened 2024-01-24
fix: update Rust v0.52 hash to fix build error

thomaseizinger:
See https://github.com/libp2p/rust-libp2p/pull/5118.

### DIFF
--- a/transport-interop/impl/rust/v0.52/Makefile
+++ b/transport-interop/impl/rust/v0.52/Makefile
@@ -1,5 +1,5 @@
 image_name := rust-v0.52
-commitSha := 51070dae6395821c5ab45014b7208f15975c9101
+commitSha := 6fe72bde6b519d98d545bb6d5a2803b5790d277e
 
 all: image.json
 


mxinden (commented):
Bump works for me. But all `rust-v0.52` testcases are failing now.

thomaseizinger:
> Bump works for me. But all `rust-v0.52` testcases are failing now.

Yeah I noticed, what the heck. How can the glibc version not match when it is built literally in that container.

thomaseizinger:
> > Bump works for me. But all `rust-v0.52` testcases are failing now.
> 
> Yeah I noticed, what the heck. How can the glibc version not match when it is built literally in that container.

It appears the glibc version in distroless is not recent enough.

thomaseizinger:
https://github.com/GoogleContainerTools/distroless/issues/1342

thomaseizinger:
Hmm, should I use `debian-12` as the base image or back-port the static-linking stuff we already have on main in `rust-libp2p`?

See https://github.com/libp2p/rust-libp2p/blob/master/interop-tests/Dockerfile.native.

thomaseizinger:
> Hmm, should I use `debian-12` as the base image or back-port the static-linking stuff we already have on main in `rust-libp2p`?
> 
> See [libp2p/rust-libp2p@`master`/interop-tests/Dockerfile.native](https://github.com/libp2p/rust-libp2p/blob/master/interop-tests/Dockerfile.native?rgh-link-date=2024-01-24T16%3A28%3A14Z).

I think I'll move forward with the updated base image. The static linking might need some tweaks actually.

thomaseizinger:
And of course the glibc version in the selenium base image is also old.

thomaseizinger:
I am disabling the `chromium` tests for now to unblock this. I don't have the resources to investigate this right now.

PR #344, opened 2024-01-25
chore(interop): update to nim 1.6.16

diegomrsantos:
Because of nimble limitations - our package manager - the build can fail even though nothing changed. It can happen when new dependencies of libraries (that can be transitive) are released. We need to upgrade nim to fix the current build issue.


p-shahi:
Taking the liberty of merging here

diegomrsantos:
@p-shahi thanks a lot!

PR #361, opened 2024-05-05
fix: handle multiple strings that look like they might be JSON

achingbrain:
The test output can have occurrences of `{` that aren't the string we are looking for, eg:

```
dialer-1    | test browser
dialer-1    | [JavaScript Error: "Clear-Site-Data header found. Unknown value “"cache"”." {file: "http://127.0.0.1:34451/" line: 0}]
dialer-1    | [JavaScript Error: "Clear-Site-Data header forced the clean up of “cookies” data." {file: "http://127.0.0.1:34451/" line: 0}]
dialer-1    | [JavaScript Error: "Clear-Site-Data header forced the clean up of “storage” data." {file: "http://127.0.0.1:34451/" line: 0}]
```

These fail to parse so improve the results parsing by:

1. Try parsing every instance of `{...}` in the output, not just the first
2. Test the output for the `handshakePlusOneRTTMillis` and `pingRTTMilllis` keys

### DIFF
--- a/transport-interop/src/compose-stdout-helper.ts
+++ b/transport-interop/src/compose-stdout-helper.ts
@@ -15,30 +15,18 @@ export function dialerStdout(composeStdout: string): string {
 export function dialerTimings(dialerStdout: string): Object {
     let openBracket = dialerStdout.indexOf("{")
     let error
-
     while (true) {
-        if (openBracket === -1) {
-            break
-        }
-
+        if (openBracket === -1)  break
         const closeBracket = dialerStdout.indexOf("}", openBracket)
-        if (closeBracket === -1) {
-            throw new Error("Invalid JSON. No closing curly bracket found")
-        }
-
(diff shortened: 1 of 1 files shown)


MarcoPolo:
Why is this on stdout? Shouldn’t these logs be stderr?

achingbrain:
I’m not sure but the method that parses the output should be more robust.

achingbrain:
There's a fix for this specific issue at https://github.com/hugomrdias/playwright-test/pull/661 but I'm going to merge this because making the detection more robust will help us in the future.

We might want to think of a way to transmit results in a less fragile way but that can be done as a follow up if it's desirable.

MarcoPolo:
> We might want to think of a way to transmit results in a less fragile way but that can be done as a follow up if it's desirable.

Reserving stdout for the test results and stderr for everything else (as is convention) would work just fine. My guess is that the test framework we're using just dumps everything to stdout and doesn't distinguish.

Issue #364, opened 2024-05-22
js-libp2p tests failing in interop


MarcoPolo:
Not sure why yet, I haven't debugged this but I'm seeing every transport interop test with js-libp2p failing:

https://github.com/libp2p/test-plans/actions/runs/9182341165?pr=363
https://github.com/libp2p/test-plans/actions/runs/9182417914?pr=359
https://github.com/libp2p/go-libp2p/actions/runs/9181929857?pr=2807
https://github.com/libp2p/go-libp2p/actions/runs/9180086151?pr=2806
https://github.com/libp2p/go-libp2p/actions/runs/9179966225?pr=2805

I'll update here or in a PR as I learn more.

MarcoPolo:
Works on my machine (of course) with the following steps:

1. Checking out this repo
2. (cd transport-interop/impl/js/v1.6 && make)
3. npm ci
4. ./node_modules/.bin/ts-node testplans.ts -e --name-filter="js-v1.6 x js-v1.6"
5. Save that output to a compose.yml
6. docker compose up

diegomrsantos:
could it be related to this https://github.com/libp2p/js-libp2p/issues/2505 ?

diegomrsantos:
I can't reproduce this issue on my machine either.

MarcoPolo:
Unlikely to be related since it wouldn't explain why it happened suddenly.

I also pulled the exact image used by CI with:
```
curl https://s3.amazonaws.com/libp2p-by-tf-aws-bootstrap/imageCache/8d6bd5587fd7d648c4663358526240cb83225b8f7c4633854c17736b0b55892f-x86_64.tar.gz  | docker image load
```

and it still worked locally. I'm thinking this is something to do with the hosts and maybe the version of tools it's running.

MarcoPolo:
Fixed. See https://github.com/libp2p/test-plans/pull/359#issuecomment-2125963356

PR #365, opened 2024-05-22
fix: add js tests to this repo

achingbrain:
Trying to debug these tests in CI is incredibly difficult and since they depend on older versions of previously published modules, if something breaks due to a bug in released code it's very hard to fix it.

Instead follow the pattern in the perf tests and include the full test implementations here.

The changes here seem to make the js-libp2p transport interop CI job much happier!


MarcoPolo:
> Instead follow the pattern in the perf tests and include the full test implementations here.

👍 

Related: #364

transport-interop/impl/js/v0.45/.aegir.js
```diff
@@ -0,0 +1,117 @@
...
+          console.error('Error in redis command:', err)
+          res.writeHead(500, {
+            'Access-Control-Allow-Origin': '*'
+          })
+          res.end(err.toString())
+        }
+      }
+
+      const proxyServer = http.createServer(requestListener)
+      proxyServer.listen(0)
```
MarcoPolo, line 88: This is what fixed it. Previously this [was](https://github.com/libp2p/js-libp2p/blob/libp2p-v0.46.21/interop/.aegir.js#L71):

```
  await new Promise(resolve => { proxyServer.listen(0, 'localhost', () => { resolve() }) })
```

The reason this broke it is kind of funny. For whatever reason this now listens on only the IPv6 loopback address `::1`, but when we do the fetch:

```
 fetch(`http://localhost:...
 ```
 
 it resolves to the ipv4 loopback address at `127.0.0.1`.
 
 ---
 
 Apparently in this environment just doing `.listen(0)` listens on both `::1` and `127.0.0.1` (unless IPv6 is disabled and you only listen on 127.0.0.1).

MarcoPolo:
Can we either:
1. Merge #359 and then update v1.6 to include an in-repo test (or do the thing where we copy the test from the main repo like Rust and Go do)
2. Update this PR to include the in-repo test for the last two versions of js-libp2p.

I'd lean for option 1 since #359 should be good to go now after I fixed it with an ugly hack.

achingbrain:
My preference would still be to use released artefacts to run these tests, if nothing else because installing all the dev tools and building things takes time which will only get worse the more releases there are, but it seems that the runtime environment will change over time (ex, these tests used to pass, now they don't because of the IP4/IP6 thing identified above) so this is probably the pragmatic thing to do.

----

I see #359 got merged, I didn't realise we were going to add every minor release too - TBH I'm not sure of the utility of testing `1.3`, `1.4`, `1.5` etc because of how js-libp2p is released:

The transport and identify/ping protocol modules do all the heavy lifting here so are the ones that would be likely to introduce bugs that break things but `js-libp2p` does not always have a corresponding release for each version of these because the monorepo publish step only releases modules that have changed in order to minimise dependency churn for users.

That is to say, we could be on `js-libp2p@1.6` until the end of the year but there could be many `@libp2p/tcp` releases in the interim, which would not be reflected here.

Instead we could just have the latest `js-libp2p` major tested here (e.g. whatever the latest released `1.x.x` version of `libp2p` and all `@libp2p/*` modules is).  The js-libp2p repo tests against that in CI, then we can update the lockfile here after each monorepo publish.  That should act like a kind of ratchet in that if the current monorepo CI is compatible with the latest released version, and if it's running the same test, then it should be compatible with previous versions too.

For total certainty we'd need have to have versions here for every monorepo publish but that would lead to many jobs here and ever increasing build times.

`1.x` should be good enough, what do you think?

transport-interop/impl/js/v0.45/.aegir.js
achingbrain: Yeah, it's generally better to listen without specifying a host, then it'll bind to all available addresses/interfaces, then the host can resolve `"localhost"` to whatever it wants and we can have a quiet life 😄

MarcoPolo:
> My preference would still be to use released artefacts to run these tests, if nothing else because installing all the dev tools and building things takes time which will only get worse the more releases there are, but it seems that the runtime environment will change over time (ex, these tests used to pass, now they don't because of the IP4/IP6 thing identified above) so this is probably the pragmatic thing to do.

I'll defer to your opinion on how to do the js-libp2p implementation here. I'll just mention that the time for installing/building is probably a non-issue since we cache the builds anyways. Go and Rust (and others) only build the test once. I would suggest considering following how Go/Rust implement the test in the main repo and copy it during a build. That gives you the benefit of having the test defined alongside the implementation and being able to debug things like this, or, the more likely failure, an incompatibility.

> I see #359 got merged, I didn't realise we were going to add every minor release too - TBH I'm not sure of the utility of testing `1.3`, `1.4`, `1.5` etc because of how js-libp2p is released:

It's only testing 1.5 and 1.6 (the last two versions). The other versions are vestiges that I haven't remove yet.

> ...
> Instead we could just have the latest `js-libp2p` major tested here (e.g. whatever the latest released `1.x.x` version of `libp2p` and all `@libp2p/*` modules is). The js-libp2p repo tests against that in CI, then we can use dependabot to update the lockfile here after each monorepo publish. That should act like a kind of ratchet in that if the current monorepo CI is compatible with the latest released version, and if it's running the same test, then it should be compatible with previous versions too.
> 
> `1.x` should be good enough, what do you think?

That sounds fine by me. As long as a given commit is reproducible I'm happy.

achingbrain:
> I would suggest considering following how Go/Rust implement the test in the main repo and copy it during a build.  That gives you the benefit of having the test defined alongside the implementation and being able to debug things like this, or, the more likely failure, an incompatibility.

The test in js-libp2p main should be compatible with the currently released major (v1 at the time of writing), so on the face of it, it makes sense to copy over during the build.  But if we release v2, it may become incompatible with v1, so we should copy the v1 test from the latest `v1.x.x` tag and not main, which means we can't change it in response to external breakage (like the `localhost` docker thing here) and we're back to the current problem.

We can always revisit this later if the situation changes.

PR #381, opened 2024-06-11
go-libp2p holepunch interop

MarcoPolo:
This doesn't interop in all cases with rust!

Part of it is this issue: https://github.com/libp2p/rust-libp2p/issues/5466

And part of it is still unknown to me and needs more debugging. I wanted to open this to share my work

Thanks @thomaseizinger for making this! works great.

Future versions will have this impl live in go-libp2p. For now though it's easier for debugging if this is here.


MarcoPolo:
The two issues we fixed in go-libp2p that let us pass these tests:

1. Don't observe the relay's address.
2. Fix the roles when doing a tcp sim-connect

MarcoPolo:
We need to release those fixes before we can merge this.

PR #429, opened 2024-08-01
fix buildx and buildkit versions

diegomrsantos:


dhuseby:
Moved this to here: https://github.com/libp2p/test-plans/pull/700

Thank you for the contribution but next time please allow maintainers to edit.

PR #465, opened 2024-09-03
chore: parameterise s3 build cache setup

galargh:
As we're setting up a new cache bucket, we'd like to be able to control its' configuration via GitHub vars/secrets fully.

### DIFF
--- a/hole-punch-interop/helpers/cache.ts
+++ b/hole-punch-interop/helpers/cache.ts
@@ -1,4 +1,4 @@
-const AWS_BUCKET = process.env.AWS_BUCKET || 'libp2p-by-tf-aws-bootstrap';
+const AWS_BUCKET = process.env.AWS_BUCKET;
 const scriptDir = __dirname;
 
 import * as crypto from 'crypto';
@@ -72,6 +72,11 @@ async function loadCacheOrBuild(dir: string, ig: Ignore) {
     files = files.map(f => path.join(dir, f))
     const cacheKey = await hashFiles(files)
     console.log("Cache key:", cacheKey)
+
+    if (AWS_BUCKET === "") {
+        console.log("Cache not found", new Error("AWS_BUCKET not set"))
+        return
+    }
 
     if (mode == Mode.PushCache) {
         console.log("Pushing cache")
--- a/hole-punch-interop/helpers/cache.ts
+++ b/hole-punch-interop/helpers/cache.ts
@@ -73,7 +73,7 @@ async function loadCacheOrBuild(dir: string, ig: Ignore) {
     const cacheKey = await hashFiles(files)
     console.log("Cache key:", cacheKey)
 
-    if (AWS_BUCKET === "") {
+    if (!AWS_BUCKET) {
         console.log("Cache not found", new Error("AWS_BUCKET not set"))
         return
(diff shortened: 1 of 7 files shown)


AlejandroCabeza:
Are you planning to make the bucket publicly accessible? As Protocol Labs' was previously?

achingbrain (approved):

hole-punch-interop/helpers/cache.ts
```diff
@@ -73,6 +73,11 @@ async function loadCacheOrBuild(dir: string, ig: Ignore) {
     const cacheKey = await hashFiles(files)
     console.log("Cache key:", cacheKey)
 
+    if (AWS_BUCKET === "") {
```
achingbrain, line 76: This will catch `undefined`, `null` and `''`.  As written it'll only catch `''`.

```suggestion
    if (!AWS_BUCKET) {
```

transport-interop/helpers/cache.ts
```diff
@@ -62,6 +62,11 @@ switch (modeStr) {
             const cacheKey = await hashFiles(files)
             console.log("Cache key:", cacheKey)
 
+            if (AWS_BUCKET === "") {
```
achingbrain, line 65: This will catch `undefined`, `null` and `''`.  As written it'll only catch `''`.

```suggestion
            if (!AWS_BUCKET) {
```

diegomrsantos:
@achingbrain complementing @AlejandroCabeza question, is there any cache that can be used at the moment, or from now on projects using this repo will need to set up their own individually?

achingbrain:
I don't know, sorry - you may need to stand your own s3 bucket up.

If that adds too much overhead, you could apply for a grant from the libp2p foundation to contribute to a shared infrastructure?

AlejandroCabeza:
> I don't know, sorry - you may need to stand your own s3 bucket up.
> 
> If that adds too much overhead, you could apply for a grant from the libp2p foundation to contribute to a shared infrastructure?

Alright, thanks for the answer!

transport-interop/helpers/cache.ts
```diff
@@ -85,18 +90,17 @@ switch (modeStr) {
...
-                        }
-                    } else {
-                        console.log("Cache not found")
+                    if (!AWS_BUCKET) {
+                        throw new Error("AWS_BUCKET not set")
+                    }
+                    const cachePath = fs.mkdtempSync(path.join(os.tmpdir(), 'cache'))
+                    const archivePath = path.join(cachePath, 'archive.tar.gz')
+                    const dockerLoadedMsg = child_process.execSync(`aws s3 cp s3://${AWS_BUCKET}/imageCache/${cacheKey}-${arch}.tar.gz ${archivePath} && docker image load -i ${archivePath}`).toString();
+                            const loadedImageId = dockerLoadedMsg.match(/Loaded image( ID)?: (.*)/)[2];
```
achingbrain, line 99:
```suggestion
                    const loadedImageId = dockerLoadedMsg.match(/Loaded image( ID)?: (.*)/)[2];
```

achingbrain (commented):
LGTM but CI is failing

achingbrain (approved):

PR #520, opened 2024-10-22
Add go-libp2p@v0.37.0

MarcoPolo:


MarcoPolo:
Build without secrets will fail as the public cache has been turned off for cost reasons. Maybe it would make sense to make releases and turn the outputs of the docker builds into release artifacts so that forks can use them?

MarcoPolo:
Another idea is to use a container registry.

PR #603, opened 2025-02-04
chore: build webkit images during test run

achingbrain:
In order to test webkit interop, build docker images with the `BROWSER` build arg set to `webkit`.

I think we might need this before the tests in https://github.com/libp2p/js-libp2p/pull/2541 will actually run.


achingbrain:
https://github.com/libp2p/js-libp2p/pull/2541 has been merged, all WebKit transport-interop tests now pass apart from WebTransport.

PR #606, opened 2025-02-04
chore: test Firefox WebTransport

achingbrain:
WebTransport has been supported in Firefox since 125, so we can test interop now.


achingbrain:
That said, it's probably broken until 136 is released and Playwright upgrades to it - likely some time in March 2025

PR #614, opened 2025-02-18
Update transport interop for go-libp2p v0.40 & v0.39

MarcoPolo:


sukunrt:
lotus is on v0.37 and prysm is on v0.36. Should we keep those till they upgrade?

PR #642, opened 2025-04-23
specify known-good image by hash to force the correct version

dhuseby:
for some reason the incorrect nim docker image is being pulled in other PRs. i am testing if specifying a known-good version of the image via the hash will fix this.


dhuseby:
So this worked and fixes the nim builds.

PR #650, opened 2025-05-17
adding message dissemination rate vs time plots

hopinheimer:
@MarcoPolo  small effort toward adding more plots for analysis


MarcoPolo:
Could you try refactoring this slightly to avoid the duplication in parsing the log files?

Ideally we'd have a type and a function that parses the log files and returns a value of that type. Then different analysis functions accept that type of value and plot/analyze it.

If you don't get to this, I can take it

hopinheimer:
Oh that makes sense! let me take a stab at it and will ping you if I'm stuck

PR #654, opened 2025-05-29
transport interop: Bump go version fix Go Makefiles

MarcoPolo:
Adds go-libp2p v0.41.1, removes go-libp2p versions older than v0.40.0.

Updates the makefile to not cause rebuilds. We incorrectly set the verify-checksum step as a dependency to building the image. Since that was a phony target it would always cause the image to rebuild. Now we have the verify step as a function that is called.

@dhuseby This should fix any rebuild you were seeing for Go implementations. I was able to run make from the transport-interop directory and not see any rebuilds. The one caveat is that it seems like the Nim builder is broken. This means that setup was not reproducible, so we should fix that in a another PR.


transport-interop/impl/go/v0.39/version.lock
sukunrt, line 1: lotus is on this version: 
https://github.com/filecoin-project/lotus/blob/master/go.mod

transport-interop/impl/go/v0.36/version.lock
sukunrt, line 1: prysm is on this version: 
https://github.com/OffchainLabs/prysm/blob/develop/go.mod

This is now really old though.

sukunrt (approved):

MarcoPolo:
I think we should commit to interop testing the last two versions on each. Otherwise we'll always be hesitant to keep this up to date due to increasing testing times.

sukunrt:
last two is way too aggressive. Blockchain update times are rather long. Maybe a timeframe like 6 months?

MarcoPolo:
That's fair 👍 

fwiw I don't see this as us dropping support for previous versions. Just not explicitly testing interop on every PR. 

I think I can also make these interop tests faster, and then we can worry less about adding to testing time.

PR #656, opened 2025-06-03
feat(nim-libp2p): update transport-interop to v1.10.x

richard-ramos:
cc: @vladopajic, @kaiserd, @gmelodie

Updates the nim-libp2p version used for transport-interop.
This makes the [mainv1.nim] .code match https://github.com/vacp2p/nim-libp2p/tree/master/interop/transport
Also fixes a compilation issue happening due to an update in nim-chronicles.


richard-ramos:
In the dockerfile, there's a `nimble install_pinned` instruction. What this does is install specific commits for the dependencies, which are obtained from this file https://github.com/vacp2p/nim-libp2p/blob/master/.pinned

richard-ramos:
this happened because nim-chronicles introduced a breaking change in v0.11.0 ([commit](https://github.com/status-im/nim-chronicles/commit/a537556963bb42e048c7f7229456995355d360e5)), and nim-libp2p had its dependency set as "chronicles >= 0.10.3" ([link](https://github.com/vacp2p/nim-libp2p/blob/c3dae6a7d4d9961f1e99b14bcd996422e68b674b/libp2p.nimble#L12)), which allowed v0.11.0 to be pulled.

the latest version of nim-libp2p fixes this by capping the version: `"chronicles >= 0.10.3 & < 0.11.0"`

oddly, even though we use a .pinned file, `nimble`, the package manager from nim, installs dependencies before executing any tasks in libp2p.nimble (libp2p package definition file). so it tries to fetch the broken version first, fails, and never reaches the code that would download the specific dependency commits

PR #659, opened 2025-06-05
feat: add pref nim libp2p (working pr)

vladopajic:
this pr will hold work in progress - to be easily cherry picked to maintainers pr.


vladopajic:
the only folder that we are going to use from now on is `v1.10` that corresponds to `nim-libp2p` version used in perf.

PR #667, opened 2025-07-03
Add custom graph option for custom graph generation

wbjin:
The goal of this PR is to allow user defined network graphs for the gossipsub-interop testing using shadow. It branches off of https://github.com/libp2p/test-plans/pull/648 and will need to be merged after this.

It allows a user to
- Specify a node's upload and download bandwidth
- Specify latencies between specific nodes

Changes
- Add new arguments for specifying graph type and file paths for the "custom" option
- Modify `network_graph.py` to support the "random" graph generation (the existing one) and "custom" graph generation.

The graph is specified with two csv files, one for nodes and one for edges (more in the Readme). This change is part of an ongoing effort to replicate mainnet as close as possible. As we get more accurate latencies between nodes and the available bandwidth on each node, we should be able to recreate the libp2p layer of the mainnet network.

TODO

- [ ] Address PR feedback

### DIFF
--- a/gossipsub-interop/README.md
+++ b/gossipsub-interop/README.md
@@ -68,6 +68,25 @@ example, the output folder is
   - analysis_*.txt: A text file containing a high level analysis of the 3 key results
   - Charts visualizing the results.
 
+### Network graph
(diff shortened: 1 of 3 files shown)


MarcoPolo:
@wbjin , I didn't close this on purpose. It was a side effect of merging the other PR. Can you rebase on master and reopen the PR please?

PR #680, opened 2025-09-17
Add jvm-libp2p to protocol interop suite

lucassaldanha:
Adding the standard jvm-libp2p (https://github.com/libp2p/jvm-libp2p) to the test suite.

For now, I am only adding a subset of transport and muxers to validate that we are properly running the tests. Once we are happy with it, I'll add other options.

### DIFF
--- a/transport-interop/versionsInput.json
+++ b/transport-interop/versionsInput.json
@@ -308,5 +308,20 @@
     ],
     "secureChannels": ["tls", "noise"],
     "muxers": ["yamux"]
+  },
+  {
+    "id": "jvm-v1.2",
+    "transports": [
+      "tcp",
+      "quic-v1"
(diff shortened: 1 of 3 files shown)


transport-interop/versionsInput.json
```diff
@@ -288,8 +288,13 @@
...
     ],
-    "secureChannels": ["tls", "noise"],
-    "muxers": ["yamux"]
+    "secureChannels": [
+      "tls",
+      "noise"
+    ],
+    "muxers": [
+      "yamux"
+    ]
```
MarcoPolo, line 297: Can we avoid reformatting this file please?

lucassaldanha: Sure, will revert it!

PR #682, opened 2025-09-25
Add ws to jvm + remove java

StefanBratanov:
- Add ws since it's supported
- Removed java folder since jvm is supported now and tracks `jvm-libp2p`


StefanBratanov:
> I think the same WS pairs failed after rerunning: https://github.com/libp2p/test-plans/actions/runs/17999148931

I tried disabling it in the last commit. Maybe it needs to be rerun?

PR #684, opened 2025-09-26
Test Partial Message in interop tester

MarcoPolo:
adds Go implementation of partial messages and changes to the test runner to support it.


MarcoPolo:
This is ready to merge. We can have follow up PR for future clients and any other changes we need.

PR #685, opened 2025-09-26
add rust implementation for partial messages

jxs:


gossipsub-interop/rust-libp2p/src/script_instruction.rs
```diff
@@ -90,6 +90,24 @@ pub enum ScriptInstruction {
     InitGossipSub {
         gossip_sub_params: Box<GossipSubParams>,
     },
+    #[serde(rename = "addPartialMessage", rename_all = "camelCase")]
+    AddPartialMessage {
+        #[serde(rename = "r#type")]
```
dariusc93: correct me if Im wrong but wouldnt this serialize be named as `r#type` instead of `type`?

jxs: Hi, Darius, yes you are correct thanks for noticing!

PR #733, opened 2025-11-15
fix: change package.json to match package-lock.json for transport-interop js/v2.x

bismuth01:
`package.json` had updated the version of packages which resulted in `package-lock.json` mistmatch and ultimately `npm ci` failing during docker container creation. Just updating the `package-lock.json` with `npm i` did not work due to the changes in the updated packages.
I have manually rolled back the package versions after looking at the `package-lock.json` and now it works as expected.


acul71:
@sumanjeet0012 @bismuth01 

# Review of PR #733: fix package.json/package-lock.json mismatch for js/v2.x

**Reviewer:** @acul71  
**Author:** @bismuth01  
**Requested merge by:** @manusheel

---

## Summary

Hi @bismuth01! I've reviewed PR #733 and **I recommend merging this PR**. It fixes a real issue in upstream master where `npm ci` fails for `js/v2.x` due to a mismatch between `package.json` and `package-lock.json`.

## Verification

I verified the issue exists in upstream master:

- **Upstream master (d8f8145)**: `npm ci` fails with multiple version mismatches
  - `package.json` has newer versions (^16.1.3, ^3.2.15, ^2.8.9, etc.)
  - `package-lock.json` has older versions locked (16.0.3, 2.1.5, 2.8.0, etc.)
  - Error: `Invalid: lock file's @chainsafe/libp2p-noise@16.0.3 does not satisfy @chainsafe/libp2p-noise@16.1.5`

- **PR #733 branch**: Fixes the mismatch by rolling back `package.json` versions to match `package-lock.json`

## Scope

I also checked the other JavaScript versions to ensure this fix is properly scoped:

- ✅ **js/v1.x**: `npm ci` succeeds - no issues
- ❌ **js/v2.x**: `npm ci` fails - **this PR fixes it**
- ✅ **js/v3.x**: `npm ci` succeeds - no issues

The PR is correctly scoped to only fix `js/v2.x`, which is the only version with this problem.

## Approach

Your approach of rolling back `package.json` to match `package-lock.json` is valid. As you mentioned, updating `package-lock.json` with `npm install` didn't work due to breaking changes in the newer packages. Rolling back is the right solution in this case.

## Minor Question

The addition of `"peer": true` in `package-lock.json` - is this intentional or a side effect of npm operations? Either way, it's a minor metadata change and shouldn't cause any issues.

## Recommendation

**@manusheel, please merge this PR.** It fixes a real build failure that prevents Docker images from being built for `js/v2.x`. The fix is correct, properly scoped, and necessary for the CI/CD pipeline to work.

## Follow-up Action Required

Once this PR is merged and JavaScript builds are confirmed to work, **the `transport-interop/test-ignore.txt` file should be cleared** (made empty).

**Current state:**
- The file contains: `js-v1.x|js-v2.x|js-v3.x`
- This causes all JavaScript tests to be excluded from the full test suite in GitHub Actions

**Why this matters:**
- The `test-ignore.txt` file is read by `.github/workflows/transport-interop-full.yml` 
- It's used to automatically exclude tests via the `--test-ignore` parameter
- With JavaScript builds now working (after this PR), JavaScript tests should be included in the test suite

**Action needed:**
- Clear the contents of `transport-interop/test-ignore.txt` (make it an empty file)
- This will allow JavaScript tests to run in the full interop test suite

---

## Technical Details

### Changes Made
- Rolled back 13 package versions in `package.json` to match `package-lock.json`
- Added `"peer": true` flag to one dependency in `package-lock.json`

### Key Version Rollbacks
- `@chainsafe/libp2p-noise`: ^16.1.3 → ^16.0.0
- `@libp2p/circuit-relay-v2`: ^3.2.15 → ^2.1.1
- `libp2p`: ^2.8.9 → ^2.1.4
- `aegir`: ^47.0.16 → ^44.1.1
- `redis`: ^5.5.6 → ^4.6.10
- And 8 more packages...

### Why This Happened
The mismatch occurred because `package.json` was updated (likely via automated dependency updates) but `package-lock.json` wasn't regenerated. When `npm ci` runs in the Dockerfile, it strictly validates that the lock file matches `package.json`, causing the build to fail.

---

**Status:** ✅ Approved - Ready to merge

seetadev:
@bismuth01 , @acul71 : Thank you. This PR is indeed ready for final review and merge. Appreciate your efforts.

PR #755, opened 2025-12-12
fix: support git submodules in Docker builds for c-v0.0.1

acul71:
- Add support for requiresSubmodules flag in build-images.sh
- Use git clone instead of zip download when submodules are required
- Initialize and cache git clones with submodules properly initialized
- Remove c-v0.0.1 from failing alias in impls.yaml since it's now fixed

Fixes the Docker build failure where picoquic submodule was missing when building from GitHub zip archives (which don't include submodules).

### DIFF
--- a/.github/workflows/transport-interop-pr.yml
+++ b/.github/workflows/transport-interop-pr.yml
@@ -116,7 +116,7 @@ jobs:
               cd - > /dev/null
               if [ -n "$GITHUB_IMPLS" ]; then
                 echo "test-select=$GITHUB_IMPLS" >> $GITHUB_OUTPUT
-                echo "test-ignore=!$GITHUB_IMPLS" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


dhuseby:
let's remove the snapshot downloading.

dhuseby:
i started another PR: https://github.com/libp2p/test-plans/pull/760

we can keep your PR, I'm just trying out using git to clone all repos instead of downloading zip snapshots to see how it goes. if downloading zip snapshots is better, then we'll keep your logic here where it clones only when submodules are needed.

dhuseby:
@acul71 this PR is all wrong. sorry, i reorganized the structure recently. the code that handles downloading snapshots and building from github is all in the file `scripts/lib-image-building.sh`. specifically here: https://github.com/libp2p/test-plans/blob/382e1ce71a78b6b5d41d34c6ba356b0faa6d9001/scripts/lib-image-building.sh#L69

if we're going to change the behavior we need to change it there.

PR #760, opened 2025-12-14
Fix/docker c v0.0.1

dhuseby:
I started with @acul71 's PR here: https://github.com/libp2p/test-plans/pull/755 and merged the latest master into it. I also reworked the logic so that it always uses git clone instead of downloading snapshots. Let's see how that does.


dhuseby:
this PR is based off of @acul71's other PR that doesn't fix the problem correctly. I'm closing this one now.

PR #773, opened 2026-01-09
dhuseby/refactor/perf-and-transport-and-hole-punch

dhuseby:
This is a major refactor of the whole system. I did not touch any of the gossipsub testing though, just perf, hole-punch, and transport.

Key improvements:
- Everything is POSIX bash
- All variables are properly/safely used (e.g. "${FOO}")
- All common functions moved to scripts in the `lib/` subdir
- Test matrix generation is now sharded and parallel with no disk I/O in the nested loops
- Test execution is parallel (except perf, it forces 1 worker on purpose) with common global services (i.e. Redis) started before and shutdown after all of the individual tests
- Test folders only hold test-specific data and functions. This significantly simplifies tests.
- All test applications take a common set of environment variables plus test-specific environment variables.
- All test applications output YAML results to a per-test results file
- Added the ability to patch a remote test application before building the docker image. This enables quick debugging and iteration and landing the update in test-plans BEFORE the fixes land in the implementation repo
- All three tests have a common structure with very similary code. The per-test scripts are also very similar with a common set of steps
- Clean, uniform formatting of output
- Better debug output when `--debug` is passed as an option. This also sets the DEBUG environment variable for test applications to respond to

### DIFF
--- a/hole-punch/README.md
+++ b/hole-punch/README.md
@@ -671,7 +671,7 @@ Hole-punch tests have unique characteristics:
 <!-- TEST_RESULTS_START -->
 # Hole Punch Interoperability Test Results
 
-## Test Pass: `hole-punch-084517-15-01-2026`
+## Test Pass: `hole-punch-030938-16-01-2026`
 
 **Summary:**
 - **Total Tests:** 5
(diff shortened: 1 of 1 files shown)


dhuseby:
alright...it's mostly done. perf, transport, and hole-punch tests all run. i fixed the baseline tests. documented everything, removed unused code. i wrote a peer and relay in rust for the hole-punch tests but it doesn't seem to work.

This is ready for a review.

PR #815, opened 2026-02-16
backwards compatibility fixes

dhuseby:
This PR adds backwards compatiblity to the transport interop tests while preserving most of the optimizations made in the refactor. This adds the following:

- Implementations using the old interface are marked with "legacy: true" to signal that it uses lower case environment variables, does not isolate Redis keys, and outputs JSON results.
- Adds a tiny Redis proxy whenever the dialer or listener is "legacy: true". The Redis proxy takes the TEST_KEY and translates the Redis key from the legacy implementation into a properly isolated Redis key and request to the global Redis server

The net effect is that the only failing tests are due to upstream bugs. There are 13 different upstream bugs that cause test failures.

Signed-off-by: Dave Grantham <dwg@linuxprogrammer.org>


acul71:
@dhuseby If I run the tests with more than 15 workers docker run out of network addresses and many tests fail

acul71:
> 3\. This min-redis thing can be used to keep listener test implementations alive for multiple interop tests, so you only have to restart dialers. This could significantly speed up tests.


**Re: per-listener compose with sequential dialers (@MarcoPolo  / @dhuseby )**

Some concerns with the “one listener, many sequential dialers” approach:

**1. Listener crash → cascade of failures**  
If the listener crashes or misbehaves mid-run, every subsequent dialer in that 1compose will fail (connection refused, timeouts, etc.). We’d see many failed tests that all share the same root cause (the listener), which makes it harder to tell “listener X is broken” from “dialer Y is broken.” Some way to detect listener health between dialers (or to treat “same listener, many dialers failed in a row” as one incident) would help.

**2. Logging and analysis**  
With one compose = one listener + N dialers, we no longer have a single log/artifact per test. Listener and dialer logs are mixed in one run, so correlating a failure to a specific (listener, dialer) pair and scripting over “one result per test” gets harder. If we go this direction, we’d need a clear convention (e.g. one log file per service, or per logical test, and a consistent test id like `listener-id x dialer-id`) so CI and humans can still analyze results easily.

**3. Listeners assume a single dialer**
Right now listener implementations are written for one dialer: they accept one connection, run the test, then exit (or don’t expect another). To reuse one listener for multiple sequential dialers, every implementation would need to be refactored to loop (e.g. wait for dialer → handshake/ping → disconnect → wait for next dialer, repeat). That’s a non-trivial change across all language repos (Go, Rust, JS, etc.) and would have to be done before this model can be used.

I’m not against the idea for speed; just flagging these trade-offs so they can be considered in the design.

MarcoPolo:
> The discussion on this is causing additional friction. @dhuseby can you please revert back to the last widely-compatible version of interop tests, and make sure that works easily for each major implementation?
> 
> In the future, breaking changes (like frameworks & case handling) should have agreement from all major maintainers before proceeding.
> 
> cc @momack2 @johannamoran

I'll link this discussion as well: https://github.com/libp2p/test-plans/discussions/803

I added my comments above in an effort to avoid dealing with two separate test implementation interfaces, but I still believe the bash rewrite was a mistake for reasons listed in that discussion.

I will not not review this PR as I both don't want to extend any of the bash code, nor do I want to review `858 insertions(+), 307 deletions(-)` of bash changes.

acul71:
Hello @mishmosh 
> The discussion on this is causing additional friction. @dhuseby can you please revert back to the last widely-compatible version of interop tests, and make sure that works easily for each major implementation?

With this PR (**I can review it, eventually**) both world (old and new), **can live together**. 
**This PR make already possible this**, we need just to make some more tests, and add some documentation. The "old" implementations, _don't need to know a thing_, because they should work right away (We just need to modify one file `images.yaml` that is trivial) 

About me, I don't feel additional friction, we were just trying to improve the test-plans for speed, and we were just exploring what the options are (thanks @MarcoPolo for your new ideas), and what it would take to implement them. It can be addressed in a following PR, if the case. 

The @dhuseby new refactor, makes several improvements, 
- **The docker images are smaller** (without node packages needed for the test engine)
- **We have artifacts of the tests so everybody, downloading a tar file, is capable of run the same CI/CD run on his local machine** without having to build many docker images (it could take more of 30 minutes, some other contributors in Africa with bad internet and low grade hardware, it takes all day long :-)
- **A wide numbers of options to filter** the tests and implementation to run, very useful in **debug**
- **Less Redis docker images to spawn**, 2 in total vs 1 x each test (so the improvement is astonishing, and **more tests in parallel can be run**) 

My recommendation is to merge this PR, once the minimal work of testing and adding documentation is done). Also because at this point, My feeling is that reverting back to the old model, will take more work, and will cancel all the benefits added with this new re-factorization

PR #829, opened 2026-02-17
Fix(dotnet-perf): Correct multiaddr retrieval and dialer initialization bug

K-21:
**Summary**

It fixes an issue in the dotnet performance benchmarks related to the dialer logic in `Program.cs.`

**Problem**

The dialer was correctly retrieving the listener multiaddr from Redis. However, when attempting to dial, an exception was thrown indicating that no multiaddr was supplied.

This caused the benchmark execution to fail even though the multiaddr was being stored and retrieved correctly.

**Root Cause**

Although the multiaddr was fetched successfully from Redis, it was not being properly passed to the dialer initialization logic. As a result, the dial operation was executed with a null or empty multiaddr.

**Fix**

- Ensured the retrieved multiaddr is correctly validated before dialing.
- Properly passed the multiaddr into the dialer logic.
- Added safeguards to prevent dialing with an empty or null multiaddr.

**Result**

The dialer now successfully connects to the listener using the multiaddr retrieved from Redis, and the benchmark runs without throwing the previous exception.

### DIFF
--- a/perf/images/dotnet/v1.0/Program.cs
+++ b/perf/images/dotnet/v1.0/Program.cs
@@ -86,16 +86,16 @@ class Program
         // Route to listener or dialer based on IS_DIALER
         if (isDialer)
         {
-            await RunDialer(peerFactory, logger, redisAddr, transport, uploadBytes, downloadBytes,
+            await RunDialer(peerFactory, logger, redisAddr, transport, testKey, uploadBytes, downloadBytes,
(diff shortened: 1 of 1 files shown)


K-21:
@dhuseby @flcl42 @seetadev 
This fixes the dialer multiaddr issue in dotnet perf — would appreciate your review.

flcl42 (approved):

PR #831, opened 2026-02-18
fix(dotnet-perf): fix dialer bug affecting benchmark runs

K-21:
### **Fix**  [#829 ](https://github.com/libp2p/test-plans/pull/829)

It extends the changes introduced [#829 ](https://github.com/libp2p/test-plans/pull/829)

The original PR resolved a critical issue in the .NET performance benchmarking suite: although the listener multiaddr was correctly stored and fetched from Redis, it was never passed properly into the dialer logic, resulting in dialing attempts without any target address and causing benchmark failures (null/empty multiaddr exceptions).

In addition to preserving that fix, this extended description explains the scope of changes and why they matter:

**Background & Root Cause**

In the .NET performance benchmarks, the benchmarking flow includes:

- Storing the listener multiaddr (network address of the service) in a shared store (Redis).
- Retrieving that multiaddr when initiating a dialer (client) process.
- Dialing the listener from the benchmark runner to measure performance.

However, while the multiaddr was being stored and fetched correctly, it was not passed into the dialer initialization — leading to a failure at runtime where the dialing attempt had no valid multiaddr to connect to. Because the benchmark logic depended on this connection attempt succeeding, the entire benchmark run would throw exceptions and fail prematurely.

**Enhacements** 

1. Correct multiaddr handling

- Properly validate the multiaddr retrieved from Redis before attempting a connection.
- Ensure that the valid multiaddr is correctly passed into the dialer initialization logic.
- Guarantees that no dial call occurs when the multiaddr is null or empty (protects against runtime exceptions).

2. Improved Dialer Initialization

- Updated initialization logic in `Program.cs` to include explicit checks and assignments.
- Ensures that the dialer receives the correct multiaddr up front, preventing misconfiguration and making the behavior deterministic.

3. Additional Safeguards and Resiliency

- Added defensive checks to fail early if the multiaddr format is invalid (e.g., not a valid libp2p multiaddr).
- Enhanced error output to make benchmark logs more useful when diagnosing failures (e.g., clear reasons for failed initialization).Updated initialization logic in Program.cs to include explicit checks and assignments.

### DIFF
--- a/perf/images.yaml
+++ b/perf/images.yaml
@@ -94,5 +94,5 @@ implementations:
       dockerfile: Dockerfile
     transports: [tcp, quic-v1]
     secureChannels: [noise, tls]
-    muxers: [yamux, mplex]
+    muxers: [yamux]
 
--- a/perf/images/dotnet/v1.0/Dockerfile
+++ b/perf/images/dotnet/v1.0/Dockerfile
@@ -16,6 +16,18 @@ RUN dotnet publish -c Release -o /app/publish --no-restore \
 # Runtime stage (slim image with only the .NET runtime)
 FROM mcr.microsoft.com/dotnet/runtime:9.0
 
+# Install libmsquic for QUIC transport support
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends wget && \
+    wget https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb -O packages-microsoft-prod.deb && \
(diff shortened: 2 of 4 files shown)


seetadev:
@K-21 : Great, thank you so much for sharing this PR. Appreciate it.

Wish to invite @acul71, yashksaini-coder and @sumanjeet0012 to review this PR with me in tomorrow's call. 

CCing @ItshMoh too.

PR #833, opened 2026-02-20
Update transport implementations

MarcoPolo:
No framework changes. Just updating the transports and attempting to have CI run on self hosted runners.


dhuseby:
I think the smart thing to do here is to set up Github runners.

dhuseby:
I can set up one of the self-hosted runners for this repo to use. Let me do that, you'll need it. The GH hosted runners won't run long enough to do a full transport interop pass with this framework.

MarcoPolo:
This PR doesn't change anything about the test framework itself (except an explicit string casting in yaml), so I'll go ahead and merge this so future PRs can use CI properly.

PR #838, opened 2026-03-15
Add gossip interop tests for jvm-libp2p

lucassaldanha:
Adding jvm-libp2p latest stable version to gossip tests. Followed instructions from https://github.com/libp2p/test-plans/blob/master/gossipsub-interop/test-specs/implementation.md.


lucassaldanha:
@MarcoPolo finally found some time to fix it :)

PR #852, opened 2026-04-27
feat: add zig gossipsub interop

bomanaps:


gossipsub-interop/Makefile
MarcoPolo: Do you want to add a zig test case here so CI tests it as well?

MarcoPolo:
please rebase to the latest changes

bomanaps:
> please rebase to the latest changes

Sorry for the late response  added the Zig CI workflow.

.github/workflows/gossipsub-interop-pr.yml
```diff
@@ -15,13 +15,22 @@ jobs:
   test:
     runs-on: ubuntu-22.04
     steps:
-      - name: Checkout code
+      - name: Checkout test-plans
         uses: actions/checkout@v4
+        with:
+          path: test-plans
```
MarcoPolo: Why?

gossipsub-interop/run-shadow.sh
MarcoPolo, line 1: Why do we need this file?

gossipsub-interop/run.py
```diff
@@ -38,6 +38,13 @@ def main():
         "Nodes are split evenly across them.",
     )
     parser.add_argument("--output_dir", type=str, required=False)
+    parser.add_argument(
```
MarcoPolo: If the binary is already built, then rebuilding it should finish quickly. It's simpler and less error prone to always rebuild.

gossipsub-interop/zig-libp2p/src/main.zig
```diff
@@ -0,0 +1,106 @@
...
+    // ------------------------------------------------------------------ //
+    var loop: libp2p.thread_event_loop.ThreadEventLoop = undefined;
+    try loop.init(allocator);
+    defer {
+        loop.close();
+        loop.deinit();
+    }
+
+    std.log.info("Phase 2.1 skeleton started — node_id={d} peer_id={s}", .{ node_id, peer_id_str });
+    // Phase 2.2 will read params_path and execute experiment instructions here
```
MarcoPolo: It seems like you haven't implemented the instruction interpreter, so this isn't going to do much and fail interop.

MarcoPolo (changes requested):

GrapeBaBa:
I think we can't support it until shadow support QUIC, we only have QUIC transport.

MarcoPolo:
Shadow supports QUIC just fine, but you have to stick with the basic sendmsg/recvmsg syscalls and avoiding certain features. See https://github.com/shadow/shadow/issues/3623 for more context.

PR #857, opened 2026-05-30
feat(transport-interop): add Haskell (libp2p-hs) implementation

adust09:
Register libp2p-hs as a transport-interop implementation (haskell-v0.1), supporting TCP + Noise + Yamux.

- impl/haskell/v0.1/Makefile: fetches libp2p-hs at a pinned commit and builds its root Dockerfile (which produces the libp2p-interop daemon) via dockerBuildWrapper.sh, writing image.json.
- versionsInput.json: add haskell-v0.1 (transports: tcp, secureChannels: noise, muxers: yamux). containerImageID is omitted so the canonical image.json lookup is used.

### DIFF
--- a/transport-interop/impl/haskell/v0.1/.gitignore
+++ b/transport-interop/impl/haskell/v0.1/.gitignore
@@ -1,0 +1,3 @@
+libp2p-hs-*
+*.zip
+image.json
--- a/transport-interop/impl/haskell/v0.1/Makefile
+++ b/transport-interop/impl/haskell/v0.1/Makefile
@@ -1,0 +1,25 @@
+image_name := haskell-v0.1
+commitSha := 574acdc43f824090f8d397e4e36ea8a5d2fb8317
+
+all: image.json
+
+# Build the image from libp2p-hs's own root Dockerfile, which compiles the
+# libp2p-interop daemon. dockerBuildWrapper.sh runs:
+#   docker buildx build --load -t $(image_name) .
+image.json: libp2p-hs-${commitSha}
+	cd libp2p-hs-${commitSha} && IMAGE_NAME=${image_name} ../../../../dockerBuildWrapper.sh .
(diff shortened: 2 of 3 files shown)


adust09:
### CI diagnosis: failures are caused by the image not being published, not a protocol/interop issue

I dug into the two red checks (`build (haskell) / build (v0.1)` and `combine-results`) and they trace back to a single root cause: a fork PR cannot push the image to the org GHCR.

**1. `build (haskell)` — the image builds fine, only the push is denied:**

```
denied: installation not allowed to Create organization package
##[error]Process completed with exit code 1.
```

The Haskell Docker image compiles successfully; the job only fails at the final "push to `ghcr.io/libp2p/...`" step because a PR from an external fork lacks permission to create an org-level package.

**2. `combine-results` — red because the Haskell image can't be pulled:**

I pulled the shard artifacts and inspected `results.csv`. **Every** pair involving `haskell-v0.1` fails (31/31), including:

```
"haskell-v0.1 x haskell-v0.1 (tcp, noise, yamux)",failure
```

Since even the haskell-vs-haskell self-pair fails, this is not a byte-level spec mismatch with any specific implementation — the interop runners simply can't pull the haskell image that step 1 never published. Meanwhile the non-haskell pairs are green (1240 success), confirming the harness itself is healthy.

**Causal chain:**
```
fork PR has no org GHCR push permission
  -> build (haskell) fails to publish the image
  -> every haskell-* pair fails to pull the image
  -> dashboard shows :red_circle:
  -> combine-results exits 1
```

The implementation's interop is verified locally (tcp+noise+yamux against go-libp2p and rust-libp2p pass). Could a maintainer re-run the workflow with package-write permissions, or advise the preferred way to publish the image for fork PRs? Happy to adjust the Makefile / Dockerfile if there's a recommended pattern. Thanks!

PR #863, opened 2026-08-31
Add Go implementation and relay to hole-punch test

iand:
Adds a go-libp2p v0.49 client and a go relay to hole-punch-interop, and moves the simulated internet onto routable address space so the go client can hole punch.

There are three principle changes:

***Move the internet network to routable space (src/generator.ts).***

This pins each test's internet network to a per-test /24 in 11.0.0.0/8, hashed from the test name so parallel runs don't collide. The LAN networks are unchanged.

`go-libp2p` starts DCUtR only once the host holds an address multiaddr reports as public; it drops any address failing IsPublicAddr. On Docker's default pool the relay-observed address is RFC 1918, so the candidate list empties and no CONNECT is sent. rust skips this filter, which is why the suite passes today. 11.0.0.0/8 is  allocated but unannounced. 

***Add the go-libp2p v0.49 client (impl/go/v0.49/, versions.ts)***

The default matrix gains go x go, go x rust, rust x go on both transports. The listener reserves with client.Reserve for a definite completion signal; both peers force private reachability; both wait for the DCUtR handler before acting; the dialer judges a connection direct by the absence of /p2p-circuit and pings over it, so the RTT is the hole-punched path. Transports are tcp (noise + yamux) and quic using muxer since v0.49 doesn't support mplex.

***Add the go relay and make the relay selectable (relay/, testplans.ts, Makefile)***

I moved both relays move under `relay` to mirror `impl`. Rust is now at `relay/rust` and the new go relay is at `relay/go`. testplans.ts picks the image via RELAY_IMPL (default rust). The relay is a shared component, so one relay serves every cell. The go relay matches the rust on: listens tcp + quic, forces public reachability, pushes each address twice to the redis keys

***Verification***

Full eight-cell matrix passes on both relays (local docker, WORKER_COUNT=1, zero failures). Each cell reports ~200 ms direct RTT vs ~250 ms over the relay, confirming the hole-punched path. rust x rust passing confirms the subnet change is no regression. A WORKER_COUNT=4 run through the rust relay passes all eight cells with eight distinct, non-overlapping subnets.


iand:
@MarcoPolo

seetadev:
Thanks for the solid effort on this PR, @iand! 🙌

This is a substantial addition to the hole-punch test plan. Adding the go-libp2p implementation, Go relay, routable test networks, and the expanded cross-implementation/transport matrix gives the test suite much broader coverage. The verification across both relay implementations and parallel workers is especially useful.

Would be great to get a couple more reviews before merging.  @sukunrt, @MarcoPolo and @gmelodie, could you please take a look when you get a chance? Thanks!

gmelodie (approved):

PR #864, opened 2026-09-04
Add JS DCUtR implementation to hole-punch tests

iand:
Note that this is stacked on the Go implementation PR (#863). Commit ebb8ebe346c968fc8ee54dc044a4883b5f89dff0 onwards is the JS implementation. 

The JS implementation is currently marked as "known-failing" since neither the TCP nor the QUIC tranports can successfully holepunch. The two transports fail at different points:

**TCP**: Node.js does not support `SO_REUSEPORT`, so identify drops every observed TCP address and the simultaneous-open dial uses a fresh source port rather than the listener's. The client then advertises only undialable addresses, and the punch never completes.  

**QUIC**: the exchange runs end to end but the dial still times out. There are two causes (1) js-libp2p confirms addresses through AutoNAT so DCUtR omits the address unless the client confirms it directly and (2) `@chainsafe/libp2p-quic` dials from a separate socket, so the observed address points at the relay-dial socket rather than the listener

See [js-libp2p issue 2620](https://github.com/libp2p/js-libp2p/issues/2620) and [discussion 2388](https://github.com/libp2p/js-libp2p/discussions/2388).


seetadev:
Thanks for the effort on this PR, @iand! 🙌

Great to see the JS DCUtR implementation added to the hole-punch test plan, along with the relay setup and the documentation of the current TCP/QUIC limitations. This should provide a useful foundation for tracking and improving JS hole-punching support.

Would be great to get a few more eyes on this before merging. @tabcat, @gmelodie @dozyio,  @johannamoran : could you please take a look and review when you get a chance? Thanks!

gmelodie:
Can this be broken down into smaller PRs? it's very easy to miss stuff in a 3k+ loc PR

iand:
The PR is stacked on the previous one https://github.com/libp2p/test-plans/pull/863 which sets up the test harness. This PR only adds a JS implementation

PR #865, opened 2026-09-15
autonat: add AutoNAT v2 interop suite with go-libp2p and rust-libp2p

iand:
Adds a new autonat-interop suite alongside transport-interop and hole-punch-interop. It
checks AutoNAT v2 reachability between libp2p implementations: a client offers one of
its own addresses and asks a server to confirm it is reachable; the server dials that
address back and reports the verdict.

The suite ships with two implementations, go-libp2p v0.49 and rust-libp2p 0.56. Each one
runs as both the client and the server, over both TCP and QUIC, to give eight test
combinations. The peers sit on a single flat Docker network where both hold private
addresses and can dial each other directly, and both run AutoNAT with private addresses
allowed, so there is no need for any NAT or routing setup. Coordination happens over
redis in the same way as the other suites: the server publishes its address, the client
picks it up, runs the reachability check, prints a one-line JSON verdict, and exits
with a non-zero code if the verdict is negative. The runner uses that exit code as the
result for the cell. 

Seven of the eight combinations pass. The one that fails is a Go client asking a Rust
server to verify a QUIC address.

In the failing case the dial-back connection is established and the nonce exchange
completes, but the Rust server never receives the client's dial-back response, so it
reports the check as failed. The reason is that the Rust server strictly waits to read
the response, while the Go client writes the response and then immediately resets the
stream. Over QUIC a reset truncates any data still in flight, so the response is lost
before the server reads it.

The same pair over TCP passes, because TCP delivers the buffered response before the
reset takes effect.


iand:
Matrix results from github action. I chose to follow the pattern of `transport-interop` that shows failures as yellow dots using a `knownErrors.json` file. 

## Using: tcp
| ⬇️ client 📞 \  ➡️ server 🎧 | go-v0.49 | rust-v0.56 |
| ---------------------------- | -------- | ---------- |
| go-v0.49 | :green_circle: | :green_circle: |
| rust-v0.56 | :green_circle: | :green_circle: |

## Using: quic
| ⬇️ client 📞 \  ➡️ server 🎧 | go-v0.49 | rust-v0.56 |
| ---------------------------- | -------- | ---------- |
| go-v0.49 | :green_circle: | :yellow_circle: |
| rust-v0.56 | :green_circle: | :green_circle: |

